PR #24, opened 2011-09-19
Added SUUpdater delegate method publicDSAKeyForUpdater as discussed

ekscrypto:
Signed-off-by: Dave Poirier <dave@madefresh.ca>


pornel:
This method won't work with sandboxing, since a compromised host app could set attacker's key.

PR #138, opened 2012-02-06
Add support for sparkle:requiredVersion to signify mandatory program updates.

toddfoster:
Our cloud-centric application occasionally requires server-side changes which break older clients. In that case, the application should not be allowed to run without updating.

Check the appcastItem's enclosure for a field labeled
"sparkle:requiredVersion". If found, compare that version to the host's
currently-running version. If the host is not running at least the
required version, do not allow it to launch without an update.  This
means hiding the "Skip" and "Later" buttons in the Update alert, and
terminating the program if the user closes the update alert dialog (via
the window control) without launching the update.

### DIFF
--- a/SUBasicUpdateDriver.h
+++ b/SUBasicUpdateDriver.h
@@ -33,6 +33,7 @@
 - (BOOL)hostSupportsItem:(SUAppcastItem *)ui;
 - (BOOL)itemContainsSkippedVersion:(SUAppcastItem *)ui;
 - (BOOL)itemContainsValidUpdate:(SUAppcastItem *)ui;
+- (BOOL)itemContainsRequiredUpdate:(SUAppcastItem *)ui;
 - (void)didFindValidUpdate;
 - (void)didNotFindUpdate;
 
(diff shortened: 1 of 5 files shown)


pornel:
I think silent updates are a better user experience and achieve the same goal.

gaojinhsu:
> Our cloud-centric application occasionally requires server-side changes which break older clients. In that case, the application should not be allowed to run without updating.
> 
> Check the appcastItem's enclosure for a field labeled
> "sparkle:requiredVersion". If found, compare that version to the host's
> currently-running version. If the host is not running at least the
> required version, do not allow it to launch without an update. This
> means hiding the "Skip" and "Later" buttons in the Update alert, and
> terminating the program if the user closes the update alert dialog (via
> the window control) without launching the update.



> Our cloud-centric application occasionally requires server-side changes which break older clients. In that case, the application should not be allowed to run without updating.
> 
> Check the appcastItem's enclosure for a field labeled
> "sparkle:requiredVersion". If found, compare that version to the host's
> currently-running version. If the host is not running at least the
> required version, do not allow it to launch without an update. This
> means hiding the "Skip" and "Later" buttons in the Update alert, and
> terminating the program if the user closes the update alert dialog (via
> the window control) without launching the update.

the update alert window is not modal, users still can use the app by ignoring the update alert window?

Issue #348, opened 2014-06-30
Run headless installer command


pornel:
Currently Sparkle uses `open -b com.apple.installer` to install `.pkg` via GUI installer.

This is problematic, because:
- It's a hassle for the user to click through the same steps on every update
- It looks ugly when Sparkle window and installer are open at the same time #288 
- GUI installer doesn't report failures back to Sparkle

How about using `/usr/sbin/installer` instead? It's headless, command-line tool.

PR #377, opened 2014-07-13
Added headless guided package installation

grahammiln:
Added support for headless guided installation of `pkg` and `mpkg` files; see https://github.com/sparkle-project/Sparkle/issues/348. This allows the installation of Installer packages without a user interface. This method also allows one application to update a suite of tools through an Installer package.

Ported existing code to use ARC and inlined the documentation. The original code has been well tested and used in commercial products for many years. That code supported 10.4 and later. I have minimised the changes to avoid introducing bugs or problems. I expect further changes can be made given the shift to supporting 10.7+.

Ideally, the use of `AuthorizationExecuteWithPrivileges` should be replaced with an XPC or an external authtool like process.


grahammiln:
Thank you for accepting this pull request.

I have added documentation for guided package installation to https://github.com/sparkle-project/Sparkle/wiki/Package-Updates

PR #469, opened 2014-12-23
Add ability for host bundle to specify a custom name

kainjow:
At our work our app names have the version number in them directly, for example "MyApp 3" and so some of Sparkle's messages read awkwardly when the version comes immediately after the name. So this change allows us to set the name that Sparkle uses to "MyApp" by adding an entry to the Info.plist file.

### DIFF
--- a/Sparkle/SUHost.m
+++ b/Sparkle/SUHost.m
@@ -102,7 +102,13 @@ typedef struct {
 
 - (NSString *)name
 {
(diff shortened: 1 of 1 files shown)


pornel:
Thanks!

Could you document it the Wiki? (e.g. https://github.com/sparkle-project/Sparkle/wiki/Make-Preferences-UI)

kainjow:
Updated https://github.com/sparkle-project/Sparkle/wiki/Customization

pornel:
Thanks!

PR #471, opened 2014-12-27
Using a weak id delegate property will crash during instantiation of NIB on OS X 10.7

GerTeunis:
When adding SUUpdater in IB (in the NIB) and making a reference from the SUUpdater delegate to a ViewController will crash on OS X 10.7 because the delegate is a weak property.

Crash report:

```
objc[308]: garbage collection is OFF
objc[308]: cannot form weak reference to instance (0x7fc55b481cf0) of class MainWindowViewController

Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   libobjc.A.dylib                 0x00007fff97cb5768 _objc_trap + 4
1   libobjc.A.dylib                 0x00007fff97cb58aa _objc_fatal + 195
2   libobjc.A.dylib                 0x00007fff97cc14ad weak_register_no_lock + 346
3   libobjc.A.dylib                 0x00007fff97cc1a59 objc_storeWeak + 360
4   com.apple.CoreFoundation        0x00007fff8d49770d -[NSObject performSelector:withObject:] + 61
5   com.apple.AppKit                0x00007fff8c7c5baa -[NSNibOutletConnector establishConnection] + 405
6   com.apple.AppKit                0x00007fff8c7c32a1 -[NSIBObjectData nibInstantiateWithOwner:topLevelObjects:] + 1079
7   com.apple.AppKit                0x00007fff8c7b98bb loadNib + 322
8   com.apple.AppKit                0x00007fff8c7b8db8 +[NSBundle(NSNibLoading) _loadNibFile:nameTable:withZone:ownerBundle:] + 217
9   com.apple.AppKit                0x00007fff8c7b8cd3 +[NSBundle(NSNibLoading) loadNibFile:externalNameTable:withZone:] + 141
10  com.apple.AppKit                0x00007fff8c7b8c16 +[NSBundle(NSNibLoading) loadNibNamed:owner:] + 364
11  com.apple.AppKit                0x00007fff8ca29cd7 NSApplicationMain + 398
12  com.VortexApps.NZBVortex3       0x0000000100785f43 main + 275
13  com.VortexApps.NZBVortex3       0x000000010073e604 start + 52
```

### DIFF
--- a/Sparkle/SUUpdater.h
+++ b/Sparkle/SUUpdater.h
@@ -25,7 +25,7 @@
  */
 SU_EXPORT @interface SUUpdater : NSObject
 
-@property (weak) IBOutlet id<SUUpdaterDelegate> delegate;
+@property (unsafe_unretained) IBOutlet id<SUUpdaterDelegate> delegate;
 
 + (SUUpdater *)sharedUpdater;
 + (SUUpdater *)updaterForBundle:(NSBundle *)bundle;


jakepetroules:
We could use unsafe_unretained only if `MAC_OS_X_VERSION_MIN_REQUIRED < 1080`?

GerTeunis:
Those are framework compile time not client compile time.

jakepetroules:
I'm aware of that; it would still be better than leaving unsafe_unretained in the codebase.

GerTeunis:
It will not solve or remove the need for the unsafe_unretained as it has to be in there during compile of the release build of the framework with release target 10.7+ which the framework does. 

It will not solve the problem. Documentation the line including a warning when complied for 10.8 our newer does prevent forgetting this patch.

jakepetroules:
> It will not solve or remove the need for the unsafe_unretained as it has to be in there during compile of the release build of the framework with release target 10.7+ which the framework does.

People can still build their own Sparkle for 10.8+ using `xcodebuild MAC_OS_X_VERSION_MIN_REQUIRED=1080`, so it does help some.

> Documentation the line including a warning when complied for 10.8 our newer does prevent forgetting this patch.

No, no warnings. We can use `if MAC_OS_X_VERSION_MIN_REQUIRED < 1080`. Then once Sparkle itself drops support for 10.7 the code base is ready,

kainjow:
@jakepetroules I think that would be a good solution.

GerTeunis:
@jakepetroules you have a point there indeed.
Created a pull request where I add comment explaining the work-around and also use weak property when building for target 10.8+

https://github.com/sparkle-project/Sparkle/pull/473

PR #473, opened 2015-01-02
Use a weak property instead of a unsafe_unretained in SUUpdater when target SDK is 10.8

GerTeunis:
Use a weak property instead of a unsafe_unretained in SUUpdater when target SDK is 10.8

### DIFF
--- a/Sparkle/SUUpdater.h
+++ b/Sparkle/SUUpdater.h
@@ -25,7 +25,13 @@
  */
 SU_EXPORT @interface SUUpdater : NSObject
 
+// Work-around for a OS X 10.7 bug when using a NSViewController / NSWindowController as a weak property will crash.
+// Crash report: objc[308]: cannot form weak reference to instance (0x7fc55b481cf0) of class MainWindowViewController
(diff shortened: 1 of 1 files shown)


Sparkle/SUUpdater.h
```diff
@@ -25,7 +25,13 @@
  */
 SU_EXPORT @interface SUUpdater : NSObject
 
+// Work-around for a OS X 10.7 bug when using a NSViewController / NSWindowController as a weak property will crash.
+// Crash report: objc[308]: cannot form weak reference to instance (0x7fc55b481cf0) of class MainWindowViewController
+#if MAC_OS_X_VERSION_MIN_REQUIRED >= 1080
```
jakepetroules: prefix with __, it's `__MAC_OS_X_VERSION_MIN_REQUIRED`

GerTeunis:
Thanks for feedback but without __ is more common now and work perfectly without the __ prefix. Check the AvailabilityMacros.h

There is another bigger problem: when the framework is build for 10.7 and client side the header is used on 10.8+

Closing pull request

jakepetroules:
good point...

Issue #441, opened 2014-10-20
Sparkle creates broken app bundles for some users


fzwo:
We have the strangest problem.

We have a relatively large app, and ship mono within the app bundle. The .app is zipped for Sparkle (~250 MB zip, 600+ MB app), and is signed.

We've had reports from several (but not all) users that the app is broken after an update. We were scratching our heads because all the nib files seemed to be missing. SparkleUpdateLogs were always clean, and we couldn't find anything in our own crash logs (both through Hockey and the log files on the machine) except for the mysteriously missing xib files.

We were now able to take a closer look at one of the broken apps. Sure enough, the nibs were missing, as was everything else starting with S (the first letter of our company prefix) in the bundle's resources folder. Actually, everything after M was missing. Or rather, everything after Mono. Digging deeper, it seems that sometime during copying mono.framework, the updater stopped.

This resulted in an app bundle that reports as broken when trying to run on a Mac with Gatekeeper on, because the signature does not match anymore. On a Mac with gatekeeper off, it's more insidious: The app starts, and runs, but certain Windows won't open because their nibs can't be found.

This has happened to a handful of users now, some of them multiple times. We'd really love to find out what's causing this, and how to protect against it. Any ideas? (We're on commit https://github.com/sparkle-project/Sparkle/commit/7e1b47976dd80c09abd9f772a4117eeb08dc15f1 - there haven't been any functional changes since then)

steiger:
FWIW, another log file excerpt and crash report (https://gist.github.com/steiger/380f99126cd52f5c53e8).

14.02.15 08:48:19.258	Mailplane 3	Sparkle: ===== Mailplane 3 =====
14.02.15 08:50:59.334	Mailplane 3	Sparkle: Extracting /Users/jo/Library/Application Support/Mailplane 3/.Sparkle/Mailplane 3 1319/Mailplane_3_1319.tbz using 'tar -jxC "$DESTINATION"'
14.02.15 08:51:38.805	ReportCrash	Saved crash report for Autoupdate[5109] version 1.8.0 (1.8.0) to /Users/jo/Library/Logs/DiagnosticReports/Autoupdate_2015-02-14-085138_betty.crash
14.02.15 09:59:20.719	Mailplane 3	Unable to load nib file: MainMenu, exiting

pornel:
I've recently merged 65238c35a9dec983f8c49f3475c2cfd5f926f3f2. Does that fix the nib loading problems for you?

[...]

PR #476, opened 2015-01-14
Fix Autoupdate.app crash on 10.7 (issue #441)

greenbuffo:
Autoupdate.app can crash in the middle of installing an update, leaving
the user with a broken application. The crash is caused by the status
window’s controller (SUStatusController) being prematurely deallocated
before the window is closed. This change prevents such a crash by
keeping a strong reference to the SUStatusController in
TerminationListener.


pornel:
Good catch!

PR #481, opened 2015-01-20
Have Sparkle build a framework module.

MaddTheSane:
This makes Xcode build a module file for the Sparkle framework. This is useful for Swift code, as it uses modules to import code.
The reason for adding the `-Wno-auto-import` warning flag is because any `#import` or `#include` is translated to an `@import`, with the current warning settings complaining.


SevenBits:
Maybe @pornel or @jakepetroules could take a look at this? Modules are the hip new thing now...

pornel:
Yeah, I'm pleasantly surprised that modules are a thing now. Let's roll with it! Thanks @MaddTheSane

PR #483, opened 2015-01-27
Update podspec for 1.9.0

stel:
It also would be great to update a spec in a Cocoapods trunk


1024jp:
Hi, it looks like that the Cocoapods' spec hasn't been updated yet. Do you have any plan for that?

jakepetroules:
Done.

1024jp:
Thanks!

PR #489, opened 2015-02-25
Ability to add custom http headers to appcast request

mgunneras:
I had a need for customizing the request further than just URL and User Agent.

This patch allows for SUAppcast (via SUUpdater) to take arbitrary headers as NSDictionary.

### DIFF
--- a/Sparkle/SUAppcast.h
+++ b/Sparkle/SUAppcast.h
@@ -19,6 +19,7 @@ SU_EXPORT @interface SUAppcast : NSObject <NSURLDownloadDelegate>
 
 @property (weak) id<SUAppcastDelegate> delegate;
 @property (copy) NSString *userAgentString;
+@property (copy) NSDictionary *httpHeaders;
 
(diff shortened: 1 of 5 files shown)


pornel:
What's the use case?

mgunneras:
I'm building an internal app where I'm hosting my appcast.xml behind an api
endpoint that requires auth. This way I can add http headers for
authorizing the request using an api token or basic auth etc.




*MATTIAS GUNNERAS*
CO-FOUNDER / CHIEF TECHNOLOGY OFFICER AT BREAKFAST


*347 829 9919   breakfastny.com <http://breakfastny.com/>  @breakfastny
<http://twitter.com/breakfastny>*

pornel:
Oh, OK, I can see it can be useful. The code is well written. Thanks for the patch!

grit96:
Is there any documentation on how to set custom headers?

grit96:
I think the custom headers will only apply for the appcast download, not the actual application download - is there a workaround for using the headers for both downloads?

PR #491, opened 2015-02-26
Remove binaryDeltaSupported…

MaddTheSane:
…as libxar is no longer weak-linked.


MaddTheSane:
This came from me building Sparkle on Xcode 6.3 beta 2.

However, there seems to be quite a few warnings saying that `using %s directive in NSString which is being passed as a formatting argument to the formatting method [-Wcstring-format-directive]`, most of them from Apple's assert macros. Those warnings aren't present on Xcode 6.1

Issue #508, opened 2015-03-30
Should Sparkle call LSRegisterURL?


zorgiepoo:
LSRegisterURL() registers or updates an application into Launch Service's database - should Sparkle use this after a plain or delta update of an application?

I found this bug, but perhaps it was not reported:
https://jira.atlassian.com/browse/SRCTREE-985

I see no reference to this function in Sparkle's source code. Note that this function can't be called from a sandboxed application (https://developer.apple.com/library/mac/documentation/Security/Conceptual/AppSandboxDesignGuide/AppSandboxDesignGuide.pdf).

michelf:
Sparkle calls the `mdimport` command line tool [here][1]. I'd tend to believe that would register whatever is in the Info.plist file, but perhaps not.

[1]:  https://github.com/sparkle-project/Sparkle/blob/bc99dbd560356aab8de86d6d5548d29efd36d647/Sparkle/SUInstaller.m#L144

pornel:
That's interesting. I thought that launching a newly updated app would be enough. I think we can add it.

pornel:
Fixed, thanks @grahammiln

PR #509, opened 2015-04-01
Register update with Launch Services; see #508

grahammiln:
Call [LSRegisterURL](https://developer.apple.com/library/mac/documentation/Carbon/Reference/LaunchServicesReference/index.html#//apple_ref/c/func/LSRegisterURL) on the updated application. See #508 for details.

### DIFF
--- a/Sparkle/SUInstaller.m
+++ b/Sparkle/SUInstaller.m
@@ -162,10 +162,24 @@ static NSString *sUpdateFolder = nil;
     }
 }
 
++ (void)registerWithLaunchServices:(NSString *)installationPath
+{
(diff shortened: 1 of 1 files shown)


pornel:
Brilliant. Thanks!

kainjow:
Why is installationPath being casted to a string if it's already a string?

grahammiln:
Sorry. It is an oversight during development. Feel free to push a fix to remove the `NSString` cast.

grahammiln:
Sorry. It is a mistake. I have pushed an update to remove the redundant `NSString` cast in #510

Issue #474, opened 2015-01-05
BinaryDelta fails when tree contains a file with changed case filename


nikolaikrill:
So I've just spent the better part of a day trying to figure out why my deltas are failing to be applied with the latest version of my app. Turns out the problem is that I renamed a file in my app tree, e.g. myfile.plist to MyFile.plist. It would appear that Sparkle does not behave correctly with OS X's case insensitive file system. When applying the delta with the file name change, BinaryDelta fails with "Destination doesn't have expected hash (0e50e73926230b1daaf2819ef8ad3dfe725fd177 != dac90ddf9a293857cbd0b340239b4b4d5dd84da9).  Giving up.".

I took a look at the destination tree that had the wrong hash, and sure enough I found that the myfile.plist was completely missing from the destination tree. Please fix this guys, and thanks for this great framework!

PR #513, opened 2015-04-03
Fix BinaryDelta failing on filenames changing case

zorgiepoo:
When a filename changes case, make sure we have the file deleted in the old tree before being extracted into the new one.
Note that we don't try creating a diff if a file is renamed with a different case, just like we don't track file names changing in other ways.
See issue #474


pornel:
That looks good. Thanks for the fix.

I'm concerned that we don't have tests for binary delta. Would you like to create some?

zorgiepoo:
I'll try to take a jab at tests. I was also able to pinpoint another issue with binary delta's, so I've submitted another pull request for that as well.

By the way, the wiki states that BinaryDelta must have Sparkle.framework be present in the same directory, but I see no evidence for this being true.

pornel:
Thank you.

I've merged these pull requests and I'm waiting for the tests.

Issue #158, opened 2012-02-27
Impossible Delta


Mento:
If a complex structure of symlinks is used in the tree it's impossible to apply a created delta patch.

Steps that will reproduce the problem:
Extract the files from http://dl.dropbox.com/u/61193827/7dje76shi8dsi78sdf/Impossible%20Delta.zip```
BinaryDelta create A B patch
BinaryDelta apply A C patch
```
What is the expected result:
C is identical to B

What happens instead:
BinaryDelta says *Destination doesn't have expected hash XXX != YYY*

PR #514, opened 2015-04-04
Fix BinaryDelta sometimes failing when a dir becomes a dir link

zorgiepoo:
This can occur in the following case:
```
original_tree:
	A/
	Current/
		B

new_tree:
	A/
		B
	Current -> A/
```
Because symbolic links are not followed, Current/B can be marked as deleted. If A/B is added before the deletion, then A will end up as an empty directory.

Our fix is to remove all child paths of a directory from consideration when it changes from a directory to another type of file. This should also decrease the size of the diff even in harmless cases such as when a directory changes into a regular file.


zorgiepoo:
This is likely related to issue #158; not sure why it was closed.

PR #515, opened 2015-04-05
Change bsdiff's suffix sort algorithm to sais. Fixes #132

### LINKED ISSUE #132: BinaryDelta fails with... Bus error: 10
Trying to do a delta between two versions of Adium.
Fails with -> Generating delta...  Bus error: 10

Here are some logs: 
http://cl.ly/0L1v3o3o2q2n1n2I1u15

Files used for deltas are in this dmg's:
http://download.adium.im/Adium_1.5b8.dmg
http://download.adium.im/Adium_1.5b9.dmg

System info:

OS X Lion 10.7.2
Compiled with Xcode 4.2.1 as 64bit

Notes:

Works fine on OS X SL 10.6.8
with Xcode 3.2.6, compiled as 64bit

Thanks.

zorgiepoo:
Although #132 is marked as closed, the issue is still real. bsdiff can crash or hang when creating diffs on some files. In particular, bsdiff hangs/crashes in its recursive split function.

As you can see from:
http://stackoverflow.com/a/20305493/871119
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=409664

bsdiff 4.3 uses qsufsort for its suffix sorting algorithm, which has a bug causing the hanging/crashing. Our solution is to change the algorithm to use sais for suffix sorting:
https://sites.google.com/site/yuta256/sais

Notes:
* I had to change sais_index_type #define from int to off_t in sais.h/c and in some other places, to eliminate all warnings. Compiles cleanly with clang's -Weverything

* sais is claimed to be more efficient than other algorithms including qsufsort. If a more efficient (and stable) algorithm comes out, we may be able to swap it out in the future.

--
I'll start to look into writing actual tests soon. Going through other people's issues/tests just seems to prioritize higher ;).

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -30,6 +30,9 @@ EXTERNAL LICENSES
 bspatch.c and bsdiff.c, from bsdiff 4.3 <http://www.daemonology.net/bsdiff/>:
     Copyright (c) 2003-2005 Colin Percival.
 
+sais.c and sais.c, from sais-lite (2010/08/07) <https://sites.google.com/site/yuta256/sais>:
+    Copyright (c) 2008-2010 Yuta Mori.
+
 SUDSAVerifier.m:
     Copyright (c) 2011 Mark Hamlin.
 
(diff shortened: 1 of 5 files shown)


pornel:
Thanks for the fixes.

It's curious that the issue hasn't been fixed upstream. There's [a fork](https://github.com/mendsley/bsdiff) with [a PR with sais as well](https://github.com/mendsley/bsdiff/pull/9) but unfortunately it's not compatible.

…which makes me wonder, is this modification backwards-compatible with existing diffs?

zorgiepoo:
Hm, that's interesting. Looks like the PR request for that repo got their fix idea from the same person I did in that Stack Overflow thread.

That fork of bsdiff is not intended to be compatible with the original bsdiff version that we use, which is not related to using sais. My pull request shouldn't require changes to bspatch.

PR #522, opened 2015-04-11
Binary delta tests

zorgiepoo:
Here are some tests for the Binary Delta tool as requested. The tests try to create & apply patches, and it tests the tree hashes as well. I have found one bug with the binary delta tool not being able to create proper patches and do proper verification if file permissions change, but this could be opened into another issue. Let me know if you have any questions.


pornel:
Brilliant. Thank you very much. This gives a lot more confidence in the binary delta tool.

Issue #391, opened 2014-07-24
error while extracting archive and relaunching app.


minorblend:
After downloading and extracting update archive file, I select "install and relaunch" and get the following error.

> An error occurred while relaunching MyApp, but the new version will be available next time you run MyApp.

I couldn't figure out what's wrong since there's nothing left in Console. and the app is actually not updated at all.

How can I fix this?

Issue #376, opened 2014-07-12
Release 1.8


kornelski:
Is there any outstanding development blocking release?

Also, shall we do betas?

PR #523, opened 2015-04-13
Verify final code signature for issue #441

iwankerl:
This should help prevent situations like issue #441 where Sparkle
creates broken app bundles.

Once the app has been installed, we check to make sure the final
copy has a valid code signature (if it's required to be signed).


zorgiepoo:
It's worth investigating if you have to do an OS check for a deep signature test.

See:
https://github.com/yahoo/Sparkle/blob/master/SUCodeSigningVerifier.m#L51 (fork)
https://github.com/sparkle-project/Sparkle/issues/376#issuecomment-48824267
https://github.com/sparkle-project/Sparkle/issues/391

pornel:
Just to let you know — it looks like a very important improvement. I'm currently very busy, but I'm going to give it a thorough review as soon as I can.

iwankerl:
I may back out of using the SecCSCheckNestedCode flag.  Thanks for pointing out its potential issue.

I also plan on doing more testing of this today.

kainjow:
What if the app isn't code signed at all? AFAIK, Sparkle doesn't require apps to be code signed.

zorgiepoo:
That should still be handled but I think Sparkle is too lenient on providing an invalid DSA or code sig as long as one of them works, but not both have to be (if both are provided).

zorgiepoo:
To elaborate on what I was saying above. If we look at the current code in this pull request at https://github.com/iwankerl/Sparkle/blob/8bbe487c86886d27136054f0dc9923599baeec3f/Sparkle/SUBasicUpdateDriver.m#L222

We can see the logic allows a broken code signature to be specified, as long as the DSA is valid. Or vise versa: the logic allows the DSA to be invalid, as long as the code signature is valid.

I think the logic should change to:
* First check If DSA is provided. If it's provided, but not valid, then fail.
* Check if it's an app or package. If it's a package we can stop here and succeed.
* Then check if the new app is code signed (not unsigned). If it's code signed, but not signed properly, then fail.
* If a DSA was not provided, and a code signature was not provided, then fail.
* If no DSA was provided, but a valid code signature was provided, but it differs from the host signature, then fail.
* If we get here, then validation is successful.

DSA is still useful and shouldn't be treated as an "alternative" to code signing IMO. It's for validating the download. Plus, archives can preserve important attributes that Apple's code signing doesn't bother to check for (file permissions!), and we know from the past that the code signing is not perfect. If both DSA and code signing is provided, we should make sure both are correct.

iwankerl:
zorgiepoo has a valid point.  The DSA of the downloaded file should probably be checked every time if it is provided.

On another note, in practice my pull request is not testing very well and I don't recommend merging it.  It is proving difficult to replace the old app if a corrupt one is being or has been installed. What I would like to do in the mean time is add another updater delegate method so the host app can do some verification of the extracted item before the autoupdate install tool is launched.  Something like:

`- (BOOL)updater:(SUUpdater *)updater shouldContinueInstallingUpdate:(SUAppcastItem *)item extractedAppPath:(NSString *)extractedAppPath;`

We could still do a final code signature validation from the autoupdate tool and log unfavorable results but perhaps not cause an install failure. It's still not clear to me if the root cause of this bug is the in the extraction phase or the replacement of the old app phase.

zorgiepoo:
If the issue lies on Sparkle crashing during copying the app to its file destination, then no amount of verification is going to improve this issue. Which begs the question of why a copy operation of the app is being done after the application has been unarchived/patched? Often with copying files, file moves are often done as the final steps since they're atomic operations. If the destination is on a different volume from the source, you can copy it to a temporary path near the destination, then do a move as a final step. On the other hand, if they're on the same volume and permissions are already sufficient or what not, a copy just takes extra time.

pornel:
This is very well written. Thanks for refactorings and checking the docs.

Still, the root cause of breakage needs to be fixed, but this check won't hurt.

PR #524, opened 2015-04-18
Binary delta Improvements

zorgiepoo:
OK, been working on this for a bit.

I found several bugs with BinaryDelta's tree hashing, and added the ability for BinaryDelta to diff/preserve permissions. I also created major/minor version fields in the file format. If the major field differs, this indicates the patch not being compatible. When the minor version differs, compatibility still works. Due to having to add permission diffs and having to change the hashing, I had to bump the major version.

Changes summary:
* Permissions are preserved and hashed on regular files, symbolic links, and directories now in all cases. Only permission modes 0755 and 0644 are allowed on files in the after-tree. Practical use: scripts turning into executable or vise versa, .app being compressed incorrectly with wrong permissions, fixing permissions from an old app in a new update...
* ACLs (Access Control Lists) are not allowed in either before-tree or after-tree; we don't try to preserve or hash them, and we don't want to diff on them. Extended file attributes are not hashed and completely ignored as they shouldn't contain vital information.
* _DARWIN_NO_64_BIT_INODE define was removed because I believe it's no longer needed since we only target 64-bit now.
* Directories are now hashed. They were intended to be by the original code since _hashOfFileContents has a case for them. This was a bug.
* File types (regular file, directory, symbolic link) are now hashed. This prevents hash collisions for two different trees in a few cases.
* More error handling is now done. Cleaned up code a bit by adding constants. Added a lot of tests.
* Added major/minor version fields to the patch format. Bumped the (major) patch version to 2. Patches created by BinaryDelta now will not work on older versions unless you create a patch by supplying --version=1; this is meant for transition purposes. BinaryDelta can still apply version 1 patches. Since major/minor versioning wasn't around before, for version 2 patches I had to change the before and after hash property keys to make the patch gracefully fail on older versions of BinaryDelta.

After this, I'm thinking of adding a verbose option and improve the current logging.

### DIFF
--- a/Sparkle/SUBinaryDeltaApply.m
+++ b/Sparkle/SUBinaryDeltaApply.m
@@ -36,46 +36,46 @@ int applyBinaryDelta(NSString *source, NSString *destination, NSString *patchFil
     uint16_t majorDiffVersion = FIRST_DELTA_DIFF_MAJOR_VERSION;
     uint16_t minorDiffVersion = FIRST_DELTA_DIFF_MINOR_VERSION;
 
-    NSString *expectedBeforeHashv1_0 = nil;
-    NSString *expectedAfterHashv1_0 = nil;
+    NSString *expectedBeforeHashv1 = nil;
+    NSString *expectedAfterHashv1 = nil;
     
-    NSString *expectedBeforeHashv1_1 = nil;
-    NSString *expectedAfterHashv1_1 = nil;
+    NSString *expectedNewBeforeHash = nil;
+    NSString *expectedNewAfterHash = nil;
     
     xar_subdoc_t subdoc;
     for (subdoc = xar_subdoc_first(x); subdoc; subdoc = xar_subdoc_next(subdoc)) {
         if (!strcmp(xar_subdoc_name(subdoc), "binary-delta-attributes")) {
             const char *value = 0;
             
-            // only available in version 1.1 or later
+            // only available in version 2.0 or later
             xar_subdoc_prop_get(subdoc, "major-version", &value);
             if (value)
                 majorDiffVersion = (uint16_t)[@(value) intValue];
             
-            // only available in version 1.1 or later
+            // only available in version 2.0 or later
             xar_subdoc_prop_get(subdoc, "minor-version", &value);
             if (value)
                 minorDiffVersion = (uint16_t)[@(value) intValue];
             
-            // deprecated since version 1.1
+            // deprecated since version 2.0
             xar_subdoc_prop_get(subdoc, "before-sha1", &value);
             if (value)
-                expectedBeforeHashv1_0 = @(value);
+                expectedBeforeHashv1 = @(value);
 
-            // deprecated since version 1.1
+            // deprecated since version 2.0
             xar_subdoc_prop_get(subdoc, "after-sha1", &value);
             if (value)
-                expectedAfterHashv1_0 = @(value);
+                expectedAfterHashv1 = @(value);
             
-            // only available in version 1.1 or later
-            xar_subdoc_prop_get(subdoc, "before-sha1-v1.1", &value);
+            // only available in version 2.0 or later
+            xar_subdoc_prop_get(subdoc, "before-tree-sha1", &value);
             if (value)
-                expectedBeforeHashv1_1 = @(value);
+                expectedNewBeforeHash = @(value);
             
-            // only available in version 1.1 or later
-            xar_subdoc_prop_get(subdoc, "after-sha1-v1.1", &value);
+            // only available in version 2.0 or later
+            xar_subdoc_prop_get(subdoc, "after-tree-sha1", &value);
             if (value)
-                expectedAfterHashv1_1 = @(value);
+                expectedNewAfterHash = @(value);
         }
     }
     
@@ -89,10 +89,10 @@ int applyBinaryDelta(NSString *source, NSString *destination, NSString *patchFil
         return 1;
     }
     
-    BOOL usesPatch1_1 = DIFF_VERSION_IS_AT_LEAST(majorDiffVersion, minorDiffVersion, 1, 1);
+    BOOL usesNewTreeHash = MAJOR_VERSION_IS_AT_LEAST(majorDiffVersion, BEIGE_MAJOR_VERSION);
     
-    NSString *expectedBeforeHash = usesPatch1_1 ? expectedBeforeHashv1_1 : expectedBeforeHashv1_0;
-    NSString *expectedAfterHash = usesPatch1_1 ? expectedAfterHashv1_1 : expectedAfterHashv1_0;
+    NSString *expectedBeforeHash = usesNewTreeHash ? expectedNewBeforeHash : expectedBeforeHashv1;
+    NSString *expectedAfterHash = usesNewTreeHash ? expectedNewAfterHash : expectedAfterHashv1;
 
     if (!expectedBeforeHash || !expectedAfterHash) {
         fprintf(stderr, "Unable to find before-sha1 or after-sha1 metadata in delta.  Giving up.\n");
@@ -100,7 +100,7 @@ int applyBinaryDelta(NSString *source, NSString *destination, NSString *patchFil
     }
 
     fprintf(stdout, "Verifying source...  ");
-    NSString *beforeHash = hashOfTreeWithVersion(source, majorDiffVersion, minorDiffVersion);
+    NSString *beforeHash = hashOfTreeWithVersion(source, majorDiffVersion);
     if (!beforeHash) {
         fprintf(stderr, "Unable to calculate hash of tree %s\n", [source fileSystemRepresentation]);
         return 1;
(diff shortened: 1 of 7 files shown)


Sparkle/SUBinaryDeltaApply.m
```diff
@@ -74,21 +130,42 @@ int applyBinaryDelta(NSString *source, NSString *destination, NSString *patchFil
...
-        if (!xar_prop_get(file, "binary-delta", &value))
-            applyBinaryDeltaToFile(x, file, sourceFilePath, destinationFilePath);
-        else
-            xar_extract_tofile(x, file, [destinationFilePath fileSystemRepresentation]);
+        if (!xar_prop_get(file, BINARY_DELTA_KEY, &value)) {
+            if (!applyBinaryDeltaToFile(x, file, sourceFilePath, destinationFilePath)) {
+                fprintf(stderr, "Unable to patch %s to destination %s\n", [sourceFilePath fileSystemRepresentation], [destinationFilePath fileSystemRepresentation]);
+                return 1;
+            }
+        } else if (xar_prop_get(file, MODIFY_PERMISSIONS_KEY, &value)) {
```
pornel: I got lost here. What is the `file`? Why presence of permissions changes behavior here?

Sparkle/SUBinaryDeltaApply.m
```diff
@@ -15,38 +15,84 @@
...
-    xar_t x = xar_open([patchFile UTF8String], READ);
+    xar_t x = xar_open([patchFile fileSystemRepresentation], READ);
     if (!x) {
-        fprintf(stderr, "Unable to open %s. Giving up.\n", [patchFile UTF8String]);
+        fprintf(stderr, "Unable to open %s. Giving up.\n", [patchFile fileSystemRepresentation]);
         return 1;
     }
+    
+    SUBinaryDeltaMajorVersion majorDiffVersion = FIRST_DELTA_DIFF_MAJOR_VERSION;
+    SUBinaryDeltaMinorVersion minorDiffVersion = FIRST_DELTA_DIFF_MINOR_VERSION;
```
pornel: I think minor version is not needed, since it doesn't affect behavior. New, optional features can be added as new keys.

Sparkle/SUBinaryDeltaCommon.m
```diff
@@ -155,12 +198,41 @@ static void _hashOfFile(unsigned char* hash, FTSENT *ent)
...
+    NSFileManager *fileManager = [NSFileManager defaultManager];
+    NSDictionary *attributes = [fileManager attributesOfItemAtPath:path error:nil];
+    if (!attributes) {
+        return NO;
+    }
+    NSNumber *permissions = attributes[NSFilePosixPermissions];
+    if (!permissions) {
+        return NO;
+    }
+    mode_t newMode = ([permissions unsignedShortValue] & ~PERMISSION_FLAGS) | desiredPermissions;
```
pornel: How do you remove permissions?

Sparkle/SUBinaryDeltaCreate.m
```diff
@@ -138,11 +220,25 @@ int createBinaryDelta(NSString *source, NSString *destination, NSString *patchFi
         }
 
         NSDictionary *info = infoForFile(ent);
+        if (!info) {
+            fprintf(stderr, "Failed to retrieve info for file %s", ent->fts_path);
+            return 1;
+        }
         originalTreeState[key] = info;
+        
+        if (aclExists(ent)) {
```
pornel: That looks like an unrecoverable error. Maybe it's better to ignore or override ACLs?

Sparkle/SUBinaryDeltaCreate.m
```diff
@@ -138,11 +220,25 @@ int createBinaryDelta(NSString *source, NSString *destination, NSString *patchFi
...
+            fprintf(stderr, "Diffing ACLs are not supported. Detected ACL in before-tree on file %s", ent->fts_path);
+            return 1;
+        }
     }
     fts_close(fts);
+    
+    NSString *beforeHash = hashOfTreeWithVersion(source, majorVersion);
```
pornel: If I understand correctly, the beforeHash will include permissions. However, existing permissions aren't necessary to patch to be correct — they could simply be replaced with new values.

For that reason maybe it's best to ignore them when creating beforeHash?

Or even don't include them in the hash at all (which would simplify some logic). If you want to check them at all, then add a new, separate "afterPermissionsHash" that's built only from files permissions and is only checked against the final state.

pornel:
The code overall looks very good. Kudos for good error checking and new tests!

Sparkle/SUBinaryDeltaApply.m
zorgiepoo: File permissions changes can only happen:
a) After a binary diff is done, if permissions have changed
b) if the file's content does not change, but the permissions do change

They don't happen when a new file is extracted, because the new file's permissions from the xar will just be used. The confusion might stem from having to deal with delete and delete-then-extract properties.

Sparkle/SUBinaryDeltaApply.m
zorgiepoo: They might not be necessary, but it may be nice info to have. They can actually affect behavior though. For example, before this PR I made a couple backwards compatible changes for creating the diffs: case sensitive file renaming issue and a failing case where a directory became a symbolic link. For these cases, the minor version would be bumped (and in this patch I bumped it to 1.1 when creating version 1 patches). If I sent a version 1 patch to someone to apply, they can tell if the patch was created by version 1.0 or 1.1, which may be the difference for the patch applying correctly.

Thoughts?

Sparkle/SUBinaryDeltaCommon.m
zorgiepoo: Not sure I understand. Permission modes are replaced completely with what they are in the newer file, rather than being added/removed.

Sparkle/SUBinaryDeltaApply.m
pornel: Having a way to tell clients that are too old to give up is useful, but it's a binary decision: the patch will either work well enough to be acceptable, or won't.

> For example, before this PR I made a couple backwards compatible changes for creating the diffs: case sensitive file renaming issue and a failing case where a directory became a symbolic link. For these cases, the minor version would be bumped (and in this patch I bumped it to 1.1 when creating version 1 patches).

But it works fine without clients knowing about this. 1.0 client can't do anything with that information, so the patch applied by 1.0 must be correct by definition. 

1.1 client has a choice of using its 1.1 behavior (for which it doesn't need minor version information) or emulating 1.0, so that patch is applied in a bug-compatible way.

I find bug-emulation unappealing and try to avoid it, but if was necessary, it could be achieved by adding an optional key that's a flag `"do-this-weird-buggy-thing-that-1.0-did"` that newer versions emit and recognize.

>  If I sent a version 1 patch to someone to apply, they can tell if the patch was created by version 1.0 or 1.1, which may be the difference for the patch applying correctly.

How so? If there's possibility of patch being applied incorrectly, then it should be a major version change.

In case diff format v1.1 added a new optional feature, then v1.1 patch program could look for the feature's data, rather than at the version number.

e.g. if you add `xattr` support under a new key, then v1.0 patch will apply content changes, but won't change xattrs, and v1.1 patch will apply content and xattr changes, just by looking for the xattr key - it won't need to look at the version.

Sparkle/SUBinaryDeltaCommon.m
pornel: Ah, right. I've misread `~`.

Sparkle/SUBinaryDeltaCreate.m
zorgiepoo: I think this could be complicated. Let's take the scenario where you create a patch between two trees that both contain the same file. After creating the patch, you then alter the permissions of the file in the original tree and we're assuming this doesn't change the directory hash. When applying the patch to the original tree, BinaryDelta has no way to know to update the permissions of that file and we'll end up with an after-tree with the wrong permissions. We would have to do something like storing the permissions of every file in the diff that is not set to be extracted from the xar, and we wouldn't be able to mark any file as "not needed to patch" early on.

pornel: OK, I was assuming it'd be easy to have a separate list of permissions to apply.

Sparkle/SUBinaryDeltaApply.m
zorgiepoo: > How so? If there's possibility of patch being applied incorrectly, then it should be a major version change.

I should have rephrased this. The patch will apply correctly, but the patch may not have been created correctly. I was thinking of minor versions in cases where there's a bug in creating a patch but not a bug in applying a patch. In this case the major version would not change. A minor version is not needed as you say, unless a human (not the program) wants to know which version of BinaryDelta created the patch.

zorgiepoo: Other cases may include "oh this was created by version 1.2 so I know this patch was made more efficiently" or "I have BinaryDelta installed, I want to know which version I'm running and if I'm up to date to latest one"

zorgiepoo:
Thanks for reviewing the changes so far! I've just pushed a new commit adding an extract property, hopefully making the logic more clear of how applying the patch is done.

pornel:
Yup, it's clearer this way.

PR #527, opened 2015-04-19
BinaryDelta Verbosity/Improved Logging

zorgiepoo:
Just added more logging when creating/applying patches to show what the program is doing when --verbose is supplied (no logging if it's not except for errors), redirected logging messages to stderr, and added a version option to obtain the version of BinaryDelta running and the version from a specified patch file.

Diff screenshot


pornel:
:clap: 

Args parsing is getting a bit complicated. If you plan to add more options I recommend switching to `getopt`.

PR #529, opened 2015-04-24
Expose Sparkle symbols to unit test app

pornel:
SUInstaller and some other internal classes aren't public, so can't be tested directly.

Currently tests compile these files separately, but eventually this causes problems with duplicated symbols like SUError.


jakepetroules:
Yeah, turning it on for debug mode seems sufficient.

PR #534, opened 2015-04-26
Improved app update validation

zorgiepoo:
See discussion in issue #523

If an invalid DSA or invalid code signature is provided, the update will fail, regardless if one of these requirements is valid. The logic before is that the update will succeed if either a valid DSA or matched code signature is provided.

This still allows a code signature to change, be added, or be removed, as long as the DSA and the code signature (if provided) is valid. If no DSA is provided, then this is not possible, and the code signature from the host and update app must match.

I've not yet looked how feasible it'd be to create automated tests for this method, and also do not know how to run the pkg test with xctest.

[edit]: Shrug. It's also worth mentioning that code signature verification will be skipped if newBundlePath == nil, for example when the executable name changes as is the case with Sparkle's test application.

### DIFF
--- a/Sparkle/SUBasicUpdateDriver.m
+++ b/Sparkle/SUBasicUpdateDriver.m
@@ -228,8 +228,14 @@
         return NO;
     }
     
-    NSString *newBundlePath = [SUInstaller appPathInUpdateFolder:extractedPath forHost:self.host];
+    BOOL isPackage = NO;
+    NSString *newBundlePath = [SUInstaller installSourcePathInUpdateFolder:extractedPath forHost:self.host isPackage:&isPackage isGuided:NULL];
     if (!newBundlePath) {
+        SULog(@"No suitable install is found in the update. The update will be rejected.");
+        return NO;
+    }
+    
+    if (isPackage) {
         return canValidateByDSASignature;
     }
     
--- a/Sparkle/SUBasicUpdateDriver.m
+++ b/Sparkle/SUBasicUpdateDriver.m
@@ -249,7 +249,7 @@
     NSString *newPublicDSAKey = newHost.publicDSAKey;
     
     if (newPublicDSAKey != nil) {
-        if (![SUDSAVerifier validatePath:downloadedPath withEncodedDSASignature:DSASignature withPublicDSAKey:publicDSAKey]) {
+        if (![SUDSAVerifier validatePath:downloadedPath withEncodedDSASignature:DSASignature withPublicDSAKey:newPublicDSAKey]) {
             SULog(@"DSA signature validation failed. The update will be rejected.");
             return NO;
         }
--- a/Sparkle/SUBasicUpdateDriver.m
+++ b/Sparkle/SUBasicUpdateDriver.m
@@ -231,7 +231,8 @@
         return (DSASignature != nil);
     }
     
-    BOOL canValidateByCodeSignature = [SUCodeSigningVerifier hostApplicationIsCodeSigned] && [SUCodeSigningVerifier applicationAtPathIsCodeSigned:newBundlePath];
+    BOOL isAppCodeSigned = [SUCodeSigningVerifier applicationAtPathIsCodeSigned:newBundlePath];
+    BOOL canValidateByCodeSignature = isAppCodeSigned && [SUCodeSigningVerifier hostApplicationIsCodeSigned];
 
     if (!DSASignature && !canValidateByCodeSignature) {
         SULog(@"The appcast item for the update has no DSA signature. The update will be rejected, because both DSA and Apple Code Signing verification failed.");
@@ -241,7 +242,7 @@
     NSError *error = nil;
     
     if (DSASignature) {
-        if (canValidateByCodeSignature && ![SUCodeSigningVerifier codeSignatureIsValidAtPath:newBundlePath error:&error]) {
+        if (isAppCodeSigned && ![SUCodeSigningVerifier codeSignatureIsValidAtPath:newBundlePath error:&error]) {
             SULog(@"The application to update has an invalid code signature: %@. The update will be rejected.", error);
             return NO;
         }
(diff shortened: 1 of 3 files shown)


pornel:
What if the developer loses their DSA private key? Currently they can still update if they keep Developer ID the same.

Allowing code signing check to be skipped if bundle ID changes seems like a security vulnerability to me (bundle ID is trivial for the attacker to change, and it doesn't even protect the original bundle id, because the attacker can change it back again).

Please ignore how the test app works. It's insecure by design (we're serving the same update to every developer who runs it, which is exactly what should not be possible - it's an attack).

zorgiepoo:
If they lose their DSA private key, they won't provide DSA signature in their appcast item for the app. If they don't provide that, then DSASignature should == nil. As long as newBundlePath != nil, update should be able to validate via code signing. 

However if they want to change their DSA private signature they can't do so immediately with these changes; they'd have to provide an update with updated public key and new appcast but no DSA verification, and wait for the next update which would use the new DSA signature.

Correct me if any of this sounds wrong.

It's also worth mentioning in the current situation where people provide code signatures but no DSA key in their app can end up screwed. Can't be signed by multiple individuals or different code signature until next update with new appcast, can't delta update until next update.

zorgiepoo:
I have removed the appPathInUpdateFolder: method which made it difficult to tell if the install source path was found or if it found a package. If the install path is not found, we can simply fail, otherwise we can check if it's a package, or just continue on. The install path seems to be found for the test application now, so I'm not sure what issue I was running into before with newBundlePath being nil (or maybe I was just hypothesizing incorrectly before).

pornel:
> If they lose their DSA private key, they won't provide DSA signature in their appcast item for the app.

I've realized that this is a vulnerability (existing in Sparkle already). We have to assume that the appcast is untrusted and can be modified by an attacker, so if an attacker wants to bypass DSA validation, they can by hacking the appcast.

zorgiepoo:
Good point, this has been crossing my mind as well..

How about this:

If an application has a public key in their app, and the updated app has the same public key, then a DSA signature must be specified in the appcast and DSA validation will be done. The new code signing signature must be valid if it's present, but it can change identity or be removed if the developer loses access to their original.

If the public key changes, or ceases to exist in the new app (in case the developer loses their private key), then DSA validation won't be done regardless of what the appcast says. However, both apps must be code signed and the identity's of both have to be the same. Our basic code signing validation should be able to confirm that the new info.plist and the app's main resources (including the public key) came from the developer and not someone else.

Which is basically to say, if an app carries a public key, the developer is expected to provide a signature in their app cast. If they lose their private key, they generate a new public key from a new private key, or they stop DSA signing -- which is fine if the developer has been code signing all along.

zorgiepoo:
I pushed changes regarding to my last comment. The idea is that apple's code signing is reliable to tell if the info.plist and public DSA keys came from the developer and are in tact. I have not thoroughly tested this yet. A couple questions:

1) These changes rightfully make the current Test Application fail updating, because both of the apps aren't code signed, and the public DSA key is missing from the after updated app. Could you fix this by putting the SUPublicDSAKeyFile in the after-app and updating it on your server?

2) I avoided doing code signing validation on packages only because that is what the current code seems to be doing. I am also not very familiar with package updating. Why does the current code do this, and do we really want to avoid checking the code signature for packages?

Please let me know what you think of the overall idea of these recent changes..

zorgiepoo:
Alright, I can see that code signing works differently for installer packages and see why Sparkle may want to ignore them.

pornel:
Nice, the new logic looks clearer than the old one.

Sparkle/SUBasicUpdateDriver.m
```diff
@@ -221,27 +221,61 @@ - (void)download:(NSURLDownload *)__unused d decideDestinationWithSuggestedFilen
...
+    if (newBundle == nil) {
+        SULog(@"No suitable bundle is found in the update. The update will be rejected.");
         return NO;
     }
+    
+    SUHost *newHost = [[SUHost alloc] initWithBundle:newBundle];
+    NSString *newPublicDSAKey = newHost.publicDSAKey;
+    
+    if (newPublicDSAKey != nil) {
+        if (![SUDSAVerifier validatePath:downloadedPath withEncodedDSASignature:DSASignature withPublicDSAKey:newPublicDSAKey]) {
```
pornel: This should be using old DSA key. I know you're later checking if they're the same, but it's risky — it could create a vulnerability if the check later is ever removed/skipped.

pornel:
To sum it up, the logic is now:

If the update is a package, then it must be signed using DSA. No other verification is done.

If the update is a bundle, then it must satisfy one of:

 1. old and new DSA public keys are the same and valid (it allows change of Code Signing identity), or

 2. old and new Code Signing identity are the same and valid (allows change of DSA private key).

pornel:
The threat model for Sparkle is that attacker can replace appcast feed and new bundle with anything they want.

So here are possible attacks:

* They replace appcast to send a DSA-signed package installer. All security depends on DSA verification. Code Signing verification is bypassed.

* They replace public DSA key in an existing Code-Signed bundle and send appcast signed with their DSA key. All security depends on Code Signing protecting Info.plist. DSA verification is bypassed.

zorgiepoo:
Glad you like the changes!

We should also consider the cases where only one of the methods is being used. One gotcha I noticed is that if the old app *only* has a DSA public key, and the new app *only* has a valid code-signing signature, the update will fail. We could either just let it fail and make the developer ship the same DSA public key in the new app, or use the public DSA key in the old app to validate the new app from the signature in the app-cast in this special case.

Having an old app that is *only* code signed and the new app *only* has a DSA public key will fail and there's nothing we can do about that.

Also, it's worth mentioning this allows the old app's code signature to be invalid. Sparkle can still match the identity of the old signature against the new one however. Perhaps this is desirable in case the user or developer screw up their app in some way (not too uncommon).

PR #535, opened 2015-04-27
Properly import Foundation to fix errors when building as a module

tonyarnold:
I was seeing errors like

```
SUAppcast.h:19:18: Declaration of 'NSDictionary' must be imported from module 'Foundation.NSDictionary' before it is required
SUAppcastItem.h:16:28: Declaration of 'NSDate' must be imported from module 'Foundation.NSDate' before it is required
```

This PR addresses that, by importing the Foundation framework's umbrella header into `SUAppcast.h`.


jakepetroules:
Thanks!

PR #538, opened 2015-05-04
Only check if new path is code signed

zorgiepoo:
If the original app is signed, and the update is not, this is actually fine if the DSA validation goes through. And this may be desirable for people that don't want to pay cash to code sign their application anymore.

See issue #523 where this change was brought in.


pornel:
Thank you. That's a good idea. I'll merge that after 1.10 is released.

jakepetroules:
1.10 is released, shall we merge now?

PR #540, opened 2015-05-06
BinaryDelta Code Signed Extended Attributes Validation

zorgiepoo:
Apparently it is possible to code-sign any kind of non-mach-o binary file, such as a helper tool written in a scripting language that is marked as executable. The code signature information in this case is stored in the file's extended attributes. Unlike other kinds of extended attributes, these ones are important since they can invalidate the signature on an application if they're not present or if they're wrong, assuming these files are in a directory that should contain signed files.

The good news apparently is that Apple doesn't recommend using signed non-mach-o files inside an app and recommends inserting scripts in the app's Resources directory for example rather than making them executable and signing them. This means all we have to do is abort if we encounter files with a signed extended attribute, which is good because we don't preserve extended attributes in other circumstances anyway.

In the end, this pull request serves as a validation for not creating a diff when code-signed xattr files are around.

### DIFF
--- a/Sparkle/SUBinaryDeltaCommon.h
+++ b/Sparkle/SUBinaryDeltaCommon.h
@@ -16,6 +16,10 @@
 #define IS_VALID_PERMISSIONS(mode) \
     (((mode & PERMISSION_FLAGS) == 0755) || ((mode & PERMISSION_FLAGS) == 0644))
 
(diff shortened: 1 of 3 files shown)


pornel:
Thank you! 

I've ran into these xattr-signed scripts before (it's so annoying - the app works until you zip it)

Issue #452, opened 2014-11-07
Many “Unable to quarantine: 93” messages


codecaffeine:
I have a FooBar.app. When I autoupdate, I get a bunch of messages saying `Unable to quarantine: 93`. Is this anything I should be concerned about? This is occurring on OS X Yosemite with the latest 1.8.0 release from CocoaPods.

```
11/7/14 5:38:50.294 PM FooBar[13648]: Sparkle: ===== FooBar.app =====
11/7/14 5:38:50.823 PM FooBar[13648]: updater:0x608000240420 didFinishLoadingAppcast:<SUAppcast: 0x61000065a670>
11/7/14 5:38:50.824 PM FooBar[13648]: New Update Found: 1.8.8.1.9. Installing…
11/7/14 5:38:52.409 PM FooBar[13648]: Sparkle: Extracting /Users/matt/Library/Application Support/FooBar/.Sparkle/FooBar 1.8.8.1.9/FooBar.1.8.8.1.9.app.zip using 'ditto -x -k - "$DESTINATION"'
11/7/14 5:38:52.560 PM FooBar[13648]: updater:0x608000240420 willInstallUpdateOnQuit:<SUAppcastItem: 0x6080004c4ad0> immediateInstallationInvocation:<NSInvocation: 0x610000663f80>
11/7/14 5:38:52.562 PM FooBar[13648]: updater:0x608000240420 didCancelInstallUpdateOnQuit:<SUAppcastItem: 0x6080004c4ad0>
11/7/14 5:38:52.656 PM FooBar[13648]: updaterShouldRelaunchApplication:0x608000240420
11/7/14 5:38:52.657 PM FooBar[13648]: updater:0x608000240420 shouldPostponeRelaunchForUpdate:<SUAppcastItem: 0x6080004c4ad0> untilInvoking:<NSInvocation: 0x600000469600>
11/7/14 5:38:52.658 PM FooBar[13648]: updater:0x608000240420 willInstallUpdate:<SUAppcastItem: 0x6080004c4ad0>
11/7/14 5:38:52.000 PM kernel[0]: FooBar[13648] Unable to quarantine: 93
11/7/14 5:38:52.000 PM kernel[0]: FooBar[13648] Unable to quarantine: 93
11/7/14 5:38:52.000 PM kernel[0]: FooBar[13648] Unable to quarantine: 93
...
11/7/14 5:38:52.697 PM FooBar[13648]: updaterWillRelaunchApplication:0x608000240420
[...]

PR #541, opened 2015-05-09
Only remove quarantine with -[NSURL setResourceValue:] when it's present

zorgiepoo:
This should stop "Unable to quarantine: 93" messages from showing up in the console. Fixes issue #452


MaddTheSane:
Ah, good catch.
Sadly, when I wrote the code, 10.10 wasn't publicly released yet and I had no way to test it.

PR #560, opened 2015-06-24
pt-BR warning fix

ksuther:
Fixed a warning about the window being offscreen (introduced from the pt-BR localization update).


pornel:
IB is weird sometimes...

PR #561, opened 2015-06-28
Nullability

pornel:
Fixed some `null`/`const` warnings in Xcode 7.

Xcode 6 doesn't work on El Capitan, so I can't check if it's backwards-compatible.

### DIFF
--- a/Sparkle/SUGuidedPackageInstaller.m
+++ b/Sparkle/SUGuidedPackageInstaller.m
@@ -11,7 +11,7 @@
 
 #import "SUGuidedPackageInstaller.h"
 
-static BOOL AuthorizationExecuteWithPrivilegesAndWait(AuthorizationRef authorization, const char* executablePath, AuthorizationFlags options, const char* const* arguments)
+static BOOL AuthorizationExecuteWithPrivilegesAndWait(AuthorizationRef authorization, const char* executablePath, AuthorizationFlags options, char* const* arguments)
 {
 	sig_t oldSigChildHandler = signal(SIGCHLD, SIG_DFL);
 	BOOL returnValue = YES;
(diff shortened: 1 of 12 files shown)


Sparkle/SUGuidedPackageInstaller.m
```diff
@@ -55,11 +55,11 @@ + (AuthorizationRef)authorizationForExecutable:(NSString*)executablePath
 		const char* executableFileSystemRepresentation = [executablePath fileSystemRepresentation];
 		
 		// Prepare a right allowing script to execute with privileges
-		AuthorizationItem right;
-		memset(&right,0,sizeof(right));
-		right.name = kAuthorizationRightExecute;
-		right.value = (void*) executableFileSystemRepresentation;
-		right.valueLength = strlen(executableFileSystemRepresentation);
+        AuthorizationItem right = {
```
jakepetroules: This is causing the Travis build failure I think. Just revert this part of the PR.

pornel: It fails on the new `__nonnull` attribute. I'll shim it with an `ifdef`ed macro.

> /Users/travis/build/sparkle-project/Sparkle/Sparkle/Autoupdate/Autoupdate.m:154:35: error: expected ')'
        `[statusCtl setButtonTitle:SULocalizedString(@"Cancel Update", @"") target:nil action:Nil isDefault:NO];`

pornel: Ugh, it wasn't easy to make it backwards-compatible: http://stackoverflow.com/questions/31105433/how-to-use-objective-c-nonnull-in-a-backwards-compatible-way

PR #564, opened 2015-06-30
Add public DSA file to updated test app

zorgiepoo:
The test app will still need to be updated on Sparkle's server with an updated signed DSA signature in the appcast (http://sparkle-project.org/files/sparkletestcast.xml).

Would be good if Sparkle's test application works..


kainjow:
Should look into making the test app just run a local HTTP server. Something as easy as NSTask and `python -m SimpleHTTPServer 8080` from a directory with files could do the job.

zorgiepoo:
Seems like a good idea to do in future. I think git didn't add my Xcode project change by the way, so might need to add the file into the project..

pornel:
Uh, I can't find the private key that was used for this public key, so the keys need to be replaced.

Issue #562, opened 2015-06-29
DSA signature not working on master


PipoloyJo:
Hi,

I'm currently integrating Sparkle in my project and I found an issue with DSA signing on master.

It works fine on the last release (1.10) but seems to be broken on HEAD. I got this message:
____
Sparkle: Public DSA keys differ or are absent, and both apps are not code signed. The update will be rejected.
Sparkle: Error: An error occurred while extracting the archive. Please try again later. The update is improperly signed. (URL (null))
____
From what I can see from git log, it seems to be related to Zorg branch, starting with this commit 28ea47f796b2f433720a40b4b4903f844c4b6967

I don't have much time to investigate this issue, but if I can help let me know. 

Really appreciate your work, thanks guys!

Pipo

pornel:
Thanks for the report. 

Do you use DSA keys? Do you use Apple Code Signing?

PipoloyJo:
I'm using DSA keys.

zorgiepoo:
The relevant code is here: https://github.com/sparkle-project/Sparkle/blob/master/Sparkle/SUBasicUpdateDriver.m#L277

Which is to say, the code doesn't think that the DSA public keys in your old app and in your new app match, or the public DSA key is absent in one of these apps.

Can you verify that your DSA public key is properly specified in both the old and new app, and that they are the *same*?

PipoloyJo:
As I said earlier, DSA code signing worked with 1.10 release. Steps I followed:

- I generated two packages of my app with incremental version number, both packaged with the same public DSA key
- I installed the first one on my machine, and placed the other one on the server, with the DSA key and the appcast properly updated (and the private DSA signature)
- From the installed one I ran the "Check for Update..." check which properly found that a newer version of my app was available
- After that Sparkle was able to shutdown my app using the Autoupdate.app launch the new one and place it back in Applications folder. Smooth and easy.

With master I got stuck after downloading the update. Same code, same app, same key. I can try to rebuild the package with more debugging log, and see what changed. I've already started to look in the SUBasicUpdateDriver code.

zorgiepoo:
You will want to look why dsaKeysMatch BOOL is NO. (Is publicDSAKey nil? is newPublicDSAKey nil? Does [publicDSAKey isEqualToString:newPublicDSAKey] evaluate to NO?)

zorgiepoo:
[...]

PR #565, opened 2015-06-30
More verbose error message when DSA keys don't match

pornel:
To help diagnosing problems like #562

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -19,7 +19,7 @@ An easy-to-use software update framework for Cocoa developers.
 * True self-updating—the user can choose to automatically download and install all updates.
 * Displays a detailed progress window to the user.
 * Supports authentication for installing in secure locations.
-* Supports Apple code signing and DSA signatures for ultra-secure updates.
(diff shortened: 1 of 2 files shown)


zorgiepoo:
Left a couple of comments. Also, I don't think an application has to be code signed with a certificate issued from Apple. A certificate from someone else, or self-signed, should be able to go through the validation code too.

pornel:
> I don't think an application has to be code signed with a certificate issued from Apple

Do you mean the error messages should say code signing rather than apple code signing?

zorgiepoo:
Yeah.

pornel:
So in this case I've meant Apple's technology for Code Signing, as opposed to Sparkle's DSA way of signing code.

Do you have an idea how to call it to make it very clear that "code signing" is not the same as Sparkle's DSA signature?

zorgiepoo:
I personally don't associate code signing with Sparkle's DSA signatures.. Technically speaking, Sparkle only signs downloadable archives, not code. One possibility is to associate Sparkle's DSA signatures with archives and code signing with apps?

Seems to look good otherwise.

PR #566, opened 2015-07-01
Don't use main bundle if passed in bundle is nil

zorgiepoo:
This seems like a terrible idea..

### DIFF
--- a/Sparkle/SUHost.m
+++ b/Sparkle/SUHost.m
@@ -41,7 +41,6 @@ typedef struct {
 {
 	if ((self = [super init]))
 	{
-		if (aBundle == nil) aBundle = [NSBundle mainBundle];
         self.bundle = aBundle;
         if (![self.bundle bundleIdentifier]) {
             SULog(@"Error: the bundle being updated at %@ has no %@! This will cause preference read/write to not work properly.", self.bundle, kCFBundleIdentifierKey);
--- a/Sparkle/SUHost.m
+++ b/Sparkle/SUHost.m
@@ -41,6 +41,7 @@ typedef struct {
 {
 	if ((self = [super init]))
 	{
+        NSParameterAssert(aBundle);
         self.bundle = aBundle;
         if (![self.bundle bundleIdentifier]) {
             SULog(@"Error: the bundle being updated at %@ has no %@! This will cause preference read/write to not work properly.", self.bundle, kCFBundleIdentifierKey);


jakepetroules:
This stems from a very large commit made in 2009 which doesn't really explain why this was added. Also, I wonder if it's better to just assert if bundle is nil. SUHost is pretty useless without one.

@pornel, thoughts?

pornel:
This pattern is all over the place and it seems to be part of the public API

    + (SUUpdater *)updaterForBundle:(NSBundle *)bundle;

jakepetroules:
Perhaps we should document that nil == mainBundle in SUUpdater. I think we should add an assert in SUHost because it's private API and the two places in which it's used ensure or should ensure that it's non-nil (SUBasicUpdateDriver, Autoupdater)

kainjow:
This is common behavior in Cocoa, so I don't see why it's bad. Plus, it's a non-backwards compatible change. Maybe better to document it and leave it as it is.

jakepetroules:
A change to SUHost is backwards compatible, changing SUUpdater isn't. But I guess we could leave it as is either way.

zorgiepoo:
Added an assertion for the parameter.

SUHost isn't meant to be used for only the main bundle. The method is referenced at:

Autoupdate.m - https://github.com/sparkle-project/Sparkle/blob/master/Sparkle/Autoupdate/Autoupdate.m#L149

SUBasicUpdateDriver.m - https://github.com/sparkle-project/Sparkle/blob/master/Sparkle/SUBasicUpdateDriver.m#L258

SUInstallerTest.m - https://github.com/sparkle-project/Sparkle/blob/master/Tests/SUInstallerTest.m#L53

SUUpdater.m - https://github.com/sparkle-project/Sparkle/blob/master/Sparkle/SUUpdater.m#L99

In SUUpdater.m, it's just redundant and doesn't break any backwards compatibility due to the previous nil check up above. For the installer test, we can let it fail; if it's a problem somehow, we can fix it and we'll actually be aware of it now. In other cases, you *definitely don't* want to use the main bundle.

For Autoupdate.m we should probably add a check to make sure the bundle isn't nil. If it's nil, we can't do an install and shouldn't continue on (this could be a real problem).

Note that in general, something being common behavior in cocoa isn't a good case to make, especially now more apparent with Swift pointing some of these out.. Non-Optionals may have saved us here..

PR #568, opened 2015-07-06
Fix Lucida Grande is always used for release notes

xhacker:
Use "-apple-system-font" to use system default font.

### DIFF
--- a/Sparkle/SUUpdateAlert.m
+++ b/Sparkle/SUUpdateAlert.m
@@ -107,18 +107,16 @@
 
 - (void)displayReleaseNotes
 {
-    // Set the default font
-    [self.releaseNotesView setPreferencesIdentifier:SUBundleIdentifier];
+    self.releaseNotesView.preferencesIdentifier = SUBundleIdentifier;
     WebPreferences *prefs = [self.releaseNotesView preferences];
-    NSString *familyName = [[NSFont systemFontOfSize:8] familyName];
-    if ([familyName hasPrefix:@"."]) { // 10.9 returns ".Lucida Grande UI", which isn't a valid name for the WebView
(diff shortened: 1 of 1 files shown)


Sparkle/SUUpdateAlert.m
```diff
@@ -107,15 +107,10 @@ - (IBAction)remindMeLater:(id)__unused sender
...
-    // Set the default font
     [self.releaseNotesView setPreferencesIdentifier:SUBundleIdentifier];
     WebPreferences *prefs = [self.releaseNotesView preferences];
-    NSString *familyName = [[NSFont systemFontOfSize:8] familyName];
-    if ([familyName hasPrefix:@"."]) { // 10.9 returns ".Lucida Grande UI", which isn't a valid name for the WebView
-        familyName = @"Lucida Grande";
-    }
-    [prefs setStandardFontFamily:familyName];
-    [prefs setDefaultFontSize:(int)[NSFont systemFontSizeForControlSize:NSSmallControlSize]];
+    [prefs setStandardFontFamily:@"-apple-system-font"];
```
jakepetroules: Not `-apple-system`? (as implied by http://trac.webkit.org/changeset/185750)

xhacker: You are right `-apple-system` is the new recommended token. But it was added recently (https://bugs.webkit.org/show_bug.cgi?id=144484), so I believe we can’t use it in older systems.

jakepetroules: OK... please add a comment explaining the `-apple-system` and `-apple-system-font` tokens and why we're using it.

xhacker: Updated. Also did a little refactoring.

PR #569, opened 2015-07-07
Add delegate methods for pre-download and immediately post-failed-download.

vokal-isaac:
This enables modifying the download request (e.g., to add authentication credentials) and catching download-specific errors separately.


jakepetroules:
:+1: 

@pornel?

pornel:
Sure, why not?

vokal-isaac:
Thanks.

Out of curiosity, is there a planned release cycle of some sort?

pornel:
We don't have a fixed release cycle. If you'd like to use that change quickly I suggest testing version from master. We try to keep master stable.

vokal-isaac:
I've been using master for the past several days and it does seem stable and completely usable from my perspective.  I was mostly curious because it's slightly cleaner to use a tagged version through Cocoapods than to use master, but not a problem at all.

Thanks!

PR #570, opened 2015-07-10
Add script to generate a report comparing the Sparkle.strings files

kainjow:
This is a Swift script that generates an HTML report comparing the localized strings files to the English version. It will show keys that are missing, unused, and unlocalized.

Already one benefit is it's revealing a bug where the key "You're up to date!" needs to be changed to "You're up-to-date!" in some files to pull in the localization.

This could be expanded to run in a special mode where it could check for unused localizations and report an error, which could be ran from the `ci` target to catch bugs with localization changes, but if that were to happen, it may need to be rewritten in a language that works on older Xcode versions (Python may do).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,6 +28,5 @@ ci:
 	done
 
 check-localizations:
-	xcrun swiftc -sdk `xcrun --show-sdk-path` -o "$(TMPDIR)/CheckLocalizations" Sparkle/CheckLocalizations.swift
-	"$(TMPDIR)/CheckLocalizations" -root . -htmlPath "$(TMPDIR)/LocalizationsReport.htm"
+	./Sparkle/CheckLocalizations.swift -root . -htmlPath "$(TMPDIR)/LocalizationsReport.htm"
 	open "$(TMPDIR)/LocalizationsReport.htm"
--- a/Makefile
+++ b/Makefile
@@ -26,3 +26,8 @@ ci:
 			xcodebuild -sdk "macosx10.$$i" -scheme Distribution -configuration Debug test ; \
 		fi ; \
 	done
+
+check-localizations:
+	xcrun swiftc -sdk `xcrun --show-sdk-path` -o "$(TMPDIR)/CheckLocalizations" Sparkle/CheckLocalizations.swift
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -26,3 +26,8 @@ ci:
 			xcodebuild -sdk "macosx10.$$i" -scheme Distribution -configuration Debug test ; \
 		fi ; \
 	done
+
+check-localizations:
+	xcrun swiftc -sdk `xcrun --show-sdk-path` -o "$(TMPDIR)/CheckLocalizations" Sparkle/CheckLocalizations.swift
```
jakepetroules: remove

Makefile
```diff
@@ -26,3 +26,8 @@ ci:
...
+	"$(TMPDIR)/CheckLocalizations" . "$(TMPDIR)/LocalizationsReport.htm"
```
jakepetroules: ./Sparkle/CheckLocalizations.swift . "$(TMPDIR)/LocalizationsReport.htm"

Sparkle/CheckLocalizations.swift
```diff
@@ -0,0 +1,89 @@
+import Foundation
```
jakepetroules: #!/usr/bin/swift

Sparkle/CheckLocalizations.swift
```diff
@@ -0,0 +1,89 @@
...
+
+let args = NSProcessInfo.processInfo().arguments
```
jakepetroules: Can you easily access getopt_long from Swift? :D

Not a requirement, just would be nice.

jakepetroules:
This is very cool, I'm looking forward to seeing people use Swift more in scripting scenarios. The static typing and performance is a huge win over Python/Bash, not to mention the more powerful language features. And with it going open source later this year it can be reasonably used in cross platform projects. :)

If this is fast enough we could even run it as part of the normal build in Xcode and have warnings show up in the IDE.

jakepetroules:
Also, please try to use NSXMLNode in any case where input strings could inject malformed markup into the output.

kainjow:
When executed via the shebang, many extra args get passed in. It seems from a quick Google that getopt_long is usable, but it seems ugly. I'm going to try using NSUserDefaults.

PR #572, opened 2015-07-22
Declare WebPolicyDelegate, WebFrameLoadDelegate

pornel:


Sparkle/SUUpdateAlert.h
```diff
@@ -11,8 +11,9 @@
 
 #import "SUWindowController.h"
 #import "SUVersionDisplayProtocol.h"
+#import <WebKit/WebKit.h>
 
-@protocol SUUpdateAlertDelegate;
+@protocol SUUpdateAlertDelegate, WebPolicyDelegate, WebFrameLoadDelegate;
```
jakepetroules: What's the point of the forward declarations here? They're already present from WebKit.h.

pornel: I wasn't sure which SDK added them (I've started getting warning in 10.11), so they're for back-compat.

jakepetroules: You can't conform to a forward declared protocol though; that's a build warning (and should be an error TBH). Looks like these were informal protocols that got formalized in SDK 10.11, so you'll have to ifdef out the conformance if it's not available.

So... just include WebKit and add the protocol conformances in an ifdef SDK >= 10.11.

jakepetroules:
Obsoleted by other patch.

PR #573, opened 2015-07-22
Use bundleForClass instead of finding bundle by ID

pornel:
It should prevent confusion when multiple copies of Sparkle are available

### DIFF
--- a/Sparkle/SUBasicUpdateDriver.m
+++ b/Sparkle/SUBasicUpdateDriver.m
@@ -419,7 +419,7 @@
         [updaterDelegate updater:self.updater willInstallUpdate:self.updateItem];
     }
 
-    NSBundle *sparkleBundle = [NSBundle bundleWithIdentifier:SUBundleIdentifier];
+    NSBundle *sparkleBundle = self.host.sparkleBundle;
 
     // Copy the relauncher into a temporary directory so we can get to it after the new version's installed.
     // Only the paranoid survive: if there's already a stray copy of relaunch there, we would have problems.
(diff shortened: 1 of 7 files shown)


jakepetroules:
:+1:

zorgiepoo:
I don't think this change will quite work. SUHost, and plenty of the other code, is not necessarily called from an Application that uses Sparkle.framework. For example, SUHost.m, SUPlainInstaller.m, and -installationPath method are called from Autoupdate.app (and maybe unit testing) which doesn't use Sparkle.framework, and won't have SUNormalizeInstalledApplicationNameKey or SUEnableAutomatedDowngradesKey set in its bundle info.plist.

SUHost is ideally tied to an input bundle provided in its initializer. Could we find the Sparkle bundle *inside* the specified host's bundle instead of using `-[NSBundle bundleForClass:]`?

I wasn't aware that Sparkle had an info.plist of its own. It's in Sparkle's Resources, and so kind of surprises me that it can be located.

Additionally, does `SULocalizedString` have to change now?

pornel:
Oh, you're right. Thanks for catching that!

So I think solution to that would be to stop using Sparkle's own Info.plist - move these settings to compile-time constants (they're so unlikely to be changed that maybe even drop the setting completely)

For path _of_ Autoupdate.app it will work, so I'd keep bundleForClass for security there (we really really don't want OS to pick executables from the unverified bundle).

I haven't fixed SULocalizedString, because it's used in too many places, and hopefully wrong strings can't do serious damage.

zorgiepoo:
If we stop using sparkles info.plist we shouldn't even need or use sparkleBundle or SUBundleIdentifier.

PR #574, opened 2015-07-22
Don’t keep important settings in Sparkle’s Info.plist

pornel:
Re #573


zorgiepoo:
This seems like a good idea to me. `sparkleBundle` property in SUHost is completely useless/unused now (and technically wrong anyway). It should be removed.

Can we use `-[NSBundle bundleForClass:]` for `SULocalizedString()`? Just pick a class (e.g.: SUHost) that's shared between Sparkle.framework and Autoupdate.app. I presume Autoupdate.app has its own copy of string files.

zorgiepoo:
OK, I was wrong about `sparkleBundle` not being used. The code seems fine considering where `-userAgentString` and `-installWithToolAndRelaunch:displayingUserInterface:` is being called, but I am wary about `sparkleBundle` being a property of SUHost, when `[NSBundle bundleForClass:[self class]]` can just be executed outside.

pornel:
Autoupdate.app doesn't have its own strings files. I'd rather not split them, as this makes strings maintenance harder.

I don't mind moving the method away from SUHost, but to which class?

pornel:
OK, SUUpdater seems like a better fit for the bundle

zorgiepoo:
What are all the .strings files I see generated inside of Autoupdate.app for? I see that some of the resource files are copied over from Sparkle.framework to Autoupdate.app perhaps maybe to avoid this problem

If you grep for `[NSBundle bundleForClass:[self class]]` you can see it's already being used in `SUSystemProfiler.m` too. SUUpdater.m and SUBasicUpdateDriver.m would be to affected files for us.

pornel:
> What are all the .strings files I see generated inside of Autoupdate.app for

Oh, I didn't notice there were there. It's a dupe of framework's strings files, so hopefully it could be avoided? Seems wasteful.

> If you grep for [NSBundle bundleForClass:[self class]] you can see it's already being used in SUSystemProfiler.m

That class is currently independent of SUUpdater, and that call works, so I'd leave it as is.

zorgiepoo:
> Oh, I didn't notice there were there. It's a dupe of framework's strings files, so hopefully it could be avoided? Seems wasteful.

It may seem 'wasteful' but could help save us from this issue, which I think would be worth it.

michelf:
If it's too wasteful to have two copies, you can put only one copy in Autoupdate and have a relative symlink in the framework that points to the file in Autoupdate.

Issue #558, opened 2015-06-16
Does sparkle have a plan if deprecated functions are removed?


msphair:
specifically: `AuthorizationExecuteWithPrivilages`?

zorgiepoo:
The modern way is to use SMJobBless API to perform privileged operations. This has a few consequences though. One is that it requires a helper tool that must be in the app (it can't be in the Sparkle framework). Another is that it requires code signing and I believe for the identity to be specified in the app's info plist. I believe the API copies the helper tool somewhere in /System (which also can handle updating the tool), so every Sparkle based app would have their own helper tool installed somewhere into there. This would probably only be necessary if a privileged operation needs to occur though (which isn't the common case). I think Sparkles operations are simple enough where we don't need any fancy kind of IPC.

Side note: the authorization and file moving/copying code needs to be somewhat rewritten IMO. A bunch of deprecated APIs are being used (evident if you increase the deployment target), unnecessary app copies are done, update installations can be incomplete, many bits of code can be replaced with a Cocoa equivalent.

pornel:
My overall vague plan is to rearchitect Sparkle from "update yourself" approach to "update that bundle". 

When Sparkle can update apps that it's not running in, then it'll be easy to make it a helper app.

But, it's hard, e.g. lots of things are tied to delegates or `SUHost`, so it'll need to be a gradual process with lots of small refactorings. Patches welcome :)

zorgiepoo:
I'm currently working on a rewrite of `SUPlainInstallerInternals` that addresses some of the issues I mentioned above.. Will follow up on progress soon.

zorgiepoo:
I took a look at `SMJobBless()` a bit which is the only modern/correct way of doing privilege escalation today (replacement for `AuthorizationExecuteWithPrivilages()`). [EvenBetterAuthorizationSample](https://developer.apple.com/library/mac/samplecode/EvenBetterAuthorizationSample/Introduction/Intro.html) is the relevant & modern code to look at. In summary, from what I've gathered:

* The host app (Autoupdate.app in our case?) and helper tool both need an Info.plist entry for what code signing certificate should be used. They both also need the bundle IDs set up uniquely & together; thus setting up Sparkle with this cannot be a simple drag & drop into an Xcode project. Both the app and helper tool need to also be code signed. The helper tool needs to reside at /Contents/Library/LaunchServices/ inside an app, so we definitely don't want to put it directly inside a framework (again, AutoUpdate could possibly suffice).
[...]

PR #575, opened 2015-07-23
Rewrite file operations for updating an app

### LINKED ISSUE #498: How does the updater handle Install on Quit for large apps?
If my application is 1.5gb and I run install on quit - I then quit the application and immediately turn off my computer. Could I end up with a corrupt install? Meaning, if it takes the machine 5 minutes to update, but I turn off the machine before it completes the update, does Sparkle have any way to prevent corruption?

zorgiepoo:
The improvements we want to achieve:
* Stability - No more corrupt / incomplete updates by ensuring atomic move operations
* Efficiency - Faster updates because of reduction of file copies that need to be done
* Separation from Host and Helper - No usage of SUFileManager in the host application
* Cleanness - Minimized deprecated APIs (no FSPathMakeRef, FSGetCatalogInfo, FSFindFolder, etc), use modern Cocoa APIs when possible, robust error handling, strong documentation, easier to read code.

Many other small notes:
* SUFileManager does not have class methods. Methods across an instance share an AuthorizationRef, instead of having to authorize every time a privileged operation needs to be done
* The odd logic of `while (authStat == errAuthorizationDenied)` was removed
* Instead of checking if we have write access, we try to invoke a normal operation first. If that doesn't work, with the right error returned back, we try authorizing, which is what Apple recommends doing
* The trash is not used as a temporary location anymore. Rather we now request for a temporary directory to be created. This is more likely to succeed too (eg: on a drive that has no trash folder)
* The old method for releasing the quarantine did not work until I used the "com.apple.quarantine" key instead of the LS constant that was being used.
* For the old method for releasing the quarantine, we now check if a file has the quarantine xattr bit set before removing it. Just like how the new method works.
* Quarantines can now be removed with authorization
* The result of calling temporaryNameForPath: was never actually used in the old code
* -[NSString fileSystemRepresentation] which can throw an exception is not used. Only getFileSystemRepresentation:maxLength: is used and the error on it is always checked
* The desperate/hopeless check for if the code signature for the update is valid after the installation is complete is removed

Potential issues that still needs to be looked in to:
* ~~Currently the method to move a file to the trash only works on 10.8+~~
* ~~Appending of version number / picking destination name of app moved to trash~~

--

This is still a work in progress but most of the work is done. This may improve #441 and #558, and would resolve #498. This pull request is definitely a big change / should be reviewed carefully.


Sparkle/SUBasicUpdateDriver.m
```diff
@@ -420,17 +418,30 @@ - (void)installWithToolAndRelaunch:(BOOL)relaunch displayingUserInterface:(BOOL)
...
     // Copy the relauncher into a temporary directory so we can get to it after the new version's installed.
     // Only the paranoid survive: if there's already a stray copy of relaunch there, we would have problems.
     NSString *const relaunchPathToCopy = [sparkleBundle pathForResource:[[sparkleBundle infoDictionary] objectForKey:SURelaunchToolNameKey] ofType:@"app"];
     if (relaunchPathToCopy != nil) {
         NSString *targetPath = [self.host.appCachePath stringByAppendingPathComponent:[relaunchPathToCopy lastPathComponent]];
-        // Only the paranoid survive: if there's already a stray copy of relaunch there, we would have problems.
+        
+        NSFileManager *fileManager = [[NSFileManager alloc] init];
+        
+        BOOL shouldAbortUpdate = NO;
```
pornel: Instead of using a state variable, could you move the code to a function, so `if (!shouldAbortUpdate &&` becomes `if (![self doSomeStuff] && [`?

This code is fine, but I'm afraid such variables can cause trouble later if somebody moves the code around.

zorgiepoo:
Added commits to handle < 10.8 compatibility, appending the version to the older app's destination name, and making strings more user friendly. All work is pretty much done for review except for merging the recent `sparkleBundle` lookup code.

Sparkle/SUPlainInstaller.m
```diff
@@ -7,13 +7,165 @@
...
+    
+    return trimmedVersion.length > 0 ? trimmedVersion : nil;
+}
+
++ (BOOL)performInstallationToURL:(NSURL *)installationURL fromUpdateAtURL:(NSURL *)newURL withHost:(SUHost *)host error:(NSError * __autoreleasing *)error
+{
+    SUFileManager *fileManager = [[SUFileManager alloc] init];
+    
+    // Create a temporary directory for our new app that resides on our destination's volume
+    NSURL *tempNewDirectoryURL = [fileManager makeTemporaryDirectoryWithPreferredName:[installationURL.lastPathComponent.stringByDeletingPathExtension stringByAppendingString:@" (Incomplete Update)"] appropriateForDirectoryURL:installationURL.URLByDeletingLastPathComponent error:error];
```
pornel: Since we're installing a single item, having a temp directory seems unnecessary.

Could you just move `/tmp/foo.app` to `destination/foo (incomplete vX).app`? or maybe to `.foo (incomplete vX).app` (with a `.` prefix to prevent temp file from popping up in Finder)

Sparkle/SUPlainInstaller.m
```diff
@@ -7,13 +7,165 @@
...
+    } else {
+        oldDestinationName = oldURL.lastPathComponent.stringByDeletingPathExtension;
+    }
+    
+    NSString *oldDestinationNameWithPathExtension = [oldDestinationName stringByAppendingPathExtension:oldURL.pathExtension];
+    
+    // If the installation and old URL differ, the sooner we can install our new app
+    // because we won't have to move the old app out of the way first
+    // increasing our chances of having it installed in case something goes wrong
+    if (![installationURL isEqual:oldURL]) {
```
pornel: I'm worried that two codepaths complicate things here, and the case where names differ is not common. Could you simplify that to one codepath? (without an `else` that does a lot)

e.g.

1. rename `/tmp/foo.app` to `dest/foo (new)`
2. rename `/dest/foo` to `/dest/foo (old)`
3. rename `/dest/foo (new)` to `/dest/foo`

where new/old suffixes could be versions or even random strings if needed.

Sparkle/SUPlainInstaller.m
zorgiepoo: Creating a temp directory for this kind of operation is consistent with how some of the Cocoa APIs work. For example, the docs for replaceItemAtURL:withItemAtURL:backupItemName:options:resultingItemURL:error allows this, but wouldn't allow what you're suggesting. URLForDirectory:inDomain:appropriateForURL:create:error: may even do this as a last ditch effort. They are used for single files (e.g., documents).

Parking the temp directory in the same directory as the final destination is a last resort effort that won't be hit often. Of course we wouldn't want the user to see the incomplete directory visible, but if it's in a temporary location and the installation is interrupted, it may be cleaned out after a couple days or next reboot. Further, there might be some safety implications with having the temp directory be at an unpredictable/hashed location.

I'd rather use a provided Cocoa/System API for creating a temp file/directory where it's viable... I don't think this bit is much different from the old code, except it preferred using the trash instead of a temp location, which is worse.

Sparkle/SUPlainInstaller.m
zorgiepoo: The code path that assumes they can be equal can be used for both cases. So I just did that. Like you said, when they differ is not a common case.

Sparkle/SUPlainInstaller.m
zorgiepoo: Plus I just realized immediatly parking at the final destination directory would mean we may have to shell out to privileged operations there for copying and quarantine removal, when it may not be needed. The less we have to shell out to  privileged operations, the better.

Sparkle/SUFileManager.m
```diff
@@ -0,0 +1,708 @@
...
+    }
+    
+    return YES;
+}
+
+// Wrapper around getxattr()
+- (ssize_t)getXAttr:(NSString *)nameString fromFile:(NSString *)file options:(int)options
+{
+    char path[PATH_MAX] = {0};
+    if (![file getFileSystemRepresentation:path maxLength:sizeof(path)]) {
```
pornel: Wouldn't it be safer to use `[file fileSystemRepresentation]`?

zorgiepoo: No, that can throw an Obj-C exception and doesn't return a copy.

pornel:
Merged. Sorry for keeping it open for so long.

zorgiepoo:
It's cool since it's a big sort of change. I'll look into creating some unit tests for this class shortly.

PR #578, opened 2015-07-25
Rewrite test app so it works again, and locally

zorgiepoo:
This creates and signs the update on the fly, so
we can alter the test app without too much hassle.
This also starts up a local web server.
Refs #564


Sparkle.xcodeproj/project.pbxproj
```diff
@@ -1842,6 +1870,18 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = FA1941CB0D94A70100DD942E /* ConfigTestAppDebug.xcconfig */;
 			buildSettings = {
+				CLANG_ENABLE_OBJC_ARC = YES;
```
jakepetroules: No build settings in pbxproj -- put these in ConfigTestApp.xcconfig

TestApplication/English.lproj/InfoPlist.strings
```diff
@@ -1,4 +1,3 @@
 /* Localized versions of Info.plist keys */
 
-NSHumanReadableCopyright = "© Andy Matuschak, 2006";
-CFBundleShortVersionString = "1.5";
\ No newline at end of file
+NSHumanReadableCopyright = "© Andy Matuschak, 2006";
```
jakepetroules: Add newline at EOF

TestApplication/SUTestApplicationDelegate.m
```diff
@@ -0,0 +1,184 @@
...
+@interface SUTestApplicationDelegate ()
+
+@property (nonatomic) SUUpdateSettingsWindowController *updateSettingsWindowController;
+@property (nonatomic) NSTask *serverTask;
+
+@end
+
+@implementation SUTestApplicationDelegate
+
+#define UPDATED_VERSION @"2.0"
```
jakepetroules: Make this a static const var unless there's actually a reason to make it a define (like overriding in build settings).

TestApplication/SUTestApplicationDelegate.m
```diff
@@ -0,0 +1,184 @@
...
+    if ([[mainBundle objectForInfoDictionaryKey:(NSString *)kCFBundleVersionKey] isEqualToString:UPDATED_VERSION]) {
+        NSAlert *alreadyUpdatedAlert = [[NSAlert alloc] init];
+        alreadyUpdatedAlert.messageText = @"Update succeeded!";
+        alreadyUpdatedAlert.informativeText = @"This is the updated version of Sparkle Test App.\n\nDelete and rebuild the app to test updates again.";
+        [alreadyUpdatedAlert runModal];
+        
+        [[NSApplication sharedApplication] terminate:nil];
+    }
+    
+    NSFileManager *fileManager = [[NSFileManager alloc] init];
```
jakepetroules: `-[NSFileManager defaultManager]`

TestApplication/SUTestApplicationDelegate.m
```diff
@@ -0,0 +1,184 @@
...
+        NSLog(@"Failed to copy main bundle into server directory with error %@", copyBundleError);
+        assert(NO);
+    }
+    
+    // Update bundle's version keys to latest version
+    NSURL *infoURL = [[destinationBundleURL URLByAppendingPathComponent:@"Contents"] URLByAppendingPathComponent:@"Info.plist"];
+    
+    assert([fileManager fileExistsAtPath:infoURL.path]);
+    
+    NSMutableDictionary *infoDictionary = [[NSMutableDictionary alloc] initWithContentsOfURL:infoURL];
```
jakepetroules: You can get this from `[[NSBundle bundleWithURL:destinationBundleURL] infoDictionary]`

TestApplication/SUTestApplicationDelegate.m
```diff
@@ -0,0 +1,184 @@
...
+    infoDictionary[(NSString *)kCFBundleVersionKey] = UPDATED_VERSION;
+    infoDictionary[@"CFBundleShortVersionString"] = UPDATED_VERSION;
+    
+    assert([infoDictionary writeToURL:infoURL atomically:NO]);
```
jakepetroules: Why *not* use an atomic write?

jakepetroules:
Overall, good work. I like the decoupling of the test app and the web site. Less to update.

TestApplication/SUTestApplicationDelegate.m
zorgiepoo: I will have to write to `infoURL` anyway though, and was trying to stay away from having a NSBundle be cached in memory.

TestApplication/SUTestApplicationDelegate.m
zorgiepoo: It doesn't matter if the file write is incomplete/interrupted. If the write fails, we bail out, and the server directory is cleaned out the next time the program runs.

KenThomases: I haven't checked the rest of the project if there's some policy about asserts being always enabled, but you should never put code with important effects in `assert()`. If asserts were disabled, the file would not be written. Same just below, too.

zorgiepoo: Ah, that's true. I'd need to remove/change the asserts then.. Or just make sure asserts are always on for the test app.

zorgiepoo: From my testing, It seems like asserts are on in both debug and release mode of the test app, so it should be fine. To my understanding, the only sensible reason to disable assert is for performance reasons.

zorgiepoo:
Made the updated changes. I had to re-enable the implicit synthesize warning because I wasn't able to figure out how to turn it off for only the test application. I presume it's on for rest of Sparkle's code due to legacy reasons.. (I actually prefer private ivars the most, but didn't want to muck with settings).

TestApplication/SUTestApplicationDelegate.m
jakepetroules: assert should be disabled in release mode. If it isn't, this is a bug in the build settings. Please fix this.

zorgiepoo: Assertions are not disabled from release mode from my testing when archiving the framework and the test application. But why would they be? This is not a bug; it's a design decision to have it enabled or disabled.

It's simply untrue that assertions shouldn't be used for important effects. The whole point of using it is when your program has a bug and is in a corrupted state if a certain condition cannot be satisfied, and thus can't recover nor should be able to continue on. Removing assertions means you remove safety checks. And the only real reason to do that is for performance gains. But neither this test application, nor Sparkle's framework, have any performance critical code I'm aware of that could actually benefit from doing that; so it would just be a loss.

jakepetroules: Assertions should not be used for important side effects, as such disabling them should have no effect on the execution of the program. Instead of:

    assert([infoDictionary writeToURL:infoURL atomically:NO]);

Use:

    BOOL ok = [infoDictionary writeToURL:infoURL atomically:NO];
    assert(ok);

zorgiepoo: I can change that kind of code, but it doesn't seem too relevant to the discussion.

> Assertions should not be used for important side effects, as such disabling them should have no effect on the execution of the program

I cannot understand why you say this. Assertions clearly have an effect if left enabled, when conditions cannot be satisfied. When they are disabled, your program ignores errors and is unsafe. Assertions are used when there is a bug in the program. There is always some kind of side effect if such a condition cannot be satisfied (either assertions are enabled, and your program crashes, or program continues in a corrupt state and you crash later or something even worse happens later on).

zorgiepoo: OK, I will pull out the conditions from the assert().. I see the issue with that now. Enabling/Disabling assert is a different issue though.

Issue #546, opened 2015-05-18
Move Autoupdate.app out of Resources


zorgiepoo:
Helper Apps are not supposed to be in Resources. As it is currently, Autoupdate.app is in Resources and it can't or shouldn't be signed because it's in a directory that isn't supposed to contain signed files. It should be moved into the Executables directory or a Helpers directory according to the code signing guide. The former should be easier because Xcode doesn't provide a UI option for the latter.

After it's moved out to another location, the code that finds the Autoupdate.app path needs to slightly change. Then Sparkle being added to a new project that code signs Sparkle on copy needs to be tested, and verified that Autoupdate.app along with Sparkle's executable are both signed.

I would submit a pull request for this, but I've not had much luck with the last step because either something is screwed up with the project or I ran into an issue I didn't know how to resolve.

kainjow:
I've used "Contents/Helpers" for helper apps in the main app bundle with success, but never done it with a framework. Can a framework support a "Helper" subdirectory? I don't see any information about this on Apple's website. If Autoupdate were to be stored in the main app bundle at Contents/Helpers, that would require a second Copy Files Phase and further complicate Sparkle's integration. However Sparkle could (and probably should) add checks at launch to see if Autoupdate exists at its required location so devs could fix the issue before even attempting to test Sparkle.

I'd first check to see if it would work inside the framework's Helper folder (and code signing passes and the app runs properly on 10.7-10.11), otherwise a move to the main bundle's Helpers folder may be necessary.

zorgiepoo:
According to [Table 3](https://developer.apple.com/library/mac/technotes/tn2206/_index.html) of Apple's code signing article, the Autoupdate app can be in either a Helpers directory or the Executables directory (the latter which Xcode has a UI option for). Both are equally valid places.

Autoupdate need not be stored in the main app's bundle. We should be able to ship it inside of Sparkle's framework.

The problem is when I try changing the destination of Autoupdate to Executables, set a code signature for the test application, and try to build Sparkle's test application I receive this error:

```
CodeSign /Users/msp/Library/Developer/Xcode/DerivedData/Sparkle-fcwvfssfpluezdektysonoqniiqr/Build/Products/Debug/Sparkle\ Test\ App.app/Contents/Frameworks/Sparkle.framework/Versions/A
    cd /Users/msp/Desktop/Sparkle
[...]

PR #584, opened 2015-07-27
For Debug builds, disabling timestamps in code signing means the buil…

danielpunkass:
…ds can take place without a network connection.

### DIFF
--- a/Configurations/ConfigCommonDebug.xcconfig
+++ b/Configurations/ConfigCommonDebug.xcconfig
@@ -10,3 +10,7 @@ SPARKLE_EXTRA_DEBUG = -DDEBUG
 OTHER_CFLAGS = $(SPARKLE_EXTRA_DEBUG)
 ONLY_ACTIVE_ARCH = YES
 COPY_PHASE_STRIP = NO
+
+// For Debug builds, we don't require timestamping because Apple's
+// server may be down or we may be off-network
+OTHER_CODE_SIGN_FLAGS = --timestamp=none


jakepetroules:
Smart!

zorgiepoo:
I highly am against this. Sparkle shouldn't be built and code signed at once first off (it should be dragged in an application target and code signed on copy). And what if someone does try to sign a debug build and use it somewhere if even by accident? The time stamp is important. If you really want to be building without an Internet connection you should either not be signing at all or be using a self signed certificate.

danielpunkass:
@zorgiepoo Since as you say it's relatively unlikely that any shipping app would not re-sign the framework anyway, it only seems to diminish the importance of jumping through hoops to make sure the signature (especially for a Debug build) is perfect.

For my purposes it would serve the same purpose to turn off code signing all together on Debug builds, but my philosophy has usually been that it's a good idea to maintain as close a parity between Release and Debug as reasonably possible.

I have used this trick of disabling timestamp support for Debug builds for a long time and never paid a price for it. It's not just a pain when you're not online for whatever reason, it also leaves you prone to surprise failures when Apple's service happens to be down or slow. I wrote more about this here FWIW if anybody is interested. http://indiestack.com/2014/01/timestamp-disservice/

zorgiepoo:
Valid point that this is less severe in the common case if the framework is re-signed, but still not something that should be completely ruled out.

More importantly, why should Sparkle, a framework, which is supposed to be left unsigned in an application's repo, include details about code signing? On the one-time cirumstance someone needs to sign the framework from *this* project, why should we risk disabling the time stamp?

I maintain an open source project and the way I get around it is by using a self signed certificate. Other apps like Textual do it this way too. The advantage is that it doesn't require someone to pay a $100 fee to Apple, though it is a little more set up. However does not have the close parity you mention.

danielpunkass:
I guess this comes down to how serious a risk people feel it is that a signed "Debug" copy of the framework could be distributed. I don't understand well enough how the the timestamp ties an app to its certificate, and what the realistic consequences of shipping something with that "flawed" signature would be.

It seems to me that if somebody is shipping a Debug copy of Sparkle, AND not re-signing the binary in their final app, then they have other problems than we can help them wtih by reversing this commit ;)

I'm happy to keep the change in my own fork of Sparkle though, if others share @zorgiepoo's concern. It's supremely easy to merge, I just thought I would share it as a convenience to others using the project.

zorgiepoo:
The implications of disabling time stamp are described here (great read): http://lists.apple.com/archives/apple-cdsa/2013/Mar/msg00018.html

I worry about the situation where someone builds/signs a debug version of Sparkle and uses it in their custom build environment without knowing (And not unlikely, building debug is easier in termed of Xcode UX).

As the engineer states, this is a question of permanence not build type. I don't think we should be giving off that you shouldn't distribute such builds. They state it is our call to make whether it is a risk or not, and I personally think we shouldn't risk it, not as a library author.

danielpunkass:
Thanks for the link. That does indeed make a concise explanation of the reason for the timestamp and the risks of omitting it.

Botton line for others following along: a built Sparkle framework signed --timestamp=none will carry its code signature and be treated as trusted by code signing validation so long as the certificate (Developer ID) used to sign it does not become compromised and then invalidated by Apple. Using the timestamp gives Apple the information needed to permanently trust your code based on its age, even if the identity is later compromised.

Personally I still don't think it's a big deal to generate less permanent/failsafe code signatures on Debug builds of Sparkle, but I can see where @zorgiepoo is coming from with his concern.

pornel:
@zorgiepoo The e-mail you quoted contains

> Every signature that does not have such a timestamp will become invalid upon revocation.

As far as I understand that means it fails closed. If you don't timestamp and then revoke, then the non-timestamp build is revoked too.

So I don't see a problem with this.

zorgiepoo:
To my understanding, the point of the timestamp is to ensure that your builds before your certificate was revoked are still valid.

They mention the case if it's compromised by somebody else, but do we really want to guess what may happen when a developer stops paying for their certificate or when they (forget to) renew it? It's possible their builds may be invalid and that's not a good thing.

The decision to disable timestamp is up to the developer. So my opinion is we should respect Xcode's defaults and not disable it as a framework distributor.

zorgiepoo:
By the way, I would be pretty curious to learn why you aren't able to use the "Code Sign on Copy" feature @danielpunkass

danielpunkass:
It's been a while since I tried to use Code Sign on Copy. My memory is it had to do with not being able to fine-tune the settings for how some items get code signed. This is particularly important when you are piecing together libraries from different projects where maybe you don't want to or can't adjust the settings in the project itself

zorgiepoo:
Ok. All I can do is take your word for it. You may also be interested in issue #546 too (which is not easily fixable at the moment for the project as a whole). I think if there are real issues with code signing, we should continue file rdar's to Apple... Sigh.

Regardless, I still think it should be the responsibility of the application developer to specify how they may want to sign libraries they include.

Also, I hesitate to associate "Debug" with "should (or should not) be used this way." Again, what a debug build means may depend on the developer, as it did with you when you decided to turn on code signing.

Re-reading from the mailing list, you notice that Perry (who implemented code signing) makes no connection to debug builds either: 

> > Indeed, --timestamp=none is my workaround.  But I leave this setting *permanently* for *Debug* builds.  Is there any reason why I would ever need a codesigned timestamp in a Debug build?

> It's really not a question of build type. It's a question of permanence...

> > But I wonder why a default project in Xcode is not configured with --timestamp=none in Debug.  Seems like an obvious solution to me.

> Xcode can't know what you're going to do with the output, and it won't make these decisions for you. It errs on the side of caution

And so I believe we too should err on the side of caution, and not cause potential surprises going against Xcode's default.

danielpunkass:
Ahhhhhh. This explains a lot: I assumed that Code Signing was enabled by default in Sparkle. I submitted the Debug-only code signing flags to alleviate what I perceived as a dependency Sparkle had on the timestamp service.

Looking more closely now I see that code signing was on for the Debug build of my fork as a historical artifact. I tend to agree that as long as we don't need to code sign in the library (e.g. to support entitlements or something) then it also doesn't make sense to specify built-in code signing flags.

I'm going to create a new pull request to revert the changes and recommend we do that.

Thanks for talking through this. I didn't get you were approaching this from the beginning as being something I was opting (knowingly) into doing.

zorgiepoo:
> Thanks for talking through this. I didn't get you were approaching this from the beginning as being something I was opting (knowingly) into doing.

No problem, but even if code signing was enabled by default, I probably would have had the same stance with regards to the timestamp service ;). Typically framework authors should try to avoid code signing if they can, especially if they can be used in non-code signed projects, which Sparkle still can be. 

It's totally possible in the future there may be some opt-in code-signing required features which could complicate matters further.. we'll see.

PR #585, opened 2015-07-27
Remove unused temporaryName: parameter and supporting code.

danielpunkass:
The temporaryName: parameter to SUPlainInstaller's copyPathWithAuthentication... method was not being used internally but was still mucking up the (already long) method signature for the method and related methods up the chain. Also remove related, no-longer-used temporaryNameForPath method.


pornel:
Good catch!

zorgiepoo:
also noted in #575

PR #586, opened 2015-07-27
Finesse the display text for information-only updates.

danielpunkass:
(Sorry for the onlsaught of pull requests today. I'm cleaning house and just trying to get as sync'd with your fork as possible. I think this is about the last one for a while :) )

Finesse the wording of the message displayed in SUUpdateAlert when the item is "information only."

Previously when an information-only update is displayed to the user, it uses the inappropriate text "Would you like to download it now?" Since the alert is modified in this case to provide a 'Learn More' button, then text should match the action that is being offered. In this commit I break the localized string up into parts such that the notice about the updated version is the same in either case, but the action offered differs depending on the type of update being presented.

Note that this pull request only represents changes to the English Sparkle.strings file. Until and unless localized versions of Sparkle.strings add translations for the new broken-up strings, they will display this information in English only.


danielpunkass:
Meant to add on this one that I don't know what you normally do when a pull request affects other localizations? Should it stay in pull request, pending until all necessary localizations are amended to make it complete?

jakepetroules:
> (Sorry for the onlsaught of pull requests today. I'm cleaning house and just trying to get as sync'd with your fork as possible. I think this is about the last one for a while :) )

This is not a fork, it's the official Sparkle. ;)

Sparkle/SUAppcastItem.m
```diff
@@ -55,6 +55,11 @@ - (BOOL)isCriticalUpdate
     return [self.propertiesDictionary[SUAppcastElementTags] containsObject:SUAppcastElementCriticalUpdate];
 }
 
+- (BOOL)isInformationOnlyUpdate
+{
+	return (([self infoURL] != nil) && ([self fileURL] == nil));
```
jakepetroules: Style:

    return self.infoURL && !self.fileURL

Sparkle/SUUpdateAlert.m
```diff
@@ -231,7 +231,20 @@ - (NSString *)descriptionText
 	else {
         [self.versionDisplayer formatVersion:&updateItemVersion andVersion:&hostVersion];
     }
-    return [NSString stringWithFormat:SULocalizedString(@"%@ %@ is now available--you have %@. Would you like to download it now?", nil), [self.host name], updateItemVersion, hostVersion];
```
jakepetroules: tabs => spaces for this block

Sparkle/SUUpdateAlert.m
```diff
@@ -231,7 +231,20 @@ - (NSString *)descriptionText
...
+
+	// We display a slightly different summary depending on if it's an "info-only" item or not
+	NSString* actionSentence = nil;
```
jakepetroules: attach * to variable name.

Sparkle/SUUpdateAlert.m
```diff
@@ -231,7 +231,20 @@ - (NSString *)descriptionText
...
+
+	if ([self.updateItem isInformationOnlyUpdate])
```
jakepetroules: Use dot notation

Sparkle/SUAppcastItem.m
zorgiepoo: This style is mixed in Sparkle's project (and before in this file uses != nil too). Personally I think implicit checking can potentially remove type safety in certain circumstances.. Just my 2 cents.

jakepetroules: For methods returning BOOL, this can be quite a gotcha, true (I usually use `BOOL b = !!ptr`). However this is a lot more readable. At the very least, remove the ( ) and [ ] if you want to keep the == nil

zorgiepoo: Just to add: in a situation like `if (!number)`, you can't tell offhand if it's a NSNumber object, int*, or int. If you explicitly compare to != nil, you know offhand. If you change the type of the variable during refactoring, a C compiler can warn you that you need to change this code. I come from the standpoint where implicit testing should only be used for "boolean" (not pointers or integers) types. Languages like haskell and java enforce this. So, IMO, I find explicit testing more readable.

jakepetroules:
Please fix the other translations. Even without understanding the language in question, it should be relatively easy to extract the translated substring corresponding to "Would you like to download it now?"

danielpunkass:
Hmm. I guess you're right. OK, it's on my TODO list.

danielpunkass:
It looks like this approach will work OK for most of the localizations but may break down with Arabic, where the existing translation incorporates one of the template placeholders into the sentence in a less typical location:

```
لديك الإصدار %3$@ من تطبيق %1$@، هل ترغب بتزيل الإصدار %2$@ الآن
```

I guess that leaves things in a state where either I need to rethink the breaking down of the localized strings, or else somebody with Arabic language knowledge can chime in to say whether it can be reasonably changed to fit the pattern. For all languages except Arabic, this template works:

```"%@ %@ is now available--you have %@. %@" = "%1$@ %2$@ is now available (you have %3$@). %4$@";
```

Where %4$@ is "Would you like to download it now?" or "Would you like to learn more about this update on the web?". In the Arabic translation from above, it seems the "Would you like to download it now?" portion of the phrase incorporates the "now available" version number.

danielpunkass:
So, probably I was just being too clever, and should change to two fully written template strings: one for the "download now" case and one for the "learn more" case.

jakepetroules:
Yes, that's usually the thing to do. I didn't notice at first glance that they were concatenated.

danielpunkass:
Left out a code update from the last commit. Will update shortly.

pornel:
For translations, I think for now it may be better to keep the translated "would you like to download" part rather than to cut it and replace with English (i.e. IMHO it's better to present an understandable sentence with a slight inaccuracy, than a glued-on sentence in English, which user may not understand).

PR #588, opened 2015-07-28
Revert "For Debug builds, disabling timestamps in code signing means …

danielpunkass:
Reverts sparkle-project/Sparkle#584

The rationale behind Pull Request #584 was based on my flawed assumption that Sparkle code signs by default in its build settings. I was wrong about that. Since code signing is not done by default, I don't think it makes sense to bother with providing any default code signing options.


jakepetroules:
Yes. And I merged the original PR a little preemptively...

PR #591, opened 2015-08-02
Update layout constraints when removing release notes

### LINKED ISSUE #369: Layout does not collapse when release notes are hidden
Views that are hidden still participate in autolayout, so hidden release notes just leave a massive gap in the window :(

screen shot 2014-07-07 at 01 37 55

zorgiepoo:
We also add height constraints to the automatically install & download updates button. Since we don't have a custom layout view, we also shouldn't call -[NSView setNeedsLayout:]
Fixes #369


jakepetroules:
Thanks!

PR #592, opened 2015-08-03
Update with all the available mac models until July 2015

GabrielUlici:


kainjow:
IMO, I'd resubmit this so all those pulls into your repo don't show up here. Would be cleaner.

PR #596, opened 2015-08-15
Improvements to Autoupdate application

zorgiepoo:
* Do all the installation work after the runloop is set up (hope this helps #441)
* TerminationListener only does termination listening now..
* Renamed folderpath to updateFolderPath. Likewise with some other variables. Since these are passed by command line arguments, updateFolderPath cannot be nil, so those checks for if folderpath is nil in the old code is removed
* Removed pointless global variable access [SUInstaller updateFolder]
* Old code does not handle if host path is not installation path *and* host path is not desired executable path. This does.

--

Would appreciate if the test app pull request is merged! I pretty much have to start with that branch in order to make and test any new changes.

### DIFF
--- a/Sparkle/Autoupdate/Autoupdate.m
+++ b/Sparkle/Autoupdate/Autoupdate.m
@@ -9,160 +9,154 @@
 #include <unistd.h>
 
 /*!
- * If the Installation takes longer than this time the Application Icon is shown in the Dock so that the user has some feedback.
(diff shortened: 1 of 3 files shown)


zorgiepoo:
I don't think the "coverage" really changed at all, so don't let that check frighten much..

pornel:
Looks good!

Issue #543, opened 2015-05-14
Make DSA Signatures Mandatory


zorgiepoo:
I strongly believe that DSA signatures should be required.

From my observations:

**Code Signing** ensures the _System_ can verify that the content inside the application came from the developer.

**DSA Signatures** ensures _Sparkle_ can verify that the archive is downloaded exactly as the developer created it.

To show how these are different, one needs to realize that code signing _intentionally_ allows changes to the app that doesn't break the application's signature.

Specifically, removing localization files, removing architectures from a mach-o binary, and allowing code inside the application to be replaced with other code signed by the same developer. The last point (which I quickly tested myself) implies someone could take an old executable and insert/replace it into a new application's update.

In the past, code signing also allowed resource rules to exclude certain files from being checked. Code signing also presently allows file permission changes (rdar://20835014).

Currently, validation code ignores DSA signatures if applications are signed (see issue #534). Sparkle also _does not_ perform a deep code signature validation of an application, which makes Sparkle not being able to detect plenty of bad changes (see issue #523).

Further, even if we could perform a deep check of an app's signature, we may not want to because of how CPU intensive/time consuming it can be for large applications. Ironically, I believe that unsigned applications that use Sparkle are the ones that have a more secure updating process.

Not requiring a DSA signature, or at least a public key in the app, is also inconsistent with how Sparkle deals with other mechanisms (packages, system prefpanes, whatever else), and it's also a burden if anyone wants to transition to delta updates which we should encourage more use towards in the some-time future.

It is true requiring a DSA signature is slightly more inconvenient when making an update. But if it's a whole lot more inconvenient, this just means the developer hasn't automated their update creation process. And this could even be thought of as a favor for making them do that..

Lastly, this is not to say basic code signing validation is completely useless, since it may catch cases where the application isn't archived or unarchived properly that may go otherwise unnoticed. But thinking of DSA signatures as the "old" or not recommended way as this wiki suggests seems to be quite dangerous (and in many cases inefficient) to me.

Thoughts?

zorgiepoo:
Alternatively, serving over https might work as well as an alternative over Sparkle's DSA signatures. That might even be a little more secure since it would also prevent an attacker inserting newer updates from a beta appcast into a stable one for example. I have the feeling Sparkle handled https at one point but doesn't do that anymore.

pornel:
Thanks for starting the discussion. You make good points.

[...]

PR #599, opened 2015-08-19
Deprecate serving over http without DSA

zorgiepoo:
This deprecates only using Apple's code signing for validation. See
issue #543


pornel:
:+1:

PR #602, opened 2015-08-23
Fix Japanese localization

1024jp:
In Japanese localization, strangely enough, Apple uses normal three dots instead of the single ellipsis symbol.

### DIFF
--- a/Sparkle/ja.lproj/Sparkle.strings
+++ b/Sparkle/ja.lproj/Sparkle.strings
@@ -37,13 +37,13 @@
 
 "Cancel Update" = "アップデートを中止";
 
-"Checking for updates..." = "アップデートを確認しています…";
+"Checking for updates..." = "アップデートを確認しています...";
 
 /* Take care not to overflow the status window. */
-"Downloading update..." = "アップデートをダウンロードしています…";
+"Downloading update..." = "アップデートをダウンロードしています...";
 
 /* Take care not to overflow the status window. */
-"Extracting update..." = "アップデートを展開しています…";
+"Extracting update..." = "アップデートを展開しています...";
 
 /* the unit for gigabytes */
 "GB" = "GB";
@@ -51,14 +51,14 @@
 "Install and Relaunch" = "インストールして再起動";
 
 /* Take care not to overflow the status window. */
-"Installing update..." = "アップデートをインストールしています…";
+"Installing update..." = "アップデートをインストールしています...";
 
 /* the unit for kilobytes */
 "KB" = "KB";
 
 /* Alternative name for "Install" button if we have a paid update or other update
 	without a download but with a URL. */
-"Learn More..." = "詳しい情報…";
+"Learn More..." = "詳しい情報...";
 
 /* the unit for megabytes */
 "MB" = "MB";


jakepetroules:
Is this possibly just a mistake or is there a specific reason for Japanese to differ from other languages in not using the ellipsis symbol?

michelf:
Curious, I sampled Japanese localized strings in various Apple apps (iTunes, Numbers, GarageBand, Calendar, Automator), and I notice isn't entirely consistent. Sometime Apple is using an ellipsis character, sometime three consecutive dots. Surprisingly often: both mixed in the same file. Using three dots seems indeed more prevalent than the ellipsis character though. Not sure why.

I could guess is that some translators are a more lenient than others when it comes to typographic details such as this one. Or perhaps automatic tools to catch and correct these don't work with Japanese. I can see a bit of the same pattern in Korean, although for Korean using the proper ellipsis character seems to be in majority.

jakepetroules:
My guess it it's translators being lazy or not knowing there's a dedicated ellipsis character especially given that it's as inconsistent as you say. I would lean towards not merging this.

1024jp:
I suppose Apple consciously does this. 

I don't know the true reason. There is no official documentation about that. But to guess, it because the ellipsis symbol in Japanese is not drawn on the baseline but at the center of the line. This is the standard behavior by Japanese font and Japanese text. 

screen shot 2015-08-23 at 14 58 53
 
Actually, the ellipses in the system aren't drawn like that since they provide a special font only for the system UI (named Aqua Kana --Users can't use this font, for instance, from the font panel). 

Most of 3rd parties don't follow this rule, furthermore, as @michelf pointed out, even Apple's software sometime don't. Because it looks, in fact, almost the same. But I don't believe Japanese localization by Apple 100%. Japanese localization by Apple is not always the top-quality-localization (sadly).

OK, I show some interesting examples.

The following is a sample screenshot I made to show how Apple uses ellipses in Japanese localization:

screen shot 2015-08-23 at 15 09 10
__"Start Dictation" menu item in Edit menu__

The above two menu items were made by me. I just put the same sentence, but with different ellipses. The first one is the ellipsis symbol and the second one is with three dots. The third one is an auto-inserted menu item by Apple. You can easily see Apple uses three dots here.

The second example is a indicator window on Finder.

screen shot 2015-08-23 at 11 07 47 copy

Here, you can see a ellipsis again but in a different scenario from the first one. It shows something is processing. Also in this case, Apple uses the three dots. It's actually pretty hard to recogzise, but you can see if you compare with the following picture. Both are, of course, drawn in the System Font.

screen shot 2015-08-23 at 15 28 52

 Apple do the same in other system-level UI or in other core applications like Finder. Thus, I suppose this is the true specification of OS X anyway.


To summarize, in my opinion, at least the ellipses which represent an additional input requirement should be three dots in order to harmonize with other OS X system interface, since it is not a normal literal ellipsis but a symbol of other action. 
But you can just ignore this pull-request, because a lot of OS X applications doesn't follow this rule and there is almost no difference. The 99.999% users don't notice this :P

michelf:
There is a separate Unicode character called MIDLINE HORIZONTAL ELLIPSIS when you need it. But it's true that many Japanese fonts will put the regular ellipsis character midline anyway. If the regular ellipsis character happens to be centered because of the font they are using, that could explain why Japanese localizers would use regular dots, even though the system font does not behave like this.

pornel:
That's an interesting find. If Apple is not consistent with it, I suggest keeping consistency within Sparkle and using ellipsis character.

1024jp:
Ok, I understand your stance.
Thanks for considering my request 🍻

PR #614, opened 2015-09-04
Check for empty strings (as well as nil) in SUHost's -name method

kolpanic:
NSBundle's `-objectForInfoDictionaryKey:` returns an empty string instead of nil for `CFBundleDisplayName` in our case (at least on El Capitan Beta).

### DIFF
--- a/Sparkle/SUHost.m
+++ b/Sparkle/SUHost.m
@@ -103,13 +103,13 @@
 
     // Allow host bundle to provide a custom name
     name = [self objectForInfoDictionaryKey:@"SUBundleName"];
-    if (name && ![name isEqualToString:@""]) return name;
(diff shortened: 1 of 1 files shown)


zorgiepoo:
Probably better to check name's length isn't 0, instead.

kolpanic:
@zorgiepoo Good idea

zorgiepoo:
Also you don't have CFBundleDisplayName defined anywhere (in the info.plist file or any .strings files for example)? If you truly do not specify CFBundleDisplayName anywhere, and the method isn't returning nil, please file a bug report to Apple because it's not obvious why it'd return an empty string.

jakepetroules:
Can't hurt

PR #615, opened 2015-09-05
Improve Unarchiver Tests

zorgiepoo:
* All supported compression formats are now tested including dmg, which
has a license agreement included
* The hash of the extracted directory is now checked, which encompasses
every file's existence, its contents, and its permissions.
* Re-written using Swift

---

Currently it's using Swift 2. Let me know if it'll have to use 1.2 instead, at the moment.


pornel:
> OS X targets using Swift cannot be built against an SDK older than 10.10, but the effective SDK is 10.9.
** TEST FAILED **

zorgiepoo:
Xcode for me says it's set at "No SDK (Latest OS X)" - I'm not extremely familiar with the build configurations to know what is supposed to change. Or if it's something you have to change on travis.

pornel:
I think the Makefile is trying to run on all SDKs since 10.7: https://github.com/sparkle-project/Sparkle/blob/master/Makefile#L24

Try changing it to 10.10+

jakepetroules:
> I think the Makefile is trying to run on all SDKs since 10.7: https://github.com/sparkle-project/Sparkle/blob/master/Makefile#L24

Yes, I did that because we claim to support Xcode 5+. Ideally I'd like to keep the project as a whole buildable with the 10.7 SDK, if we can think of a good way to easily exclude the tests from SDKs < 10.10.

pornel:
In Build Settings for Sparkle Unit Tests target set SDK to 10.10?

jakepetroules:
That won't work. I pushed something that should...

jakepetroules:
> Currently it's using Swift 2. Let me know if it'll have to use 1.2 instead, at the moment.

We probably should wait to merge since that'll require Xcode 7. Might take Travis a while to support it, too.

pornel:
OK, that 637eaae9c722774e84686b43c3f232e65340c593 makes more sense :)

zorgiepoo:
Xcode 7 is now officially out, as well as Swift 2. Travis also can [support Xcode 7](http://blog.travis-ci.com/2015-09-09-xcode7-gm/) now. I think this is now or very close to being merge-able?

pornel:
Can you update travis config file to enable xcode7?

zorgiepoo:
Updated to use xcode7 and fixed CI building

jakepetroules:
Just put SWIFT_OPTIMIZATION_LEVEL settings in the global Debug/Release files; no need to create specific ones for the ConfigUnitTest target.

zorgiepoo:
I moved them into the global configs, although I'm not exactly sure how the unit test configs inherit from them.

PR #618, opened 2015-09-18
Fix CHANGELOG

1024jp:
I suppose the changes about 1.12.0 in the CHANGELOG are just the changes which were added after some point of 1.11.0rc release and thus are belong to 1.11.0.

I've just removed `# 1.12.0` header and bring the change about the code signing and DSA verification (probably the most important change in 1.11.0) to the first position. No change in the sentences themselves.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,8 @@
-# 1.12.0
+# 1.11.0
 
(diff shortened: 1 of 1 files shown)


pornel:
Thanks

PR #620, opened 2015-09-18
Add automatically-install-updates option

statico:
Sparkle is great. Since we're in beta and we want to keep our beta testers as up-to-date as possible, I wanted a flow that would check for updates and install them automatically when the application is launched. This change does that.

```objective-c
  SUUpdater* updater = [SUUpdater sharedUpdater];
  [updater setDelegate:self];
  [updater installUpdatesIfAvailable];
```

The above checks for updates interactively using the Sparkle progress windows. If an update is found, it is automatically downloaded, installed, and the application is restarted. The application can respond to `updaterDidNotFindUpdate` as usual, or do something when `userDidCancelDownload` (a new SUUpdaterDelegate method) is called.

cc @mlogan @bkcore


statico:
Also, I'm not sure if this is functionality that the Sparkle project wants. If it is, I'll extend the PR with tests if possible.

jakepetroules:
> Also, I'm not sure if this is functionality that the Sparkle project wants.

Yeah, I think it's a valuable feature. Perhaps this might have been frowned upon in the past but this technique is used by many popular applications today and as long as it's user-configurable I think it's great.

zorgiepoo:
How is this better than the Automatic Updates feature Sparkle currently provides (see SUAutomaticUpdateDriver)? I actually think this is better because it installs an update on termination without the user noticing. I feel like I'm missing something here.

statico:
@jakepetroules Cool. I'll make some tests, then.

@zorgiepoo We're using Sparkle in a wrapper application whose purpose is solely to check for updates, respond to custom URI schemes, and launch the main application which is a giant cross-platform blob. I didn't think the SUAutomaticUpdateDriver lets us check for updates and install them while showing a progress/feedback window.

zorgiepoo:
Right, SUAutomaticUpdateDriver covers a case for those wanting to do automatic non-intrusive in-the-background updates..

So if I understand correctly, your changes would allow an app, after it checks for updates, to automatically download the update, show the user UI install progress & allow them to cancel the download, and once the update is downloaded & unarchived to automatically relaunch the application without user input?

This sounds like it's designed in a way such that updates should only be checked on launch, and that the updater's automaticallyChecksForUpdates should *never* be *YES*

Personally I'm not too much of a fan of this, and the scenario you give about having a wrapper app doesn't sound like a common case to me. Adding mostly-unused features can have hidden maintainability costs. Wonder what @pornel thinks.

statico:
@zorgiepoo Correct, that's the use case and we set `setAutomaticallyChecksForUpdates` to false.

pornel:
I see it may be useful for frequently-updated betas/nightlies, but it has a potential to be very annoying.

The fact that you need to invoke 3 methods makes it seem like a hack. Perhaps the API should be something like:

``objc
[sparkle installUpdatesIfAvailable];
```

that does an immediate check + install, regardless of other settings.

pornel:
@statico I think the functionality is useful, but a better API would be [just a command that does check+install on demand](https://github.com/sparkle-project/Sparkle/pull/620#issuecomment-141840609). Would you be interested in changing the code to do that?

statico:
@pornel Sure, I could do that. It'll take me a few days

statico:
@pornel API updated to a single method as suggested, `[updater installUpdatesIfAvailable];`. Please take another look.

pornel:
Thank you

zorgiepoo:
Should the `updater:didFindValidUpdate:` delegate method be called? This patch is skipping it.

Issue #318, opened 2014-01-30
Updated application isn't active after relaunch


vincedev:
I use Sparkle to provide updates for my application.
The update process just fine, except for one thing : after clicking "Install and Relaunch", the app relaunches but isn't active.

I wonder if someone has observed this behavior and if it is due to Sparkle or not ?
Looking at the code for 'finish_installation', I found that the app is relaunched using `NSWorkspace` `-openFile:` method, which should do the job just fine.

Also the application is correctly active on a simple launch.

PR #623, opened 2015-09-22
Keep Auto Updater alive for small delay before launching app

zorgiepoo:
This increases the likelihood that the updated launched app will
be brought up front and active.

---

Sparkle has an issue where it may not launch the updated application and have it be up-front and active. This issue (at least for me) is easily reproduced by Sparkle's Test Application. Launching it with Finder (via double click) may be more likely to show the issue. Also this may not be something that can be reproduced 100% of the time, but I think it can be common and I've seen it happen in one of my own apps. See also issue #318 for it being reproduced elsewhere.

I've looked into a few other ways to do this, but ultimately couldn't find anything that worked as well as delaying the termination of the auto updater by an arbitrary time - the intent being not to terminate before the new app is "sufficiently" launched.


zorgiepoo:
By the way, the reason why launching AutoUpdater.app doesn't have this issue I'm guessing is probably because it calls -`[NSApplication activateIgnoringOtherApps:]` on launch, however we don't have that luxury when launching another developer's app.

pornel:
Sorry, conflicted with other PR :(

I must admit I don't quite understand the solution. Is that some OS X rule that the parent process must exists to raise window, or do you think it's an OS X bug/coincidence that this is needed?

zorgiepoo:
My observation and thoughts from another developer is that OS X is plagued with activation issues. Indeed we have other activation issues and workarounds in Sparkle as well. This issue in particular may be some sort of timing issue with the updater deactivating and the opened app activating.

PR #627, opened 2015-09-25
Don't show Autoupdate dock icon if we shouldn't show UI

zorgiepoo:
I think this has been an issue for quite a while even before my refactoring of Autoupdate.m. And to note it looks like automatic updates hasn't been tested very much, which isn't good.


pornel:
Good catch. Those icons are spooky.

PR #629, opened 2015-09-25
Don't install automatic updates if the system might shut off

zorgiepoo:
Abort automatic updates before terminating the app if the system is about to shut or log off. This is particularly important for automatic updates because in response to the system shutting down and terminating our app, our response shouldn't be to spawn another application/process.

Incidentally, trying to install an automatic update while the system is powering off is a scenario that can very likely lead to incomplete/corrupt updates (unless #575 is incorporated) -- all it requires is users to hit a check box. But this isn't really about incomplete updates (nor can this issue hope to fully solve that), it's about doing less work so the user can power off faster.

If we get the power off notification, we let the delegate know about it. It may be possible some other application could cancel the shut down / log off process, so we should be able to continue even if that happens.


pornel:
That's awesome!

PR #631, opened 2015-09-25
Make requirements more explicit

kainjow:


jakepetroules:
Make note that OS X 10.7 is the runtime requirement and note the OS X SDK required as well (I think it's 10.7).

kainjow:
Xcode 5 only includes the 10.8 SDK, so that should be implied?

jakepetroules:
Not necessarily. The version of Xcode (more specifically, the version of the compiler and other build tools) is separate from the SDK requirement.

Issue #624, opened 2015-09-22
Automatic Updates with Insufficient Write Permissions


zorgiepoo:
I don't think this has been given too much thought, but if you have automatic updates enabled, then Sparkle tries to install the update when the app is terminating. If the app doesn't have permission to write (e.g, non-admin user updating an app in /Applications/), then AutoUpdate will prompt for an authorization dialog requesting for permission after the app has been terminated (normally where we wouldn't expect Sparkle to show any UI at all).

Few options I can think of approaching this:
1. Do nothing - leaves user confused why an updater is asking for a password
2. Make an educated guess by looking at the app's ownership, so that we never allow automatic updates to happen if we don't have sufficient write permissions before authorizing
3. Make an educated guess and prompt on app termination if the user wants to install the update - a bit intrusive in my opinion - and what about machine being powered off?
4. Make an educated guess, keep the update around, and automatically install on next launch - potentially tricky since we may have to validate the update again, and not sure if it's safe to do this without the host app's knowledge.

Maybe there are some other options I'm missing.. This should probably be addressed somehow anyway.

pornel:
In the short term I think option 2 makes sense. I assume "not allowing automated updates" means doing updates, but displaying UI anyway.

pornel:
BTW: the UI has "automatically install in the future". If the app is write-protected, then:

a) don't display the checkbox?

b) change file owner/write permissions to make auto updates possible?

zorgiepoo:
I don't think b) is viable. For a) we can simply use NSFileManager's isWritableAtPath: method. What is really the difficult part hiding the check box. I assume we should be doing that if the developer specifies SUAllowsAutomaticUpdatesKey to false anyway?

gaocan1992:
How can I change file owner/write permissions to make auto updates possible?
I have a similar issue [#1219](https://github.com/sparkle-project/Sparkle/issues/1219)

PR #633, opened 2015-09-27
Disallow automatic updates when user can't write to the bundle

zorgiepoo:
The automatic updates checkbox is also hidden in this scenario, and
we also update the layout constraints if the releasenotes are still visible

Fixes / works around issue #624

### DIFF
--- a/Sparkle/SUHost.m
+++ b/Sparkle/SUHost.m
@@ -62,6 +62,17 @@
     return [self.bundle bundlePath];
 }
 
+- (BOOL)allowsAutomaticUpdates
+{
+    // Can we automatically update in the background without bugging the user (e.g, with a administrator password prompt)?
+    BOOL isWritableAtBundle = [[NSFileManager defaultManager] isWritableFileAtPath:self.bundlePath];
+    
+    // Does the developer want us to disable automatic updates?
+    NSNumber *developerAllowsAutomaticUpdates = [self objectForInfoDictionaryKey:SUAllowsAutomaticUpdatesKey];
+    
+    return isWritableAtBundle && (developerAllowsAutomaticUpdates == nil || developerAllowsAutomaticUpdates.boolValue);
+}
+
 - (NSString *)appCachePath
 {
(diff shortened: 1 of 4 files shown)


Sparkle/SUHost.m
```diff
@@ -62,6 +62,17 @@ - (NSString *)bundlePath
     return [self.bundle bundlePath];
 }
 
+- (BOOL)allowsAutomaticUpdates
+{
+    // Can we automatically update in the background without bugging the user (e.g, with a administrator password prompt)?
+    BOOL isWritableAtBundle = [[NSFileManager defaultManager] isWritableFileAtPath:self.bundlePath];
+    
+    // Does the developer want us to disable automatic updates?
+    NSNumber *developerAllowsAutomaticUpdates = [self objectForInfoDictionaryKey:SUAllowsAutomaticUpdatesKey];
```
pornel: Maybe reverse order would be slightly faster, avoiding disk access in some cases?

```
if (developerAllowsAutomaticUpdates) return developerAllowsAutomaticUpdates.boolValue;
return …isWritableFileAtPath…;
```

Sparkle/SUUpdater.m
```diff
@@ -435,8 +435,8 @@ - (void)setAutomaticallyDownloadsUpdates:(BOOL)automaticallyUpdates
 
 - (BOOL)automaticallyDownloadsUpdates
 {
-    // If the SUAllowsAutomaticUpdatesKey exists and is set to NO, return NO.
-    if ([self.host objectForInfoDictionaryKey:SUAllowsAutomaticUpdatesKey] && [self.host boolForInfoDictionaryKey:SUAllowsAutomaticUpdatesKey] == NO) {
```
pornel: Removed repetition :+1:

Sparkle/SUHost.m
zorgiepoo: Done. It's only faster in the case if the developer disables automatic updates.

PR #635, opened 2015-09-27
Update modification & access time for new update

zorgiepoo:
If we don't do this, Finder may not report the correct file size for the new update.
The system also may not register the new app in its database if we don't do this.

We may have been able to get away with not doing this before because we used to perform a copy
to install the update. However, now updates in some cases can be done with only move operations.

This should also be an alternative way of addressing issue #508

### DIFF
--- a/Sparkle/Autoupdate/Autoupdate.m
+++ b/Sparkle/Autoupdate/Autoupdate.m
@@ -172,6 +172,7 @@ static const NSTimeInterval SUTerminationTimeDelay = 0.5;
                        versionComparator:[SUStandardVersionComparator defaultComparator]
                        completionHandler:^(NSError *error) {
                            if (error) {
+                               SULog(@"Installation Error: %@", error);
                                if (self.shouldShowUI) {
                                    NSAlert *alert = [[NSAlert alloc] init];
                                    alert.messageText = @"";
--- a/Sparkle/SUFileManager.h
+++ b/Sparkle/SUFileManager.h
@@ -80,6 +80,19 @@
 - (BOOL)changeOwnerAndGroupOfItemAtRootURL:(NSURL *)targetURL toMatchURL:(NSURL *)matchURL error:(NSError **)error;
 
 /**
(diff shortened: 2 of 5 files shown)


zorgiepoo:
@pornel I'd look into resolving this issue before doing a release otherwise apps may not install properly.

jakepetroules:
Instead, call `LSRegisterURL` (https://developer.apple.com/library/prerelease/mac/documentation/Carbon/Reference/LaunchServicesReference/index.html#//apple_ref/c/func/LSRegisterURL) - this is what Xcode itself does when you build an app.

zorgiepoo:
I initially tried to use `LSRegisterURL` (and it was tried in issue #508) too, but it didn't work well for me in practice. Notably, the application usually didn't update when I was looking at it in Finder. Also, I'm not sure if it's an issue here, but `LSRegisterURL` may not work on items that aren't applications, so touching the modification date is a bit more versatile.

As described in in [Apple's article](https://developer.apple.com/library/mac/documentation/Carbon/Conceptual/LaunchServicesConcepts/LSCConcepts/LSCConcepts.html):

> After making any significant change in an application’s Launch Services–related information, you should either reregister the application explicitly, by calling LSRegisterFSRef or LSRegisterURL with inUpdate set to true, or update the modification time of the application to ensure that it will be updated by the automatic registration utilities described above.

> Note: You can update an application’s modification time using the BSD touch command in a Terminal window. For example, the command touch /Applications/TextEdit.app sets the modification time of TextEdit to the current time.

Updating the modification date is still a legitimate and documented way of doing it. I have some comments in the code mentioning this.

jakepetroules:
Hmm, fair enough.

zorgiepoo:
By the way, Xcode doesn't show the test app's icon half the time I build & run after cleaning, so it wouldn't be my go to example =)

pornel:
Yeah, we were going back and forth on LSRegisterURL, so I'm keen to try another approach.

PR #637, opened 2015-09-27
Test app cleanup

kainjow:
- When working with URLs, URL methods should be preferred vs converting to strings
- Changing the process working directory is not recommended by Apple docs


zorgiepoo:
Neat. I didn't know that NSURL method existed. I looked up the code on my machine and it appears to call `lstat` which means it doesn't follow symlinks - good to know.

I think longer term we may have to look at another way of starting a web server instance. It's kind of fragile. Eg, sometimes a stray python process is left open, especially if you hit the Stop button in Xcode's debugger, which is then followed by a bind address in use message.

kainjow:
@zorgiepoo what about a built-in web server? I played around and wrote one from scratch in Obj-C that just serves static GET requests from a directory. That would server Sparkle's testing needs and not require forking another process.

zorgiepoo:
That sounds like a good approach to me!

PR #638, opened 2015-09-27
Check existence of NSByteCountFormatter instead of comparing OS version

zorgiepoo:
The old code didn't look at the major version and is probably less efficient

### DIFF
--- a/Sparkle/SUUIBasedUpdateDriver.m
+++ b/Sparkle/SUUIBasedUpdateDriver.m
@@ -128,7 +128,7 @@
 
 - (NSString *)localizedStringFromByteCount:(long long)value
 {
-    if (SUHost.operatingSystemVersion.minorVersion < 8) {
+    if (![NSByteCountFormatter class]) {
         if (value < 1000) {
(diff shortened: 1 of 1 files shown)


jakepetroules:
DO NOT SUBMIT.

Major version's irrelevant, it's always 10. Also, we want to use the version, not check if the class exists. See https://developer.apple.com/videos/wwdc/2015/?id=411 at 11:35

zorgiepoo:
Please at least check the major version. It's a really bad/poor assumption that they will never bump it up. The field exists for a reason. Let me take a look at the video now..

jakepetroules:
You can use `isAtLeastOperatingSystemVersion:`, then.

zorgiepoo:
Are you implying to create such a method in Sparkle? (Since of course `isOperatingSystemAtLeastVersion:` for NSProcessInfo is only available for 10.10 or higher)

jakepetroules:
Yes - I thought I already implemented `isAtLeastOperatingSystemVersion:`, I guess I only did `operatingSystemVersion`. See #639.

PR #640, opened 2015-09-27
Fix warnings.

jakepetroules:


Sparkle/SUFileManager.m
```diff
@@ -178,9 +178,9 @@ - (BOOL)_removeXAttrWithAuthentication:(NSString *)name fromRootURL:(NSURL *)roo
         }
         return NO;
     }
-    
-    const char *xattrName = [name cStringUsingEncoding:NSASCIIStringEncoding];
-    if (xattrName == NULL) {
+
+    char xattrName[PATH_MAX] = {0};
+    if (![name getFileSystemRepresentation:xattrName maxLength:sizeof(xattrName)]) {
```
pornel: Why? `xattrName` isn't a path.

It's merely the `"com.apple.quarantine"` constant. Maybe instead of the conversion just use it as `const char *` everywhere?

Sparkle/SUFileManager.m
```diff
@@ -508,8 +508,8 @@ - (BOOL)changeOwnerAndGroupOfItemAtRootURL:(NSURL *)targetURL toMatchURL:(NSURL
     }
     
     NSString *formattedUserAndGroupIDs = [NSString stringWithFormat:@"%u:%u", ownerID.unsignedIntValue, groupID.unsignedIntValue];
-    const char *userAndGroup = [formattedUserAndGroupIDs cStringUsingEncoding:NSASCIIStringEncoding];
-    if (userAndGroup == NULL) {
+    char userAndGroup[PATH_MAX] = {0};
```
pornel: `snprintf` instead of two Cocoa functions?

Sparkle/SUFileManager.m
jakepetroules: Ah, but it requires a non-const, and casting will warn.

pornel: Eh, [constness in C is broken](http://stackoverflow.com/questions/5055655/double-pointer-const-correctness-warnings-in-c) for array of strings :(

In that case maybe completely oldschool:

```c
char quarantine[] = "com.apple.quarantine";
```

jakepetroules: Sure.

zorgiepoo:
I guess the warning isn't dumb. `AuthorizationExecuteWithPrivileges()` is free to alter the contents and the compiler won't let us ignore that.

zorgiepoo:
If we still want to stick to cocoa land:

``` objective-c
NSString *formattedUserAndGroupIDs = [NSString stringWithFormat:@"%u:%u", ownerID.unsignedIntValue, groupID.unsignedIntValue];
char userAndGroup[PATH_MAX] = {0};
if (![formattedUserAndGroupIDs getBytes:userAndGroup maxLength:sizeof(userAndGroup) usedLength:NULL encoding:NSUTF8StringEncoding options:(NSStringEncodingConversionOptions)0 range:NSMakeRange(0, formattedUserAndGroupIDs.length) remainingRange:NULL]) {
    if (error != NULL) {
        *error = [NSError errorWithDomain:NSCocoaErrorDomain code:NSFormattingError userInfo:@{ NSLocalizedDescriptionKey: [NSString stringWithFormat:@"Owner ID %u and Group ID %u could not be formatted.", ownerID.unsignedIntValue, groupID.unsignedIntValue] }];
    }
    return NO;
}
```

``` objective-c
char xattrName[PATH_MAX] = {0};
if (![name getBytes:xattrName maxLength:sizeof(xattrName) usedLength:NULL encoding:NSUTF8StringEncoding options:(NSStringEncodingConversionOptions)0 range:NSMakeRange(0, name.length) remainingRange:NULL]) {
    if (error != NULL) {
        *error = [NSError errorWithDomain:NSCocoaErrorDomain code:NSFileWriteInapplicableStringEncodingError userInfo:@{ NSLocalizedDescriptionKey: [NSString stringWithFormat:@"Extended attribute %@ is not a valid convertible string.", name] }];
    }
    return NO;
}
```
?

PR #641, opened 2015-09-28
Char conversions

pornel:
Re #640

### DIFF
--- a/Sparkle/SUFileManager.m
+++ b/Sparkle/SUFileManager.m
@@ -483,13 +483,10 @@ static BOOL AuthorizationExecuteWithPrivilegesAndWait(AuthorizationRef authoriza
         return NO;
     }
     
-    NSString *formattedUserAndGroupIDs = [NSString stringWithFormat:@"%u:%u", ownerID.unsignedIntValue, groupID.unsignedIntValue];
-    char userAndGroup[PATH_MAX] = {0};
-    if (![formattedUserAndGroupIDs getFileSystemRepresentation:userAndGroup maxLength:sizeof(userAndGroup)]) {
-        if (error != NULL) {
-            *error = [NSError errorWithDomain:NSCocoaErrorDomain code:NSFormattingError userInfo:@{ NSLocalizedDescriptionKey: [NSString stringWithFormat:@"Owner ID %u and Group ID %u could not be formatted.", ownerID.unsignedIntValue, groupID.unsignedIntValue] }];
-        }
-        return NO;
+    char userAndGroup[100];
+    int written = snprintf(userAndGroup, sizeof(userAndGroup), "%u:%u", ownerID.unsignedIntValue, groupID.unsignedIntValue);
+    if (written < 0 || written >= 100) {
+        return NO; // No custom error, because it's too unlikely to ever happen
     }
     
     if (![self _acquireAuthorizationWithError:error]) {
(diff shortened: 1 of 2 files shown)


Sparkle/SUFileManager.m
```diff
@@ -150,18 +140,12 @@ - (int)_removeXAttr:(NSString *)name fromFile:(NSString *)file options:(int)opti
...
-    }
-    
     return removexattr(path, attr, options);
 }
 
 #define XATTR_UTILITY_PATH "/usr/bin/xattr"
 // Recursively remove an xattr at a specified root URL with authentication
-- (BOOL)_removeXAttrWithAuthentication:(NSString *)name fromRootURL:(NSURL *)rootURL error:(NSError *__autoreleasing *)error
+- (BOOL)_removeXAttrWithAuthentication:(char *__nonnull)xattrName fromRootURL:(NSURL *)rootURL error:(NSError *__autoreleasing *)error
 {
```
zorgiepoo: Don't have time to verify right now, but Xcode 7 / latest SDK's might have changed __nonnull to _Nonnull (there's _Nullable too).

Sparkle/SUFileManager.m
```diff
@@ -507,13 +483,10 @@ - (BOOL)changeOwnerAndGroupOfItemAtRootURL:(NSURL *)targetURL toMatchURL:(NSURL
...
     }
     
-    NSString *formattedUserAndGroupIDs = [NSString stringWithFormat:@"%u:%u", ownerID.unsignedIntValue, groupID.unsignedIntValue];
-    char userAndGroup[PATH_MAX] = {0};
-    if (![formattedUserAndGroupIDs getFileSystemRepresentation:userAndGroup maxLength:sizeof(userAndGroup)]) {
-        if (error != NULL) {
-            *error = [NSError errorWithDomain:NSCocoaErrorDomain code:NSFormattingError userInfo:@{ NSLocalizedDescriptionKey: [NSString stringWithFormat:@"Owner ID %u and Group ID %u could not be formatted.", ownerID.unsignedIntValue, groupID.unsignedIntValue] }];
-        }
-        return NO;
+    char userAndGroup[100];
```
zorgiepoo: Can zero-initilaize `= {0}` to be consistent with other code. Don't know about snprintf but some of C's string functions are dumb.

Sparkle/SUFileManager.m
```diff
@@ -507,13 +483,10 @@ - (BOOL)changeOwnerAndGroupOfItemAtRootURL:(NSURL *)targetURL toMatchURL:(NSURL
...
+    int written = snprintf(userAndGroup, sizeof(userAndGroup), "%u:%u", ownerID.unsignedIntValue, groupID.unsignedIntValue);
+    if (written < 0 || written >= 100) {
```
zorgiepoo: Use sizeof(userAndGroup), and should it be > not >=?

zorgiepoo: Ok, >= looks right.

Sparkle/SUFileManager.m
pornel: `snprintf` guarantees `\0` on success, and the result is checked.

Sparkle/SUFileManager.m
zorgiepoo: Yeah. https://developer.apple.com/swift/blog/?id=25

__Nonnull and __Nullable are more modern to use but __nonnull and __nullable still exist for compatibility I suppose.

Sparkle/SUFileManager.m
```diff
@@ -13,11 +13,7 @@
 #include <sys/errno.h>
 #include <sys/time.h>
 
-#pragma clang diagnostic push
-// This is a private constant but it's referenced from our unit tests
-#pragma clang diagnostic ignored "-Wmissing-variable-declarations"
-NSString *SUAppleQuarantineIdentifier = @"com.apple.quarantine";
-#pragma clang diagnostic pop
+static char SUAppleQuarantineIdentifier[] = "com.apple.quarantine";
```
zorgiepoo: Isn't the point of the warning that `AuthorizationExecuteWithPrivileges()` can alter its arguments. Meaning in order to "properly" obey the warning's complaint, you need to either stick with `const char *` and make a copy when the function is invoked, or declare this as a non-static local variable that is assigned its value immediately.

pornel: No, I think that warning happens only because of [a design mistake in C's type system](http://c-faq.com/ansi/constmismatch.html). 

In C `const char *` argument can accept both `char *` and `const char *` values. However, an argument with `const char **` type only accepts `const char **` and not `char **`.
 
Therefore, in practice, you can't have a C function that takes an array of pointers and guarantees it won't modify them, because it will be unusable with non-const values.

So Apple couldn't declare that this function doesn't modify the arguments even if they wanted to.

So I'm relying on the fact that:

* Documentation doesn't say anything about them being modified
* The type is `char *const *` which is the closest practical think you can do to declare an immutable array of pointers

zorgiepoo: Fair enough.

Issue #649, opened 2015-10-07
BinaryDelta segfaults when given paths are under /tmp


rogual:
I'm getting a segmentation fault when running BinaryDelta on two versions of my app. I'm including a traceback — is this enough to go on?

```
$ lldb ~/Executables/binarydelta -- create /tmp/2.0.3-mac /tmp/2.0.4-mac /tmp-2.0.4-mac-patch.delta
(lldb) target create "/Users/me/Executables/binarydelta"
Current executable set to '/Users/me/Executables/binarydelta' (x86_64).
(lldb) settings set -- target.run-args  "create" "/tmp/2.0.3-mac" "/tmp/2.0.4-mac" "/tmp-2.0.4-mac-patch.delta"
(lldb) r
Process 17200 launched: '/Users/me/Executables/binarydelta' (x86_64)
Process 17200 stopped
* thread #1: tid = 0x8cb28, 0x00007fff97b91aee libxar.1.dylib`xar_opt_set + 82, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x8)
    frame #0: 0x00007fff97b91aee libxar.1.dylib`xar_opt_set + 82
libxar.1.dylib`xar_opt_set:
->  0x7fff97b91aee <+82>: movq   0x8(%r15), %rax
    0x7fff97b91af2 <+86>: movq   %rax, 0x18(%rbx)
    0x7fff97b91af6 <+90>: movq   %rbx, 0x8(%r15)
    0x7fff97b91afa <+94>: xorl   %eax, %eax
(lldb) bt
* thread #1: tid = 0x8cb28, 0x00007fff97b91aee libxar.1.dylib`xar_opt_set + 82, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x8)
  * frame #0: 0x00007fff97b91aee libxar.1.dylib`xar_opt_set + 82
    frame #1: 0x0000000100007bea binarydelta`createBinaryDelta(source=<unavailable>, destination=<unavailable>, patchFile=<unavailable>, majorVersion=<unavailable>, verbose='\0', error=0x00007fff5fbff880) + 2883 at SUBinaryDeltaCreate.m:442
    frame #2: 0x000000010000b37a binarydelta`main [inlined] runCreateCommand(programName=<unavailable>, args=<unavailable>) + 237 at SUBinaryDeltaTool.m:113
[...]

PR #650, opened 2015-10-07
Fix segfault when patch path has insufficient write permissions

zorgiepoo:
This addresses issue #649


Sparkle/SUBinaryDeltaCreate.m
```diff
@@ -439,6 +439,16 @@ BOOL createBinaryDelta(NSString *source, NSString *destination, NSString *patchF
 
     NSString *temporaryFile = temporaryPatchFile(patchFile);
     xar_t x = xar_open([temporaryFile fileSystemRepresentation], WRITE);
+    if (!x) {
+        if (verbose) {
+            fprintf(stderr, "\n");
```
jakepetroules: What...?

zorgiepoo: Check a few lines above where "\nGenerating delta..." is being printed out to stderr if verbose mode is on. This is done in several other places too when an error occurs and the tool aborts so the caller can print out more info if they choose to. (Generating delta.. and other messages don't end in new line in first place because the original author thought it'd be more natural to not advance to next line until current progress was done)

PR #654, opened 2015-10-09
Fix Test App crashing on systems earlier than 10.9

zorgiepoo:
Passing a NSURL instance to a NSTask's arguments calls
-[NSURL fileSystemRepresentation] which is only available on 10.9+


zorgiepoo:
Closed in favor of #655

kainjow:
Isn't this still necessary since the web server changes didn't touch the sign update parts?

PR #655, opened 2015-10-09
Replace use of Python with built-in web server

kainjow:
See comments for issue #637


zorgiepoo:
Tested it out and it works great for me :)

Issue #651, opened 2015-10-08
Test software updates over network shares


zorgiepoo:
This is a todo I need to look into more. Currently the test application (for me) crashes when I run it from a remote machine via AFP simply because copying the test app onto the local machine doesn't succeed. It doesn't have to do with the framework (I haven't tested that yet) but has something to do with it encountering a symlink and failing. Thus far, I have no clue why this happens.

zorgiepoo:
After [investigating more](https://gist.github.com/zorgiepoo/93cf1901419187c512f8), it looks like there's a bug in `copyfile()` when copying symbolic links from network mounted volumes. It also affects `NSFileManager`; of course, this tells us Apple doesn't use this function themselves for many file-copy tasks otherwise they would have found the issue. I have been testing with two 10.11 machines, and one 10.8 VM and have noticed the issue everywhere.

This is a real problem though. I will look into filing a rdar, and will be making changes to the `SUFileManager` class shortly. There may be another issue I have to look into as well...

zorgiepoo:
The real bug here was OS X's Samba client reported every symbolic link as zero bytes in size which made `copyfile()` not function.

The good news is that Apple fixed this bug in 10.12. We should eventually want to, I think, move to using the `NSFileManager` methods for copying if macOS >= 10.12.

PR #657, opened 2015-10-10
File Operations Copy Fixes

zorgiepoo:
See issue #651 and [my open radar issue](http://openradar.appspot.com/23059163) for current issues sparkle has with copying items from different network mounted drives. Also made a small improvement to allowing/disallowing the authorization code from triggering.

### DIFF
--- a/Sparkle/SUFileManager.m
+++ b/Sparkle/SUFileManager.m
@@ -310,9 +310,10 @@ static BOOL AuthorizationExecuteWithPrivilegesAndWait(AuthorizationRef authoriza
 /*
  * Copies an item from one location to another
  * This intentionally does *not* use copyfile() or any API that uses it such as NSFileManager's copy item method
- * This is because copyfile() fails to copy symbolic links from one network mount to another, which will affect copying apps
- * This failure has been reproduced on 10.11.0 and a VM running 10.8, so the issue has existed for a while.
- * For more info, see the bug report at http://openradar.appspot.com/radar?id=4925873463492608
+ * This is because copyfile() can fail to copy symbolic links from one network mount to another, which will affect copying apps
+ * This failure occurs because the system may think symbolic links on a SMB mount are zero bytes in size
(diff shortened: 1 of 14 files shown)


zorgiepoo:
So I found out OS X SMB (default nowadays) mounts [report the size of any symbolic link as zero bytes](http://openradar.appspot.com/23121198) is why copyfile fails. The other copy function I used here works because it doesn't stat the file to see how big of a buffer to allocate.

pornel:
The changes look good, but it's a change to a core functionality of Sparkle, so I think it'd be better to release v1.12 before merging this in.

zorgiepoo:
Ultimately it's a bug from Apple but updates will certainly fail from network mounted machines if you're okay with that as it is currently. The core functionality has already changed from v1.11, and from my tests 1.11 is not as heavily impacted by this bug (update works but old app isn't moved to trash).

Sparkle/SUFileManager.m
```diff
@@ -293,32 +307,191 @@ - (BOOL)_releaseItemFromQuarantineAtRootURL:(NSURL *)rootURL allowingAuthenticat
...
+    
+    char destinationPath[PATH_MAX] = {0};
+    if (![destinationURL.path getFileSystemRepresentation:destinationPath maxLength:sizeof(destinationPath)]) {
+        if (error != NULL) {
+            *error = [NSError errorWithDomain:NSCocoaErrorDomain code:NSFileReadInvalidFileNameError userInfo:@{ NSLocalizedDescriptionKey: [NSString stringWithFormat:@"Destination to copy file to (%@) cannot be represented as a valid file name.", destinationURL.lastPathComponent] }];
+        }
+        return NO;
+    }
+    
+    if (!AuthorizationExecuteWithPrivilegesAndWait(_auth, "/bin/cp", kAuthorizationFlagDefaults, (char *[]){ "-Rf", sourcePath, destinationPath, NULL })) {
```
pornel: It'd be safer if there was an `--` argument between flags and source, so that argument-like file names can't be confused with arguments.

PR #661, opened 2015-10-21
Misc fixes

osolo:
1. The change in SUUpdater.m is to support the cases where SUUpdater is subclassed from a different bundle.  In that case, using [self class] to try to determine the Sparkle Bundle is incorrect.  You must use [SUUpdater class] to ensure you're targeting the actual Sparkle bundle.

2. The other changes are just to get rid of some Xcode warnings which a a result of some .XIB's having their deployment target set to OSX 10.6 while the project is targeted at OSX 10.7.


pornel:
Good catch with the `self class`. Thank you!

zorgiepoo:
This is the type of change that probably needs to be commented in code so it's not changed back by someone in the future (re: bundle change)

Issue #660, opened 2015-10-20
didDownloadUpdate delegate API request


kartikthapar:
I need to know whenever Sparkle has downloaded an update in the background. So just like there is a `updater:failedToDownloadUpdate:error:` API, I am wondering if there could be a `updater:didDownloadUpdate:` delegate API.

Most probably implement in `downloadDidFinish:` `NSURLDownload` delegate API or after the unarchiver has finished unarchiving the update.

kartikthapar:
Well, `updater:willInstallUpdateOnQuit: immediateInstallationInvocation:` is more than enough for what I was trying to do originally. I'll close this issue for now. Thanks!

PR #662, opened 2015-10-21
New feature: didDownloadUpdate delegate method

osolo:
I added this feature to solve my special use case and may be useful to others.

Use case: In order to avoid using a .pkg which would force the user to enter their admin password during the upgrade, I decided to package my updates a regular .zip bundle.  So Sparkle will download and put the files in the right place.  But the the package had an important post-flight script that needed to be ran, which won't happen now.  The way I solved this is by adding a "didDownloadUpdate" event to the delegate.  In my case, I check to see if there is a post-processing script in the downloaded bundle and execute it.  

An example is given in the comment block for the new delegate method.

### DIFF
--- a/Sparkle/SUBasicUpdateDriver.m
+++ b/Sparkle/SUBasicUpdateDriver.m
@@ -417,6 +417,13 @@
         return;
     }
 
+    id<SUUpdaterDelegate> updaterDelegate = [self.updater delegate];
+    if ([updaterDelegate respondsToSelector:@selector(updater:didDownloadUpdate:tempPath:)])
+    {
+        NSString *installSourcePath = [SUInstaller installSourcePathInUpdateFolder:self.tempDir forHost:self.host isPackage:NULL isGuided:NULL];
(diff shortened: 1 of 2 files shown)


kartikthapar:
Simply referencing — #660. :blush:

:+1:

osolo:
@kartikthapar There is a critical difference here in that this new method actually gives you a pointer to the downloaded bundle so you can make changes to it.  I might be wrong, but I don't think you can easily get that from just an SUUpdater/SUAppcastItem pair.

Also, willInstallUpdateOnQuit is only called for auto updates, and abusing shouldPostponeRelaunchForUpdate for this purpose just seems wrong.

kartikthapar:
@osolo That is cool. Your update is a lot more comprehensive. I simply needed a callback. In any case, thanks!

pornel:
Sorry, I'm not sure if that's something we want to support. When we move to sandboxing, the delegate will be untrusted and won't have permission to access the downloaded update.

If you need to run extra tasks after the upgrade, I suggest running them on startup of the app after upgrade.

PR #674, opened 2015-11-04
Disable warning for "out of range" enum assignments

JeremyAgost:
Security framework specifies flag arguments that aren't declared as options for the flag type so these warnings are spurious.


zorgiepoo:
Where in Sparkle's code does this warning fire off?

To sidestep the warning, when you really need to, isn't it more ideal to use a typecast to the enum type?

jakepetroules:
Pretty sure I already properly fixed those warnings. What Xcode are you using?

JeremyAgost:
Typecast to the appropriate enum type isn't possible because the Security headers are poorly written and the flag argument is frequently a different type then the valid flags are defined as.

You may have fixed it already, I think I was inadvertently basing off an earlier revision. I'll check against your master.

pornel:
I think it was fixed alredy

PR #678, opened 2015-11-08
fsref func

pornel:


Sparkle/SUFileManager.m
```diff
@@ -39,6 +39,29 @@ static BOOL AuthorizationExecuteWithPrivilegesAndWait(AuthorizationRef authoriza
     return returnValue;
 }
 
+static BOOL SUMakeRefFromURL(NSURL *url, FSRef *ref, NSError **error) {
+
+    char sourcePath[PATH_MAX] = {0};
```
zorgiepoo: probably should be named path instead of sourcePath now that it's in a generalized function?

Sparkle/SUFileManager.m
```diff
@@ -39,6 +39,29 @@ static BOOL AuthorizationExecuteWithPrivilegesAndWait(AuthorizationRef authoriza
...
+    if (![url.path getFileSystemRepresentation:sourcePath maxLength:sizeof(sourcePath)]) {
+        if (error != NULL) {
+            *error = [NSError errorWithDomain:NSCocoaErrorDomain code:NSFileReadInvalidFileNameError userInfo:@{ NSLocalizedDescriptionKey: [NSString stringWithFormat:@"File to copy (%@) cannot be represented as a valid file name.", url.lastPathComponent] }];
```
zorgiepoo: This error description may need to be more generalized I guess.

zorgiepoo:
Should the double dash apply to other commands being used too? (mv, rm, xattr, chown)

pornel:
Added `--` to commands, except chown, which doesn't support it.

Sparkle/SUFileManager.m
```diff
@@ -182,7 +205,7 @@ - (BOOL)_removeXAttrWithAuthentication:(char *__nonnull)xattrName fromRootURL:(N
         return NO;
     }
     
-    BOOL success = AuthorizationExecuteWithPrivilegesAndWait(_auth, XATTR_UTILITY_PATH, kAuthorizationFlagDefaults, (char *[]){ "-s", "-r", "-d", xattrName, path, NULL });
+    BOOL success = AuthorizationExecuteWithPrivilegesAndWait(_auth, XATTR_UTILITY_PATH, kAuthorizationFlagDefaults, (char *[]){ "-s", "-r", "-d", xattrName, "--", path, NULL });
```
zorgiepoo: double dash for latter does not work for me on the command line.

> xattr -s -r -d com.apple.quarantine -- ~/Downloads/My\ App.app
> xattr: No such file: --

pornel: Thanks for catching that. I've tested it wrong with `xattr -l -- file`

zorgiepoo:
Looks good to me.

PR #681, opened 2015-11-10
Re-generated XIBs

### LINKED ISSUE #672: SUUpdateAlert broken due to invalid constraints
When setting SUAllowsAutomaticUpdates to NO, the SUUpdaterAlert view tries to hide the checkbox using a NSLayoutConstraint.

We then get the following:
```*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '*** +[NSLayoutConstraint constraintWithItem:attribute:relatedBy:toItem:attribute:multiplier:constant:]: A multiplier of 0 or a nil second item together with a location for the first attribute creates an illegal constraint of a location equal to a constant. Location attributes must be specified in pairs'```

It seems only the English xib has the proper outlets connected (releaseNotesContainerView and automaticallyInstallUpdatesButton in this case). The descriptionField outlet is also misconfigured.

pornel:
Should fix #672 

I'm not sure whether Arabic layouts are correct. Can someone check?


demoszp:
Hello. You broke many localizations (eg German)
screen shot 2015-11-16 at 12 33 12 nachm

pornel:
Sorry. I'll fix it.

PR #684, opened 2015-11-13
Automatic Encrypted DMG Support

andrewboyd-ge:
Add the capability to set a decryption password for updating with an encrypted dmg file without prompting the user for a password.

This enables the the ability to update using an encrypted disk image without having to prompt the user for a password. This is useful when access to the application binaries is only available via distribution or from a location in which authentication is required. This can help prevent an attacker from having unwanted access to application binaries for reverse engineering.

This assists with a defense in depth approach to security.


kainjow:
IMO should resubmit this without your merge commit.

andrewboyd-ge:
Will reopen another with the merge commit

pornel:
Thank you for the pull request. Can you explain what is the threat model? Who is the attacker, who's trusted?

Sparkle/SUDiskImageUnarchiver.m
```diff
@@ -70,8 +70,21 @@ - (void)extractDMGWithPassword:(NSString *)__unused password
...
-
+        NSMutableArray *arguments = [@[@"attach", self.archivePath, @"-mountpoint", mountPoint, /*@"-noverify",*/ @"-nobrowse", @"-noautoopen"] mutableCopy];
+    
+        if (self.decryptionPassword) {
+            NSMutableData *passwordData = [[self.decryptionPassword dataUsingEncoding:NSUTF8StringEncoding] mutableCopy];
+            // From the hdiutil docs:
+            // read a null-terminated passphrase from standard input
+            //
+            // Add the null terminator, then the newline
+            [passwordData appendData:[NSData dataWithBytes:"\0\n" length:1]];
```
pornel: that's 2 characters and length:1

pornel:
#685

PR #685, opened 2015-11-13
Automatic Encrypted DMG Support

andrewkboyd:
Add the capability to set a decryption password for updating with an encrypted dmg file without prompting the user for a password.

This enables the the ability to update using an encrypted disk image without having to prompt the user for a password. This is useful when access to the application binaries is only available via distribution or from a location in which authentication is required. This can help prevent an attacker from having unwanted access to application binaries for reverse engineering.

This assists with a defense in depth approach to security.


andrewboyd-ge:
The primary use case for this is for binaries that are distributed privately.  In those cases, whether an application is distributed on a private network with authentication, or explicitly given to individual users, there are scenarios in which it is desirable to update that application without placing the unencrypted on the open internet. This can allow application developers to protect IP for enterprises in custom security related applications. 

Who is trusted?
Legitimate employees or customers of a business that controls access to application binaries.

Who is the attacker?
Anyone that may want to gain access to a private network in some scenarios, or attackers that may want to gain IP, or perform recon by analyzing binaries for security solutions.

jakepetroules:
Didn't we *remove* this functionality at some point in the past?

pornel:
Yes, we did. But it looks like it's needed.

Issue #31, opened 2011-10-23
Sparkle assumes that greater versions are always newer, which is wrong


andymatuschak:
_**Peter Hosey (boredzo)_\* [reported (on Launchpad)](https://bugs.launchpad.net/sparkle/+bug/391965) on' 2009-06-25:*

When Sparkle looks for a usable version among the appcast items (see #228485), it sorts the items by date. 
This is a problem, demonstrated by Adium 1.4b6.

Our appcast for Adium betas includes both 1.4b7 and 1.3.5; of these,
1.3.5 is actually the newer (more recent) release. Because Sparkle sorts
the items by date, it looks at 1.3.5 first; it finds that 1.3.5 will run
on the host, so it stops looking for more versions. However, 1.3.5 is a
lower version number than 1.4b7, so it tells the user that “1.4b6 is the
latest version available!”—which is wrong, since b7 is available.

We're not sure exactly which version of Sparkle we're using (it was a
build from source, and the developer who committed it didn't note the
version in his commit message). All I know is that CFBundleVersion is
340, for whatever that's worth.

pornel:
Fixed

PR #689, opened 2015-12-04
Sparkle now looks for best update. Added notification when update requires newer OS

fzwo:
Two major changes:

---

Sparkle now looks for the best update, not just the first compatible one.

---

Added support for an alert when an update requiring a newer OS version is available if no currently applicable update is found.

This is off by default, and Sparkle will behave like it used to (except change 1). If you set `-SUUpdater shouldAlertForUpdatesRequiringNewerOS` to `YES`, Sparkle's UI-based updaters will now show an alert informing the user that an update is available, but requires an OS update (including what OS is needed and what they currently have installed). This alert is only shown once per available version, so the user is not bothered every time the app is launched.

---

Also added one missing German i18n string (in addition to the two new ones for change 2).


fzwo:
Do I have to do anything to get the coveralls check to complete?

jakepetroules:
> Do I have to do anything to get the coveralls check to complete?

No, it runs automatically when you push changes.

fzwo:
Wow, you're fast!
It always shows as pending for me. Travis finishes after a few minutes, but coveralls never does.

jakepetroules:
Yeah I wouldn't worry about it.

aprilnoren:
Have a custom hack of something similar to this in our fork.  Looking forward to this getting merged to master.

zorgiepoo:
There was discussion in #31 about changing the updater to look at the highest compatible version instead of by the most recent date (how it works currently). I agree that changing it by version makes more sense like the way you've done.

For the second issue here, it looks like a reasonable concern to me. I'm not sure if something similar would be wanted for other use cases though like paid upgrades but that's probably more complex..

master has been updated since with changes to the update driver so the pull request is a little bit out of date.

zorgiepoo:
> This alert is only shown once per available version, so the user is not bothered every time the app is launched.

This also just got me thinking if this is actually good enough. Every time a minor version is released, the user will be prompted about the new version that they can't install unless they upgrade their OS.

Sparkle/SUUIBasedUpdateDriver.m
```diff
@@ -86,6 +86,28 @@ - (void)didFindValidUpdate
...
+    [[NSNotificationCenter defaultCenter] postNotificationName:SUUpdaterDidFindUpdateRequiringNewerOSNotification object:self.updater];
+
+    if (self.updater.shouldAlertForUpdatesRequiringNewerOS) {
+        [self.host setObject:[self.updateItemRequiringNewerOS versionString] forUserDefaultsKey:SUSkippedVersionBecauseMinimumOSWasTooLowKey];
+
+        NSString *requiredOSVersion = self.updateItemRequiringNewerOS.minimumSystemVersion;
+        NSString * currentOSVersion = [[NSProcessInfo processInfo] operatingSystemVersionString];
+        NSAlert *alert = [[NSAlert alloc] init];
+        alert.messageText = SULocalizedString(@"Update available!", "Status message shown when the user checks for updates and one is available, but the user needs to update their operating system to be eligible.");
+        alert.informativeText = [NSString stringWithFormat:SULocalizedString(@"%@ %@ is now available, but it requires Mac OS X %@--you are running %@.\nPlease update your Mac and check for updates again.", nil), [self.host name], [self.host displayVersion], requiredOSVersion, currentOSVersion];
```
zorgiepoo: It's marketed as "OS X" these days (not "Mac OS X")

fzwo: Thanks! I had thought about that, but I think having "Mac" in there is helpful for some less-technical users (and theoretically, an update could target an older (but still new for the user) OS that is still billed as "Mac OS X"). I don't have any emotional attachment to that though, so I'm open to changing that.

jakepetroules: You must remove "Mac". We do not refer to "Mac OS X" anywhere in the Sparkle codebase.

Note that *all* Apple marketing materials say "OS X", never "Mac OS X" - http://www.apple.com/osx/, so the "less-technical users" argument is kind of moot since many of those new to Apple will have never even seen "Mac OS X" written out before.

Sparkle/SUUIBasedUpdateDriver.m
```diff
@@ -86,6 +86,28 @@ - (void)didFindValidUpdate
...
+        [alert addButtonWithTitle:SULocalizedString(@"OK", nil)];
+        [self showModalAlert:alert];
```
jakepetroules: This doesn't build anymore.

fzwo: Oops!

I'll get on this tomorrow.

fzwo: It's done. Sorry for that.

fzwo:
@zorgiepoo I was also wondering if this was maybe too aggressive, but I really wanted the user to be poked a bit about updating their OS.

Since in most cases, versions compatible with old OS versions won't receive updates anymore anyway, the obvious workaround for the user is to just disable automatic checking if they're bothered.

zorgiepoo:
Alternatively (haven't looked at this deeply) track the last minimum system version that was requested instead. I think it's a little too aggressive.

fzwo:
And then ask again when an update is out that requires an even newer OS? While that would very probably reduce the number of update notifications, it seems a bit random.

I also thought about tracking the user's OS version when last asked, and ask again after an update. But for an old OS, that would probably mean "never".

pornel:
That's an interesting problem!

- if nagging is too frequent users will get annoyed, and may disable updates completely. That's the worst scenario for us, because then users won't even be able to get security/maintenance fixes that are compatible with their OS.

- users may not be able to update their OS even if they wanted to (e.g. old Mac or OS X install with App Store bugs or account problems preventing access to updates). I think in such case users should have option to continue using the app without nags.


So how about:

- when Sparkle checks *automatically* then notify only once (or only once per month? once in 10 checks?) — basically so rarely that it won't get annoying.

- but when user checks **manually** then notify every time.

zorgiepoo:
> And then ask again when an update is out that requires an even newer OS? While that would very probably reduce the number of update notifications, it seems a bit random.

No, not if they're still behind the minimum system version before you just bumped it up. If the user couldn't upgrade before, they certainly wouldn't be able to after you bumped it up (assuming the user hasn't upgraded their OS).

I believe what you'll want to track is the lowest OS version requirement that exceeds the user's OS version. For example, if the user is on 10.5 and the app requires 10.6.0+, Sparkle keeps track of minimum OS that exceeds the user as 10.6.0. Now if you release a new version that requires 10.7.0+, Sparkle should still keep track of 10.6.0 because the user is still behind that version. If the user updates to 10.6.0, then Sparkle can notify them about the newer requirement and then should start keeping track of 10.7.0. Let me know if this makes sense.

I don't think we need to over complicate it by notifying an arbitrary amount of time like once per month. pornell brings up a good point though about showing latest update if the user initiates an update manually, which I think the PR already does.

zorgiepoo:
Hmm. This is more complicated than I thought. We can't forget about the minimum system version being bumped down too. Eg:

* User on 10.5.0 is on app version 1.0
* 1.1 comes out but requires 10.6.0+
* User is automatically notified of newer update requiring higher OS version
* 1.2 is released but drops requirement back to 10.5.0
* User updates to 1.2
* 1.3 is released but requires 10.6.0+ again
* User should be notified again of newer update requiring higher OS version

Perhaps keep track of what app version they are on too, and if it has changed, then feel free to notify them again (so reset the notification behavior logic). I think that's okay; if a user updates an app because of a security patch for example, it might be nice to remind them again that they're on an insufficient OS version.

fzwo:
I'm not sure what's going wrong with the Travis check there here as that last merge seemed quite harmless to me.

If anyone can tell me what went wrong, I'll fix it, but since I won't be near a computer for the next week, I've also given @jakepetroules, @zorgiepoo and @pornel write access to my fork.

zorgiepoo:
I submitted a PR in #780 since I had issues in my fork too. (Some of the errors were hard to pinpoint because SUUpdater delays its start cycle and issues pop up in other tests..)

On topic with this thread, the first issue has been recently merged in.

fzwo:
@zorgiepoo Thanks for merging the first issue!

pornel:
We have this partially merged. I think it still would be good to notify about new OS requirement.

@fzwo would you be interested in giving it another try?

Perhaps for now we could simplify the problem to informing user only on manual checks, i.e. change wording of "you're up to date!" alert to info about OS version.

zorgiepoo:
You may want to wait a bit, because the approach will most likely conflict with my [fork](https://github.com/sparkle-project/Sparkle/issues/962) which could be the basis for a newer version. Unless you want to take a look at that branch too :).

I agree though just doing a change for the manual check would be simpler.

fzwo:
I will be a bit short on time for the next few weeks, so I think it's probably best to wait for @zorgiepoo at the moment, if that's OK with you.

pornel:
Deferring it to Sparkle 2.

zorgiepoo:
2.x has better support for detecting unsupported min OS now (#1850, #1886)

PR #690, opened 2015-12-07
SUUpdatePermissionPrompt xib fixes

ksuther:
The regeneration of the SUUpdatePermissionPrompt xibs in #681 resulted in some incorrect layouts for some of the languages.

I fixed the button widths so they fit the content as well as fixed overlapping title and message text in some languages.

These could really use some Auto Layout and single xib love :)


pornel:
Thanks! I forgot this file doesn't use autolayout yet.

PR #692, opened 2015-12-14
Warn about ATS

pornel:
Show alert if the app is using HTTP feed URL and does not have ATS Info.plist keys

### DIFF
--- a/Sparkle/SUUpdater.m
+++ b/Sparkle/SUUpdater.m
@@ -145,9 +145,12 @@ static NSString *const SUUpdaterDefaultsObservationContext = @"SUUpdaterDefaults
 
 #if __MAC_OS_X_VERSION_MAX_ALLOWED >= 101100
     BOOL atsExceptionsExist = nil != [self.host objectForInfoDictionaryKey:@"NSAppTransportSecurity"];
-    if (!servingOverHttps && !atsExceptionsExist) {
+    if (isMainBundle && !servingOverHttps && !atsExceptionsExist) {
         [self showModalAlertText:@"Insecure feed URL is blocked in OS X 10.11"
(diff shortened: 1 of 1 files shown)


Sparkle/SUUpdater.m
```diff
@@ -137,6 +118,43 @@ - (instancetype)initForBundle:(NSBundle *)bundle
...
+                 informativeText:@"For security reasons, you need to sign your updates with a DSA key. See Sparkle's documentation for more information."];
+    } else if (isMainBundle && !(hasPublicDSAKey || hostIsCodeSigned)) {
+        [self showModalAlertText:@"Insecure update error!"
+                 informativeText:@"For security reasons, you need to code sign your application or sign your updates with a DSA key. See Sparkle's documentation for more information."];
+    } else if (isMainBundle && !hasPublicDSAKey && !servingOverHttps) {
+        SULog(@"WARNING: Serving updates over HTTP without signing them with a DSA key is deprecated and may not be possible in a future release. Please serve your updates over https, or sign them with a DSA key, or do both. See Sparkle's documentation for more information.");
+    }
+
+#if __MAC_OS_X_VERSION_MAX_ALLOWED >= 101100
+    BOOL atsExceptionsExist = nil != [self.host objectForInfoDictionaryKey:@"NSAppTransportSecurity"];
```
alonre: Very good idea!
However, having `NSAppTransportSecurity` in the `info.plist` is not sufficient guarantee that exceptions are in place.
Meaning - it might be empty or include other settings (thus update will still fail), but the alert won't show.
Consider checking if either `AllowArbitraryLoads` or `NSExceptionDomains` exists to increase chances the analysis is correct (unfortunately I can't think of a bullet-proof check - maybe someone else has a better idea).

pornel: I'd rather avoid getting into parsing of the exceptions. I assume that once somebody knows about the limitation they'll test it.

alonre: Fair enough.

PR #696, opened 2015-12-15
Change bundle ID

### LINKED ISSUE #687: An error occurred while relaunching <APP>. Couldn't find the relauncher
Background:
 * Current version of the app uses a custom build of Sparkle (from commit 0ed83cf I believe, so after v1.5b6 but before v1.6)
 * The app updates fine on OSX 10.9, 10.10, and 10.11 (not changing Sparkle)
 * We are trying to update to a new custom build of Sparkle to get the latest-and-greatest as well as add some app-specific update features. Ideally, it would be based off one of the most recent release versions (1.11.1 or 1.12.0b1)

Unfortunately, whenever I try to update from a version of the app using the old Sparkle to a version built with a newer version of Sparkle, I get the following errors in the log:

```Sparkle Error: An error occurred while relaunching MyApp, but the new version will be available next time you run MyApp.```
```Sparkle Error (continued): Couldn't find the relauncher (expected to find it at /Users/myusername/Library/Application Support/MyApp)```

I've watched that directory as the update was taking place and it seems there is no difference between an update that fails (new Sparkle) and an update that succeed (unchanged/old Sparkle). Both of them extract the app to `/Users/myusername/Library/Application Support/MyApp/MyApp <build#>/` which contains both `MyApp.zip` and an unzipped `MyApp`

According to the code, it spits out that error when it can't find the relaunching tool. I initially suspected that there was something missing in the extraction directory, but considering that it has the same structure when an update *does* succeed, I'm not sure what to make of it.

Does anyone have any idea what could be going wrong or how to fix this issue? Does the relaunching tool get placed in some other directory not indicated by the log?

pornel:
Fixes #687

### DIFF
--- a/Configurations/ConfigCommon.xcconfig
+++ b/Configurations/ConfigCommon.xcconfig
@@ -1,6 +1,6 @@
 // Common
 
-SPARKLE_BUNDLE_IDENTIFIER = org.andymatuschak.Sparkle
+SPARKLE_BUNDLE_IDENTIFIER = org.sparkle-project.Sparkle
 SPARKLE_RELAUNCH_TOOL_NAME = Autoupdate
 
 SPARKLE_APPEND_VERSION_NUMBER = 1


pornel:
I think we could change bundle ID and release 1.13 just with that, so if it causes any complications, users can easily fall back to 1.12 without losing anything.

pornel:
I'm assuming no objections :)

PR #706, opened 2016-01-04
Add some Japanese lozalized strings

1024jp:
Added to the strings that were not yet localized.

### DIFF
--- a/Sparkle/ja.lproj/Sparkle.strings
+++ b/Sparkle/ja.lproj/Sparkle.strings
@@ -10,7 +10,7 @@
 "%@ %@ is now available--you have %@. Would you like to download it now?" = "%1$@ %2$@ が入手できます（使用中のバージョンは %3$@ です）。今すぐダウンロードしますか？";
 
 /* Description text for SUUpdateAlert when the update informational with no download. */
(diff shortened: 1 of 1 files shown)


Sparkle/ja.lproj/Sparkle.strings
```diff
@@ -35,7 +35,7 @@
 "An important update to %@ is ready to install" = "%@ の重要なアップデートがインストールできます。";
 
 /* the unit for bytes */
-"B" = "B";
+"B" = "バイト";
```
jakepetroules: Can you translate KB, MB, GB, also?

1024jp: No, you don't need. It's the standard localization on OS X that only "bytes" is translated in Japanese.
For instance, NSDataFormatter localizes in the same way.

fig. A screen shot of Finder's Get Info (see the third line). 
screen shot 2016-01-04 at 14 40 23

However: I've just realized this is not "bytes" but "B". I'm actually not sure how Apple localizes "B", as "bytes" is normally used instead of "B" in such cases.
But anyway, I feel the localized "バイト" (means bytes) is more natural for here and KB, MB, GB are never localized in Japanese on OS X.

pornel:
ありがとうございます！

1024jp:
Marge ありがとうございます☕️

PR #716, opened 2016-01-17
Alert driver

pornel:
Checking of configuration in `SUUpdater` constructor can't be done properly, because at that point there's no delegate set yet. I've previously tried to work around that by deferring execution of the check, but in retrospect that wasn't a good idea, as it's non-deterministic.

To sort this out I've moved config check to be done later, right before actual update check, and the UI is driver-specific, so only the *UI* driver will show the alert as an actual modal dialog box. While it's not as eager and in-your-face as the previous setup, it's deterministic and doesn't complicate unit tests (we don't want modal dialogs popping up during unit tests :)


jakepetroules:
:+1:

PR #719, opened 2016-01-25
Translated new string, fixed double space, fixed “archive” and “relau…

vitu:
Translated new string, fixed double space, fixed “archive” and “relaunch” terminologies for pt-BR.


pornel:
Thank you

vitu:
You're welcome!

PR #723, opened 2016-01-29
Make test app available in all languages, fix #695

### LINKED ISSUE #695: Make Test App Localization-Test Friendly
I've found that localizations don't work with the test app. I think this is because the test app only has an English language folder.. It'd be nice to have to be able to switch the system language and see it reflected in the test app.

xhacker:


pornel:
Thanks!

PR #728, opened 2016-02-01
Logs

pornel:
Added logging that helps diagnose malformed XML and redirects we block.

### DIFF
--- a/Sparkle/SUAppcast.m
+++ b/Sparkle/SUAppcast.m
@@ -229,7 +229,11 @@
     }
 
     if (failed) {
-        [self reportError:[NSError errorWithDomain:SUSparkleErrorDomain code:SUAppcastParseError userInfo:@{ NSLocalizedDescriptionKey: SULocalizedString(@"An error occurred while parsing the update feed.", nil) }]];
+        [self reportError:[NSError errorWithDomain:SUSparkleErrorDomain
+                                              code:SUAppcastParseError
+                                          userInfo:@{
+                                                     NSLocalizedDescriptionKey: SULocalizedString(@"An error occurred while parsing the update feed.", nil),
+                                                     NSUnderlyingErrorKey: error ? error : [NSNull null]}]];
     } else {
         self.completionBlock(nil);
         self.completionBlock = nil;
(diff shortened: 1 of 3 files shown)


Sparkle/SUAppcast.m
```diff
@@ -229,7 +229,11 @@ - (void)downloadDidFinish:(NSURLDownload *)__unused aDownload
     }
 
     if (failed) {
-        [self reportError:[NSError errorWithDomain:SUSparkleErrorDomain code:SUAppcastParseError userInfo:@{ NSLocalizedDescriptionKey: SULocalizedString(@"An error occurred while parsing the update feed.", nil) }]];
+        [self reportError:[NSError errorWithDomain:SUSparkleErrorDomain
+                                              code:SUAppcastParseError
+                                          userInfo:@{
+                                                     NSLocalizedDescriptionKey: SULocalizedString(@"An error occurred while parsing the update feed.", nil),
+                                                     NSUnderlyingErrorKey: error ? error : [NSNull null]}]];
```
zorgiepoo: This requires careful consideration. It's *very* dangerous to touch an NSError* in the even slightest if an operation succeeded, especially if it's coming from Apple's APIs, and it may not even be nil. Also see Apple's [note](https://developer.apple.com/library/ios/documentation/Cocoa/Conceptual/ErrorHandlingCocoa/CreateCustomizeNSError/CreateCustomizeNSError.html) on the matter.

A boolean should be used to keep track if an error occurred, by only looking at the boolean return values of such operations. I'd also personally avoid reusing the same error pointer for different operations.

pornel:
Very good point. Thanks.

pornel:
I've changed error handling in 24d7bf9d46a52a51d8dc09b5cabeb792f75c93fc is that OK?

zorgiepoo:
As far as I can see, it looks better. I feel uneasy about using NSNull anywhere for the NSUnderlyingErrorKey though. We should probably just not populate the field instead, that way no one will worry about handling it being not present or a NSNull instance.

The indentation/formatting on the source file needs some cleaning btw.

zorgiepoo:
@pornel after re-looking, looks like abortUpdateWithError: doesn't handle NSNull that is being passed in elsewhere as I feared. The code should be changed to never pass NSNull anywhere.

pornel:
OK, I've removed NSNull in #744. Please check if I haven't made another mistake :)

PR #737, opened 2016-02-07
UI Tests for Test Application

zorgiepoo:
These are UI Tests for the Sparkle Test Application using Apple's new UI Testing framework. Right now there is just a single test for the test application, but there could be more in the future (eg: one for delta updates).

I played around with these before 10.11 was released but figured it'd be more appropriate to show them now. They require OS X 10.11 or later to run them first of all, which Travis now supports. Unfortunately a [hack](http://stackoverflow.com/a/34234072/871119) may be required for enabling UI accessibility over there. Worst scenario is we can add them to the repo and wait later until we can automate running them.

(They also require system language to be English at the moment, since adding accessibility identifiers to many different versions of the same nib seems kind of painful...)

I added new Xcode configuration files, a new entry to the Makefile, and a shared scheme for the UI tests. I'm a newbie at that kind of stuff, so hopefully didn't do anything too wrong there.


jakepetroules:
Yeah, looks pretty good. @pornel?

pornel:
Cool

Issue #751, opened 2016-02-16
Disable Javascript?


zorgiepoo:
Has disabling javascript or making it opt-in in the web view been given any thought? I scavenged through several apps and wasn't able to find any release notes that couldn't function without it.

pornel:
:+1: I couldn't find any appcasts using it.

jakepetroules:
Off by default and public API to enable it? And/or Info.plist key.

pornel:
That's done

PR #755, opened 2016-02-17
Disable javascript

zorgiepoo:
See issue #751

I added code to make it opt-in in via an Info.plist key, in similar spirit to App Transport Security. Also, Java is just plain disabled (although I'm not sure if it was actually enabled before because of plug-ins being disabled).


zorgiepoo:
Ack, I meant "set SUEnableJavaScript to true" in the commit message, not false. *sighs*

pornel:
Thanks

Issue #753, opened 2016-02-16
Appcast Parser Broken for Multiple Items


zorgiepoo:
@pornel - The recent changes (after v1.13.1) broke the way the appcast is supposed to be parsed.

1. Open up the Test Application.
2. Open `~/Library/Caches/org.sparkle-project.SparkleTestApp/sparkletestcast.xml` in a text editor
3. Create a 2nd appcast item with a different version, description, title into the xml file.
4. Set a breakpoint in `-[SUAppcast downloadDidFinish:]` after `appcastItems` is initialized
5. Check for updates in the Test Application
6. Examine `appcastItems` in the debugger and note that both appcast items are exactly the same

Not reproducible in v1.13.1. Testing by hand seems to always be a worthwhile exercise.

pornel:
Fixed by #760

PR #756, opened 2016-02-17
Best Update by Version & Date Removal

zorgiepoo:
Code relevant to finding the best update has been taken from #689. I thought the issue was important enough to deserve its own PR. Plus this changes/removes the date handling code we have. I have branched from v1.13.1 for this issue because I think this is currently blocked by #753.

--
Sparkle's past behavior was sorting appcast items by published dates and
picking the first compatible item. However, it makes more sense to pick
the compatible item with the highest version (see issue #31).

The NSDate instance creation & sorting has also been removed from the code.
The methods we were using to parse the string into a NSDate have been deprecated,
and are strongly discouraged. Namely, all methods taking an arbitrary/natural string
have been nuked signaling they were probably a really bad idea. Furthermore, the
date code we had created a NSCalendarDate which has been long deprecated.

So we just don't handle date parsing. Even if there is a proper format for the pubDate,
It is an extra burden on the developer to insert a conforming date.
And we shouldn't want to rely on a potentially error-prone input field.

--

I left a `dateString` field so a developer may be able to access it if they want. But we could just remove it instead if we don't feel that's important.


zorgiepoo:
I merged in changes from master, and updated the test for the appcast items along with adding tests for the delta update items. This is a major change to selecting appcast items, so it may want to be checked in later.

pornel:
Looks good.

Issue #731, opened 2016-02-03
Requiring DSA signatures over insecure HTTP (Code Signing only over HTTPS)


kornelski:
I'm proposing removing ability to use _only_ Apple Code Signing as a method of verification when delivering applications over insecure HTTP. 

Apps using HTTP can still deliver updates signed with DSA. Without DSA apps will need to use HTTPS for the appcast and the update archive for Apple Code Signing to be accepted.

Reasons for the limitation:
- Code Signing verification increases attack surface area, because Sparkle has to decompress an untrusted archive first.
- Code Signing guarantees verification of developer identity, but it does not guarantee deep integrity check.
- HTTPS is not as burdensome as it was a decade ago. Modern TLS is fast, [HTTPS certificates are free](https://letsencrypt.org), and github pages offer free HTTPS hosting on github.io (and DSA signing remains an option for those stuck with insecure HTTP).
- Apple has deprecated insecure HTTP in the latest OS X/iOS, so we'd like to gradually move away from it too.

SevenBits:
@pornel This seems like a good idea. It's time to start hammering the nails in the coffin of insecure HTTP.

zorgiepoo:
Yeah we already deprecated it so we may as well remove supporting it. The alternative of removing it is simply requiring DSA everywhere but I suspect that may not go over well ;). Hope removing it is not tricky though.

zorgiepoo:
Been done for while

PR #760, opened 2016-02-18
Fixed appcast parsing issue

pornel:


pornel:
Should disabling JS (#755) go into 1.14.0? Just in case I've based that on earlier commits

zorgiepoo:
Looks good to me 👍. +1 for the unit test.

I don't see a downside of incorporating the JS change sooner than later. Although, I think we should address #731 soon too so maybe we ought to lump those together as far as security/incompatibility changes go.

pornel:
1.13 has a bug in support for feedurl set by delegates, so I'd like to release a fix ASAP, without waiting for #731 to be implemented.

zorgiepoo:
Sounds good.

PR #769, opened 2016-02-27
Update last update check date when the update driver finishes

zorgiepoo:
This fixes the issue where if an update alert is present for a long time,
and the user dismisses it, they may be prompted with another update alert
immediately after.

--
This is easy to reproduce with the Test Application when building in Debug mode, the update interval is set to 1 minute.


pornel:
Nice.

PR #773, opened 2016-03-01
Support XML namespace prefixes when parsing appcasts

### LINKED ISSUE #771: Sparkle needs an hardcoded namespace prefix to find sparkle:version in appcast.xml
It seems Sparkle only properly works if the `<enclosure>` `sparkle:version` attribute namespace prefix is hardcoded to `sparkle`.

Example from a - what seems valid - computer generated (Jersey JAXB) `appcast.xml`:

Here, in the `<rss>` element:

```
xmlns:ns3="http://www.andymatuschak.org/xml-namespaces/sparkle"
```

And later into an `<enclosure>` element:

```
ns3:version="0.0.42"
```

Will make my app spit this warning when it checks for update:

```
Sparkle: warning: <enclosure> for URL 'https://.../mybinary.zip' is missing sparkle:version attribute. Version comparison may be unreliable. Please always specify sparkle:version
```

pornel:
Fixes #771


pornel:
Good to merge?

Sparkle/SUAppcast.m
```diff
@@ -115,6 +98,31 @@ - (void)downloadDidFinish:(NSURLDownload *)__unused aDownload
...
+        }
+        dictionary[attrName] = [attribute stringValue];
+    }
+    return dictionary;
+}
+
+-(NSString *)sparkleNamespacedNameOfNode:(NSXMLNode *)node {
+    // XML namespace prefix is semantically meaningless, so compare namespace URI
+    if ([[node URI] isEqualToString:@"http://www.andymatuschak.org/xml-namespaces/sparkle"]) {
+        NSString *localName = [node localName]; assert(localName);
```
zorgiepoo: Did you mean to leave the assert in, maybe stick it on its own line?

Should https URL be handled here too? I know I may have changed it to https in an appcast for one of my apps. I don't really follow that much with this XML stuff though.

pornel: I've meant to have the assert (I've put it there, since non-nullable annotations caused warnings).

Namespace URI in XML is just a magic string compared codepoint-by-codepoint, never used to fetch anything, so we'll never change it. The fact that everyone settled on using http:// for NS URIs just has to be added to a long list of design problems with namespaces…

PR #780, opened 2016-03-14
Fix issues with unit tests

zorgiepoo:
I added public DSA file, automatic update checking, disabled system profiling
in the unit test's info.plist so that SUUpdater won't complain that DSA is missing,
and it won't prompt for update permission the first time.

I also made SUUpdater use the correct bundle in SUUpdaterTest


zorgiepoo:
Ping about status of this issue. Without this, running the unit tests don't succeed for me.

PR #783, opened 2016-03-19
Fix german l10n.

sebcode:
Update german translation.
Update german version of SUAutomaticUpdateAlert.xib (Button widths).
Update CheckLocalizations.swift for Swift 2.


pornel:
Dankeschön!

PR #786, opened 2016-03-23
Fix warnings

zorgiepoo:
Fixes -Wpartial-availability and nullability warnings.

More detail is in the commit messages.

Note that we cannot ignore partial availability warnings for subscripting because it is incredibly useful to have enabled otherwise.


pornel:
Looks good

PR #794, opened 2016-04-11
Don't follow symbolic links for file operations

zorgiepoo:
This fixes issues with updating modification/access time, and changing owner/group
where symbolic links used to be followed. Symbolic link tests are included.


pornel:
Good stuff

PR #795, opened 2016-04-11
Check the bundle's parent directory for writability too

zorgiepoo:
This is to decide if we should allow automatic downloading & installing of updates or not.

### DIFF
--- a/Sparkle/SUHost.m
+++ b/Sparkle/SUHost.m
@@ -71,7 +71,10 @@
     }
     
     // Can we automatically update in the background without bugging the user (e.g, with a administrator password prompt)?
-    return [[NSFileManager defaultManager] isWritableFileAtPath:self.bundlePath];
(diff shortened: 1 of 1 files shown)


pornel:
Makes sense.

I wonder if it'd be acceptable to write new bundle by replacing `Contents` directory in the `.app` instead of the whole `.app` dir.  That would reduce need for permissions to the parent directory.

zorgiepoo:
Interesting thought, but I really don't think the effort is worth the trouble for this edge case. A poorly structured app could have a file reside with Contents/ and we shouldn't use eg, `/Applications/Foo.app/` as a place for temporary storage / swapping which makes it possible for `/Applications/Foo.app/` to end up being broken.

Moreover, I think the expectation should be to ask for permission as Finder would if you were to remove the app or replace it.

Issue #750, opened 2016-02-15
Insane Disk Usage


bdkjones:
I received the following email from one of my customers today:

in the course of some unrelated digging around, I have discovered that ~/Library/Application Support/CodeKit/.Sparkle/ weighs in at 4.8GB. It contains 30 subdirectories with names like CodeKit 17908 and CodeKit 19033 4. What's the deal? Can I delete some of these? Or can you set up some garbage collection in the next release?

Can anyone explain why this might be the case or what I should do to combat it? 4.8 GB of wasted space on updates is a pretty big deal. CodeKit's download is only 50 MB.

pornel:
Old versions of Sparkle had that problem. We've stopped using that a while ago, and currently we create folders in ~/Library/Caches.

swrobel:
I have the same issue, but in ~/Library/Caches. Several of my apps have downloaded the same update over and over again into sequentially numbered folders.

pornel:
Could you share which apps did that? I suspect it may happen if there is mismatch between version declared in the appcast and version in the updated app. I wonder if we need to make Sparkle detect this error and do something about it.

zorgiepoo:
We should probably clean up our cache directory before we download a new update anyway, so it'd be impossible for data to accumulate regardless of what other issues there may be.

bdkjones:
That. Exactly. If a new download is started, it means nothing in the cache was useful, so the cache should be nuked.

zorgiepoo:
Download cleanup has been merged in (fixes #243 too).

My guess is it may have occurred from automatically/silently downloading updates and failing to install for some reason.

We changed writing from `~/Library/Caches/<appbundleid>/Sparkle/` to `~/Library/Caches/<appbundleid>/org.sparkle-project.Sparkle/`. Only reason for this change is because of potential collisions.

PR #797, opened 2016-04-19
Download cleanup

zorgiepoo:
This is to improve/resolve issue #750 

If you apply just the first commit for the Test App, you can reproduce the issue that the 2nd commit fixes pretty easily. Just download an update and quit before deciding to install it. Do it again, and you'll see that data accumulates in the user's cache directory.


pornel:
That's nice. Less complicated than I expected :)

PR #798, opened 2016-04-19
Upgrade warning to alert for signing update over HTTP with no DSA key

zorgiepoo:
I have seen developers ignore the log warning :(, so I figure we should upgrade it to an alert now.


pornel:
I was afraid to pull the trigger on this one, but you're right we need to push for this to be fixed.

PR #800, opened 2016-04-28
Updated sponsored by text and links

jdorfman:
We are migrating the Bandwidth Hog brand with MaxCDN.


jdorfman:
Thanks @jakepetroules!

Issue #347, opened 2014-06-30
New icon(s)


pornel:
- Aqua bubbles are passé
- 10.10 introduces new look'n'feel of icons
- We don't have high-res original for the current icon

Also there's a problem that Autoupdate.app (formely finish_installation.app) may sometimes be user-visible, but Sparkle's icon is not immediately obvious or recognisable by an average (non-technical) user, so it causes confusion when it pops up.

We may need a new icon for the project, and perhaps a special version of the icon for the Autoupdate.app.

For the autoupdate the icon shouldn't be too branded, it should look generic, like a system utility, installer or such., so if user sees the Autoupdate.app somewhere they'll guess it's part of "plumbing" and not a user-facing app.

Mecki78:
Why don't you just show the app icon instead? You can tell OS X which icon to show there, just as you can tell it which message to display (except for a fixed part that the system will always append to your message):

    const char * prompt = ...;
    const char * iconPath = ...;
    
    AuthorizationItem authItems[2] = {
    	{
    		.name = kAuthorizationEnvironmentPrompt,
    		.valueLength = strlen(prompt),
    		.value = (void *)prompt,
    		.flags = 0
    	}, {
    		.name = kAuthorizationEnvironmentIcon,
    		.valueLength = strlen(iconPath),
    		.value = (void *)iconPath,
    		.flags = 0
    	}
    };
    
    AuthorizationItemSet authEnvironment = {
    	.count = 2,
    	.items = authItems
    };
    
    AuthorizationItem sessionOwnerAuthItem = {
    	.name = kAuthorizationRightExecute,
    	.valueLength = 0,
    	.value = nil,
    	.flags = 0
    };
    
    AuthorizationRights rightsRequested = {
    	.count = 1,
    	.items = &sessionOwnerAuthItem
    };
    
    OSStatus err = AuthorizationCreate(
    	&rightsRequested, &authEnvironment,
    	kAuthorizationFlagPreAuthorize
    		| kAuthorizationFlagExtendRights
    		| kAuthorizationFlagInteractionAllowed,
    	NULL
    );

It's only important that the icon is **exactly** 32x32 pixels, the icon must be accessible by a system user that doesn't seem to be root (if you place it into a folder that only the current user has access to, it will not show on my system), the system doesn't like spaces in the path (no idea why!) and it doesn't like very long paths to begin with (not sure about the limit, stay below 128 chars) and ideally it should be a PNG (works best).

[...]

PR #801, opened 2016-04-29
Update app icon for Autoupdate.app (#347)

1024jp:
ref. #347

Hey, I finished up the new app icon.

I've actually changed it quite a lot from the first draft ver, more than I expected.
The foreground and background colors were flipped in order to make it more like "inner process".
Furthermore, I even omitted the sparkle symbols I added. Now it looks super generic.

(I also can revert the sparkle symbols If you think it's better, as shown below.)

Good to merge or some tweak required?


pornel:
Oh, it is nicer indeed! I think I slightly prefer the sparkly version. 

Have you experimented with using some yellow accents? As a homage to Sparkle's previous palette.

1024jp:
Ok, so, I've swaped images with the sparkly version.

> Have you experimented with using some yellow accents? As a homage to Sparkle's previous palette.

Yes actually. I've tried using yellow at first.
However, it didn't work well because the combination of blue and yellow is too showy for such a helper app. Considering that Sparkle will be built in another applications, IMO, the icon of Autoupdate.app should not disterb the atmosphere of client application.

pornel:
Great. Thank you.

RobFog:
Great improvement. The old icon always struck me as a bit odd.

1024jp:
Thanks for the merge 😃
I'm glad that my work will be used in this great framework.

zorgiepoo:
Xcode gives warnings that "The app icon set "AppIcon" has 3 unassigned children" - does this need to be addressed?

1024jp:
@zorgiepoo 
Ops, I just forgot assign them in Xcode.
I've made a pull-request fixing it at #804.

Issue #803, opened 2016-04-30
Compress icon


pornel:
AppIcon.icns file built by Xcode, when it includes the 512x512@2x variant, takes a massive 630KB.

With a combination of pngquant2 & https://github.com/pornel/libicns I can reduce full-size AppIcon.icns down to 82KB.

With max 256px icon it takes 123KB from Xcode, and 43KB when optimized.

Should we switch from asset catalog to prebuilt AppIcon.icns?

zorgiepoo:
Are the PNGs we are including not already compressed or are they uncompressed, and after the icns is made, compression is done?

I suppose this means we'll lose Xcode integration, which could for example warn us if we're doing something wrong. Perhaps not a big deal though.

Is libicns trustworthy / will stay up to date in the future? Was it updated around the time Icon Composer was deprecated?

pornel:
The size difference is because Xcode generates icons with weakly-compressed RGBA PNGs, and I use Zopfli-compressed PNG8.

libicns seems to generate valid icons that work in Finder and Preview.

zorgiepoo:
It has support for 512x512@2x, yes? And it shouldn't have 1024x1024 since that was dropped, if I recall.

pornel:
AFAIK that's technically the same, icon with `ic10` ID.

pornel:
So currently the icon is larger than Sparkle binary. For large apps it may not matter, but the icon increased size of ImageOptim's update by 40%!

I see two quick solutions:

1. don't include all icon sizes in the icns file, especially the largest. I think 128px@2x is the maximum we need in practice.

2. Add and use optimized icns file to be used instead of the iconset. (leaving iconset in the repo in case it needs to be rebuilt)

What are your preferences?

zorgiepoo:
So I'm a fan of not including size representations we don't need and using your optimized icns version. I just don't want warnings popping up in Xcode about missing representations. A 32x32 representation may be needed for the auth dialog by the way.

jakepetroules:
> I just don't want warnings popping up in Xcode about missing representations.

Yes, same...

PR #804, opened 2016-04-30
Assign newly added icon images in asset catalog (#801)

1024jp:
ref. #801


pornel:
I thought you've deliberately omitted these, because it makes the icon file quite large, #803

jakepetroules:
I don't think 630 KB is much to worry about in this day and age. We can look into further optimizations if it's really necessary... continue discussion on #803.

sr55:
On it's own no, but consider an app that has 1m downloads. Then suddenly we are talking 600GB of bandwidth.

PR #807, opened 2016-05-11
Simpler conditions for security config check

pornel:
I've received privately a question about logic of this check, and realised it's hard to follow. I've split and reordered it for clarity.

### DIFF
--- a/Sparkle/SUUpdater.m
+++ b/Sparkle/SUUpdater.m
@@ -125,25 +125,31 @@ static NSString *const SUUpdaterDefaultsObservationContext = @"SUUpdaterDefaults
     BOOL hostIsCodeSigned = [SUCodeSigningVerifier hostApplicationIsCodeSigned];
     NSURL *feedURL = [self feedURL];
     BOOL servingOverHttps = [[[feedURL scheme] lowercaseString] isEqualToString:@"https"];
-    if (!isMainBundle && !hasPublicDSAKey) {
-        [self showAlertText:@"Insecure update error!"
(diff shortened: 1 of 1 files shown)


zorgiepoo:
I actually think the code in `master` is better to reason about. I can look at a particular branch without needing more context in the procedure as opposed to the changes provided here. Returning early in this procedure makes the coder *harder* to understand.

I'm all for adding comments though.. Perhaps the most confusing thing is the is main bundle check and why we care about it.

pornel:
```c
 if (isMainBundle && !(hasPublicDSAKey || hostIsCodeSigned)) { 
 } else if (isMainBundle && !hasPublicDSAKey && !servingOverHttps) {
```

I find reasoning about this very hard. Because it's a cascade of else/if, I need to notice relationship between `!hasPublicDSAKey` and `!(hasPublicDSAKey || hostIsCodeSigned)` and how subtly different it is if only `!hostIsCodeSigned` was in the previous condition.

zorgiepoo:
I'll just note that I mentally translate the first line into `isMainBundle && !hasPublicDSAKey && !hostIsCodeSigned`

The thing is, the code in `master` doesn't *have* to use else if's. It only does that so that less alerts are displayed, but otherwise it's not very significant. I find having this "redundancy" *better* to reason about the code. Bugs from code are prone to happen when order is significant in code, when code has side effects, when portions of code need outside context, when it had to be re-arranged/re-ordered for it to still be correct.

Just my take.

pornel:
How about now?

zorgiepoo:
Yeah, I like it.

PR #811, opened 2016-05-17
bsdiff fixes

### LINKED ISSUE #810: Invalid Argument from BinaryDelta
I'm getting an Invalid Argument error from the BinaryDelta on the db in my app, but its inconsistent (sometimes it actually works):

Generating delta...
✏️  Updated /Contents/Info.plist
✏️  Updated /Contents/Resources/Main.storyboardc/Info.plist
BinaryDelta: ../Builds/Dev.app/Contents/Resources/db.sqlite: Invalid argument

I run two different BinaryDelta's concurrently (background process) for two outdated apps to help speed up delta generations. Is it not possible to run BinaryDelta concurrently maybe? It looks like another user had this issue years ago, but it wasn't resolved:

https://answers.launchpad.net/sparkle/+question/108082

zorgiepoo:
1. Fix and don't hide compiler warnings for bspatch or bsdiff (can we please decide not to do this for files in the future?)
2. Use `FILE*` fread/fwrite functions instead of read() and write() to resolve #810
3. Change `err()`/`errx()` to use `warn()`/`warnx()`, cleanup, and return. Since we are invoking bsdiff/bspatch ourselves, we want to handle failures gracefully.


jakepetroules:
Can we upstream these changes to bsdiff?

zorgiepoo:
As far as I can tell, bsdiff isn't maintained and there are only significant forks out there (ours is a significant fork too, even prior to these changes). One of the reasons why I want to look into xdelta at some point.

pornel:
LGTM

mdhari:
Just FYI, this does handle files over 2 GB, but the size of the patch is quite large. I mentioned in #810 that I generated a 220.2 MB patch between a 2.15 GB and 2.16 GB file. I tested what happens if I split the 2 GB into 1 GB each. The patch generated was 31.8 MB. Interesting stuff. 

At the very least, the "Invalid Argument" error is gone.

PR #817, opened 2016-06-07
Improve Authorization Usage

### LINKED ISSUE #816: Improve AuthorizationExecuteWithPrivilegesAndWait Usage
Taking the current code in context:

``` objc
static BOOL AuthorizationExecuteWithPrivilegesAndWait(AuthorizationRef authorization, const char *executablePath, AuthorizationFlags options, char *const *arguments)
{
    sig_t oldSigChildHandler = signal(SIGCHLD, SIG_DFL);
    BOOL returnValue = YES;
    if (AuthorizationExecuteWithPrivileges(authorization, executablePath, options, arguments, NULL) == errAuthorizationSuccess)
    {
        int status;
        pid_t pid = wait(&status);
        if (pid == -1 || !WIFEXITED(status) || WEXITSTATUS(status) != 0)
            returnValue = NO;
    }
    else
        returnValue = NO;

    signal(SIGCHLD, oldSigChildHandler);
    return returnValue;
}
```

This code runs on a secondary (non-main) thread inside of Autoupdate and not inside the Sparkle framework itself.
1. We have duplicated authorization code in `SUFileManager` and `SUGuidedPackageInstaller` (something I already have code for resolving..) and poor error reporting.
2. It tries to back up the signal handler that a caller may have set up and restores it later, but I don't think that is something we should concern ourselves for our use case and attempting to deal with that may have some implications across different threads.
3. This code does not handle restarting `wait()` when encountering `EINTR`
4. This code waits for _any_ child to terminate. It may be possible for the main or other thread to spawn another child behind our back, although I'm not sure how likely this is to occur. The only workaround I see to this is to use `waitpid` - but the only way to use this is for the child to report its pid back to the parent via `stdout`, which may mean requiring an additional wrapper tool.

zorgiepoo:
This is aimed to resolve #816. I found a way to retrieve the child pid without having to create a wrapper tool, by using fcntl. For practical purposes, this is intended to fix issues (or report errors better) with the guided installer not working. In my fork, I ran into an issue where the installer aborting too early because the child wasn't properly waited, which this aims to fix.

The installer unit test should work if it's ran as root.

### DIFF
--- a/Sparkle/SUFileManager.m
+++ b/Sparkle/SUFileManager.m
@@ -15,36 +15,6 @@
 
 static char SUAppleQuarantineIdentifier[] = "com.apple.quarantine";
 
-// Authorization code based on generous contribution from Allan Odgaard. Thanks, Allan!
-static BOOL AuthorizationExecuteWithPrivilegesAndWait(AuthorizationRef authorization, const char *executablePath, AuthorizationFlags options, char *const *arguments)
-{
-    BOOL returnValue = YES;
-
-    FILE *pipe = NULL;
-#pragma clang diagnostic push
-    // In the future, we may have to look at SMJobBless API to avoid deprecation. See issue #558
-#pragma clang diagnostic ignored "-Wdeprecated-declarations"
-    if (AuthorizationExecuteWithPrivileges(authorization, executablePath, options, arguments, &pipe) == errAuthorizationSuccess)
-#pragma clang diagnostic pop
-    {
-        pid_t childPid = fcntl(fileno(pipe), F_GETOWN, 0);
-        int status = 0;
-        pid_t pid = -1;
-        do {
-            pid = waitpid(childPid, &status, 0);
-        } while (pid == -1 && errno == EINTR);
-        
-        if (pid == -1 || !WIFEXITED(status) || WEXITSTATUS(status) != 0) {
-            returnValue = NO;
-        }
-        fclose(pipe);
-    } else {
-        returnValue = NO;
-    }
-    
-    return returnValue;
-}
-
 static BOOL SUMakeRefFromURL(NSURL *url, FSRef *ref, NSError **error) {
 
     char path[PATH_MAX] = {0};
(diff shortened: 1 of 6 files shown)


Sparkle/SUFileManager.m
```diff
@@ -100,29 +76,101 @@ - (instancetype)fileManagerByPreservingAuthorizationRights
...
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+    if (AuthorizationExecuteWithPrivileges(_auth, executablePath, kAuthorizationFlagDefaults, arguments, &pipe) != errAuthorizationSuccess) {
+#pragma clang diagnostic pop
+        if (error != NULL) {
+            *error = [NSError errorWithDomain:SUSparkleErrorDomain code:SUAuthenticationFailure userInfo:@{ NSLocalizedDescriptionKey:@"Failed to run authorization tool." }];
+        }
+        return NO;
+    }
+    
+    pid_t childPid = fcntl(fileno(pipe), F_GETOWN, 0);
```
pornel: is it safe to rely on it? Could the pipe be owned by some intermediary system process that doesn't quit?

Sparkle/SUFileManager.m
```diff
@@ -100,29 +76,101 @@ - (instancetype)fileManagerByPreservingAuthorizationRights
...
+    int status = 0;
+    
+    pid_t waitResult;
+    do {
+        waitResult = waitpid(childPid, &status, 0);
+    } while (waitResult == -1 && errno == EINTR);
+    
+    fclose(pipe);
```
pornel: Do you have to read from the pipe? With NSTask I've had tasks stuck forever, because their output buffer was full and were waiting for the parent to drain it.

Sparkle/SUFileManager.m
zorgiepoo: ack, after looking again this is actually no good. `fcntl()` returns 0; not sure how this slipped by me.

Sparkle/SUFileManager.m
zorgiepoo: This code doesn't read from the pipe, but seeing as `fcntl` didn't work as I thought, I see no other choice but to execute a tool that writes the pid onto the pipe that we read from. This would just be the first thing the tool does.

PR #818, opened 2016-06-09
Improve Authorization Usage (Take 2)

zorgiepoo:
Second attempt for improving authorization (#816) usage (first failed one being #817). This uses another tool we include which reports back a pid immediately after launch, which the parent then uses to wait on. Chromium uses [this approach](https://src.chromium.org/svn/trunk/src/base/mac/authorization_util.mm) as well. Because of this change, we don't shell to 'cp', 'touch', 'chown' etc, but rather shell to our own tool which uses our file manager code. There may be a security concern of someone swapping out the executable, although I think someone could already currently alter the strings in the Autoupdate binary.

A third alternative approach may be to do something like shelling to `sh` and have it first echo the pid and then execute whatever we want. This would require minimal changes, but it may not be a good approach.

(Fourth is not using the deprecated authorization function but that may require too much architectural changing).


pornel:
I like using our own code instead of falling back to `cp` 👍 

The tests have link errors: `_SUFileOpChangeOwnerAndGroupCommand` etc.

zorgiepoo:
Updated to fix linking errors.

pornel:
Good refactoring :)

PR #823, opened 2016-06-16
update Sparkle framework headers to use modules if modules are available

nivekkagicom:
When using modules building sparkle can produce a number of warnings. 

While building module 'Sparkle' imported from SparkleUpdater.h:13:
In file included from <module-includes>:1:
SparkleMaster/Sparkle.framework/Headers/Sparkle.h:12:1: warning: treating #import as an import of module 'Cocoa' [-Wauto-import]
#import <Cocoa/Cocoa.h>
^~~~~~~~~~~~~~~~~~~~~~~
@import Cocoa;

Change the headers to conditionally use @import if modules are available otherwise fall back to the #import of the framework umbrella header.

### DIFF
--- a/Sparkle/SUAppcast.h
+++ b/Sparkle/SUAppcast.h
@@ -9,7 +9,11 @@
 #ifndef SUAPPCAST_H
 #define SUAPPCAST_H
 
+#if __has_feature(modules)
+@import Foundation;
+#else
 #import <Foundation/Foundation.h>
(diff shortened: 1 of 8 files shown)


pornel:
Thanks for the PR.

MaddTheSane:
Another way it could go was to disable the related warning by passing `-Wno-auto-import` to the compiler.

PR #825, opened 2016-06-17
Changes for Xcode 8

jerrykrinock:
This first change was necessary to build the Sparkle framework target in Xcode 8.0:

* In ConfigCommon.xcconfig, removed references to CLANG_WARN_OBJC_RECEIVER_WEAK and -Wno-receiver-is-weak, because apparently these had been deprecated and are now no longer recognized.  Unfortunately, I could not find any documentation on this, but it seems reasonable.

The remaining changes are to clear warnings from the static analyzer in Xcode 8.0:

* In SUAppcast.m, changed to method which can accept nil.
* In SUOperatingSystem.h, added a check for nil which probably never happens.
* In SUUpdateAlert.m, swapped receiver with parameter in -isEqualToString, 3 places, so that nonnull parameter could never be nil.
* In SUUpdater.m, added NSLocalizedString() wrappers around 8 recently-added (in the last few years, I think) user-facing warning strings.

### DIFF
--- a/Sparkle/SUAppcast.m
+++ b/Sparkle/SUAppcast.m
@@ -128,7 +128,7 @@
         if (!attrName) {
             continue;
         }
-        [dictionary setObject:[attribute stringValue] forKey:attrName];
+        [dictionary setValue:[attribute stringValue] forKey:attrName];
     }
     return dictionary;
 }
--- a/Sparkle/SUAppcast.m
+++ b/Sparkle/SUAppcast.m
@@ -128,7 +128,10 @@
         if (!attrName) {
             continue;
         }
-        [dictionary setValue:[attribute stringValue] forKey:attrName];
+        NSString *stringValue = [attribute stringValue];
+        if (stringValue) {
+            [dictionary setObject:stringValue forKey:attrName];
+        }
     }
     return dictionary;
 }
--- a/Sparkle/SUOperatingSystem.m
+++ b/Sparkle/SUOperatingSystem.m
@@ -20,11 +20,14 @@
     {
         NSOperatingSystemVersion version = { 0, 0, 0 };
         NSURL *coreServices = [[NSFileManager defaultManager] URLForDirectory:NSCoreServiceDirectory inDomain:NSSystemDomainMask appropriateForURL:nil create:NO error:nil];
-        NSDictionary *dictionary = [NSDictionary dictionaryWithContentsOfURL:[coreServices URLByAppendingPathComponent:@"SystemVersion.plist"]];
-        NSArray *components = [ [dictionary objectForKey: @"ProductVersion"] componentsSeparatedByString:@"."];
-        version.majorVersion = components.count > 0 ? [ [components objectAtIndex:0] integerValue] : 0;
-        version.minorVersion = components.count > 1 ? [ [components objectAtIndex:1] integerValue] : 0;
-        version.patchVersion = components.count > 2 ? [ [components objectAtIndex:2] integerValue] : 0;
+        NSURL *url = [coreServices URLByAppendingPathComponent:@"SystemVersion.plist"] ;
+        if (url) {
+            NSDictionary *dictionary = [NSDictionary dictionaryWithContentsOfURL:url];
+            NSArray *components = [ [dictionary objectForKey: @"ProductVersion"] componentsSeparatedByString:@"."];
+            version.majorVersion = components.count > 0 ? [ [components objectAtIndex:0] integerValue] : 0;
+            version.minorVersion = components.count > 1 ? [ [components objectAtIndex:1] integerValue] : 0;
+            version.patchVersion = components.count > 2 ? [ [components objectAtIndex:2] integerValue] : 0;
+        }
         return version;
     }
 #endif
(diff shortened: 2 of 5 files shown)


Sparkle/SUAppcast.m
```diff
@@ -128,7 +128,7 @@ - (NSDictionary *)attributesOfNode:(NSXMLElement *)node
         if (!attrName) {
             continue;
         }
-        [dictionary setObject:[attribute stringValue] forKey:attrName];
+        [dictionary setValue:[attribute stringValue] forKey:attrName];
     }
```
zorgiepoo: We should still use setObject:forKey: but just make sure stringValue doesn't return nil.

jerrykrinock: Hello, zorgiepoo.  The issue here is to clear clang warnings.  What matters is not whether -stringValue can _actually_ return nil in this case, but whether the clang static analyzer in Xcode 8 _thinks_ it can return nil :))

As I've always understood it, -setObject:forKey: does not accept nil as the first param, but setValue:forKey: does.  That is why changing -setObject: to -setValue: clears this particular warning.

Alternatively, we could assign stringValue to a local variable and test it for nil.  I just tested that, and clang is happy that way too.  What do you think?

zorgiepoo: > Alternatively, we could assign stringValue to a local variable and test it for nil. I just tested that, and clang is happy that way too. What do you think?

Yeah, this is what I meant, but I was in a rush. Thank you for the pull request.

jerrykrinock:
OK, zorgiepoo.  In the latest commit, we check for nil for on a local variable, and continue to use -setObject:forKey:.

Oddly, comically unless I’m missing something, clang was happy after only inserting the local variable.  We don’t need to test it for nil to clear the static analyzer warning!  But we're checking for nil anyhow :)

Sparkle/SUOperatingSystem.m
```diff
@@ -20,11 +20,14 @@ + (NSOperatingSystemVersion)operatingSystemVersion
     {
         NSOperatingSystemVersion version = { 0, 0, 0 };
         NSURL *coreServices = [[NSFileManager defaultManager] URLForDirectory:NSCoreServiceDirectory inDomain:NSSystemDomainMask appropriateForURL:nil create:NO error:nil];
-        NSDictionary *dictionary = [NSDictionary dictionaryWithContentsOfURL:[coreServices URLByAppendingPathComponent:@"SystemVersion.plist"]];
-        NSArray *components = [ [dictionary objectForKey: @"ProductVersion"] componentsSeparatedByString:@"."];
-        version.majorVersion = components.count > 0 ? [ [components objectAtIndex:0] integerValue] : 0;
-        version.minorVersion = components.count > 1 ? [ [components objectAtIndex:1] integerValue] : 0;
-        version.patchVersion = components.count > 2 ? [ [components objectAtIndex:2] integerValue] : 0;
+        NSURL *url = [coreServices URLByAppendingPathComponent:@"SystemVersion.plist"] ;
+        if (url) {
```
zorgiepoo: I think we should assert(url != nil) instead of using an if, since if this fails, the logic in the calling code is going to be broken..

zorgiepoo:
I think we have a `SULocalizedString` macro that we use everywhere rather than `NSLocalizedString`

> Oddly, comically unless I’m missing something, clang was happy after only inserting the local variable. We don’t need to test it for nil to clear the static analyzer warning! But we're checking for nil anyhow :)

Yep, probably because the local variable is treated as `_Null_unspecified` rather than `_Nonnull` or `_Nullable`, which works here.

jerrykrinock:
Ooops, I'd forgotten about `SULocalizedString`.  Those are all changed now.  And I put in the `assert()` instead of testing for nil, for the reason  you stated.  Again, just inserting the local variable was enough to satisfy clang, though.  Maybe next year clang will disallow that `_Null_unspecified ` local variables loophole, but it all works with no warnings now.

zorgiepoo:
We use xcconfig files rather than store build settings in the Xcode project file. I'm not sure which new settings we actually have to specify, and not sure if we should set `SWIFT_VERSION` to 3.0 yet - do the unit tests need to change? We may also want to continue developing with Xcode 7 because 8 is in beta.

jerrykrinock:
The few and simple changes in this pull request cause the Framework target (to be clear, the target named *Sparkle*) to Build and Analyze in Xcode 8 without warnings or errors.  These changes should be backward-compatible with Xcode 7.  They are going to be necessary sooner or later.

Indeed, the Swift conversion is more complicated.  Once the `SWIFT_VERSION` is changed to 3.0, those unit test .swift files must be converted to Swift 3.0, and then they will no longer run in Xcode 7.  So I'd think that we probably don't want to do that yet because it would create a hurdle for the many developers who have not taken the Xcode 8 plunge. The temporary down-side is that anyone testing the unit tests in Xcode 8 will see warnings.  But those warnings do not appear in a Sparkle "consumer" project such as mine, because my project includes only the *Sparkle* Framework target, which is *not* dependent on the *UI Tests* or *Sparkle Unit Tests* target.

Just for fun, I tried the Swift 3.0 conversion on Sparkle.  After about a half hour of fussing around, and one Xcode crash, I was able to once successfully build the *UITests* scheme, which runs the *UI Tests* target, , but not yet *Distribution* scheme, which runs the the *Sparkle Unit Tests* target.

jerrykrinock:
By the way, yesterday and the day before when I was experimenting with those Swift tests, I forgot to start a new branch, and I think those experimental changes got pushed to github.  I'm sorry if those showed up in the Pull Request and that's why you asked about Swift yesterday, @zorgiepoo.  A few hours ago I discovered the mistake, so I rolled back and did a push --force.  The Files Changed in this Pull Request now shows only the changes I originally intended.

pornel:
Thanks @jerrykrinock 

I don't know if @zorgiepoo has any more issues, but I'll merge it and we can fix stuff later if needed.

PR #826, opened 2016-06-18
Fix release build warning: skipping copy phase strip, binary is code signed

iosdeveloper:
Noticed this in our [Travis CI build log](https://travis-ci.org/matryer/bitbar/builds/138554729). Fixes a release build warning during copy phase, failing to strip debug symbols due to code signing. This is a default setting for project templates since Xcode 6.3 (see https://www.cocoanetics.com/2015/04/skipping-copy-phase-strip/). I confirmed that this gets added when creating a new project now.

### DIFF
--- a/Sparkle.xcodeproj/project.pbxproj
+++ b/Sparkle.xcodeproj/project.pbxproj
@@ -2232,6 +2232,7 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = FA1941CC0D94A70100DD942E /* ConfigCommonRelease.xcconfig */;
 			buildSettings = {
+				COPY_PHASE_STRIP = NO;
 			};
 			name = Release;
 		};


zorgiepoo:
Sparkle is not set up for code signing. This may not be too relevant to the project. Have you adjusted the project by switching on code signing?

We also use xcconfig files rather than modifying build settings in Xcode.

iosdeveloper:
@zorgiepoo No, I have not touched the Sparkle project. This appears when building the framework from source with a code signed app. Happy to move this to ConfigCommonRelease.xcconfig if you prefer.

zorgiepoo:
Building the framework from source shouldn't differ much from downloading the framework distributed from Sparkle's website. When the framework is built, it shouldn't be code signed, but when your app is built it will then copy the framework and be responsible for code signing everything in the app.

iosdeveloper:
@zorgiepoo Maybe building from the command line is different and the `CODE_SIGN_IDENTITY` gets also passed to the Target Dependencies.

zorgiepoo:
Maybe that's what's happening, but it sounds like `CODE_SIGN_IDENTITY` shouldn't be passed to every target. I don't imagine how building in xcodebuild would differ from building from within Xcode.

pornel:
I'm going to close this for reasons Zorg mentioned above.

PR #836, opened 2016-07-07
Don't use reachability as a preflight check

zorgiepoo:
This removes the reachability code that we have for background scheduled updates.
Apple (and others) don't recommend using reachability to test if a network connection is online.

See:
http://stackoverflow.com/questions/7178330/what-is-the-advantage-of-using-reachability
https://www.mikeash.com/pyblog/friday-qa-2013-06-14-reachability.html (just the comments)
Apple 2012/2011 WWDC videos on networking

Also, our current reachability code is rather illogical:

* reachabilityResult is always NO
* isNetworkReachable is always YES
* flags is ignored
* horrible use of dispatch_sync & putFeedURLIntoDictionary is terrifying
* NSException handling shouldn't be done

Removing this code which wasn't even coded properly, and which would still be misusing the reachability API anyway, also leads to removing some multithreading races and simpler logic :)


zorgiepoo:
Also - I think we can afford to re-try the next update interval check. Reachability is complicated and it could be beneficial perhaps, but not in the way we were trying to use it.

pornel:
Deleting code is the best :)

Issue #846, opened 2016-07-19
Couldn’t find the relauncher (Xamarin.Mac)


chbach:
Hi, I’m using the Sparkle Framework with my Xamarin.Mac project and the SparkleSharp Bindings (https://github.com/rainycape/SparkleSharp) – It all works fine until it starts to replace the old app with the new one. The output is: 

```sh
2016-07-19 15:18:56.704 PINGOMacRemote[37403:1648642] Sparkle: ===== PINGO Remote.app =====
2016-07-19 15:18:56.705 PINGOMacRemote[37403:1648642] Sparkle: Extracting using '/usr/bin/ditto' '-x' '-k' '-' < '/Users/christoph/Library/Caches/de.upb.pingo.remote.mac/Sparkle/PINGO Remote 1.1/PINGO Remote.zip' '/Users/christoph/Library/Caches/de.upb.pingo.remote.mac/Sparkle/PINGO Remote 1.1'
2016-07-19 15:18:59.775 PINGOMacRemote[37403:1647145] Sparkle: Error: Beim erneuten Starten von PINGO Remote ist ein Fehler aufgetreten. Die neue Version von PINGO Remote steht beim nächsten Programmstart zur Verfügung. Couldn't find the relauncher (expected to find it at (null)) (URL (null))
``` 

Do I have to set additional config variables or something?

pornel:
This should just work if Sparkle was bundled properly. There should be Autoupdate.app inside Sparkle.framework bundle, but it appears to be missing in your case.

chbach:
Well it is exactly as you describe it. Not sure how that happens, because the Autoupdate.app is in the Sparkle.framework before building and packaging the app. However, it is not after that. I manually put the Sparkle.framework in the .app package after building and now it works. Any clues how that can happen?

chbach:
Well apperently SparkleSharp had *.app in its .gitignore. The issue is sorted out. Thank’s for your help.

PR #847, opened 2016-07-19
Better error message when relaunch tool bundle is not found.

pornel:
Improves log of #846


zorgiepoo:
Looks good, although I don't quite get the point of the `relaunchPath` instance variable, which is only referenced in another method (`-installerForHost:failedWithError:`) that I don't even see used anywhere.

pornel:
Yeah, I don't see point either, but I've kept it just in case delegates or 3rd party subclasses mess with it.

PR #858, opened 2016-08-09
Make sign_update work when Bash is not the default

catnapgames:
The sign_update script fails when Bash is not the default shell. This patch fixes it and allows the script to run on a default Ubuntu system, for example.


pornel:
Thanks

Issue #832, opened 2016-07-01
App Translocation


zorgiepoo:
Let's have this be the general issue for app translocation / gatekeeper path randomization in the upcoming macOS 10.12.. Note this more has to do with the initial distribution of a developer's application. We may not end up having to actually change anything in Sparkle, but it may be worthwhile thinking about this.

This blog post was brought up recently: http://weblog.rogueamoeba.com/2016/06/29/sierra-and-gatekeeper-path-randomization/ - if as a developer you also feel that dmg's provide a subpar user experience and worry about users not moving your app, then I strongly suggest duping/filing a rdar to Apple as mentioned in the blog post.

Going through the list of apps that use Sparkle, I notice that most or around half of the apps don't use dmg's, so this could impact many developers.

Sparkle's website documentation has been updated to reflect the new changes, in the event that Apple decides to stick with forcing dmg's upon us.

[Note: before someone brings this up, the xip format does not avoid translocation like a signed dmg does, at least not in the current macOS 10.12 beta, and I think this is intentional]

danielchasehooper:
@zorgiepoo I've been thinking about this for principleformac.com. We distribute via a zip, so I've been wondering if there is anything we need to do to avoid trouble in Sierra, or if Sparkle will able to transparently handle this complication for us.

zorgiepoo:
This is documented on Sparkle's website now (see [step 0](https://sparkle-project.org/documentation/)). In short: ship a dmg and code sign the dmg. This is unless Apple adds some sort of [other way](http://weblog.rogueamoeba.com/2016/06/29/sierra-and-gatekeeper-path-randomization/) to avoid the path randomization during the beta. If you think they should, file a rdar :-)!

If Sparkle tried to handle it transparently, that might go in direct opposition to Apple's intent (eg: detect that application is translocated, try to find if the same app is in ~/Downloads/, and clear it from being translocated).

vitor251093:
App Translocation can be avoided if the `com.apple.quarantine` property is removed from the app after it's uncompressed. 
http://lapcatsoftware.com/articles/app-translocation.html

I mean, can't Sparkle remove that property? In the terminal it can be done with a single command:
`xattr -r -d com.apple.quarantine <APP PATH>`

As I last resource, couldn't Sparkle just run that command after extracting the app?

pornel:
Sparkle removes that from updates, but can't remove it when the app is launched from already-quarantined location.

PR #859, opened 2016-08-10
Delay check for -isRunningOnReadOnlyVolume until we know if update is available

jhammer:
This PR postpones the check for `-isRunningOnReadOnlyVolume` until
_after_ Sparkle has already discovered a valid update and the user has
chosen to download the update (either via the UI or via automatic
upgrade driver). I think there are at least two advantages to doing
this check later rather than sooner:
1. For apps that might be affected by Gatekeeper Path Randomization
   (#832), this change allows the app to still get delegate callbacks to
   know when an update is available (even if it can’t be applied). If it
   chooses, the app could then prompt the user to take some action. For
   example, moving the app out of ~/Downloads—or offering to do it on the
   user’s behalf via something like LetsMove.framework (which now works
   again in 10.12 beta 3). Without this PR, the update fails before
   the app has a chance to let the user know they’re missing out on an
   update.
2. This PR allows `-[SUUpdater checkForUpdateInformation]` to
   function properly even if app on read-only volume.

### DIFF
--- a/Sparkle/SUBasicUpdateDriver.m
+++ b/Sparkle/SUBasicUpdateDriver.m
@@ -43,14 +43,20 @@
 @synthesize tempDir;
 @synthesize relaunchPath;
 
+- (BOOL)verifyHostVolumeWritable
+{
+    if ([self.host isRunningOnReadOnlyVolume])
+    {
+        [self abortUpdateWithError:[NSError errorWithDomain:SUSparkleErrorDomain code:SURunningFromDiskImageError userInfo:@{ NSLocalizedDescriptionKey: [NSString stringWithFormat:SULocalizedString(@"%1$@ can't be updated when it's running from a read-only volume like a disk image or an optical drive. Move %1$@ to your Applications folder, relaunch it from there, and try again.", nil), [self.host name]] }]];
+        return NO;
+    }
+    
+    return YES;
+}
+
 - (void)checkForUpdatesAtURL:(NSURL *)URL host:(SUHost *)aHost
 {
     [super checkForUpdatesAtURL:URL host:aHost];
-	if ([aHost isRunningOnReadOnlyVolume])
-	{
-        [self abortUpdateWithError:[NSError errorWithDomain:SUSparkleErrorDomain code:SURunningFromDiskImageError userInfo:@{ NSLocalizedDescriptionKey: [NSString stringWithFormat:SULocalizedString(@"%1$@ can't be updated when it's running from a read-only volume like a disk image or an optical drive. Move %1$@ to your Applications folder, relaunch it from there, and try again.", nil), [aHost name]] }]];
-        return;
-    }
 
     SUAppcast *appcast = [[SUAppcast alloc] init];
 
@@ -190,6 +196,11 @@
     [[NSNotificationCenter defaultCenter] postNotificationName:SUUpdaterDidFindValidUpdateNotification
                                                         object:self.updater
                                                       userInfo:@{ SUUpdaterAppcastItemNotificationKey: self.updateItem }];
+    
+    if (![self verifyHostVolumeWritable]) {
+        return;
+    }
+    
     [self downloadUpdate];
 }
 
--- a/Sparkle/SUUIBasedUpdateDriver.m
+++ b/Sparkle/SUUIBasedUpdateDriver.m
@@ -114,6 +114,10 @@
     [self.host setObject:nil forUserDefaultsKey:SUSkippedVersionKey];
     switch (choice) {
         case SUInstallUpdateChoice:
+            if (![self verifyHostVolumeWritable]) {
+                return;
+            }
+            
             self.statusController = [[SUStatusController alloc] initWithHost:self.host];
             [self.statusController beginActionWithTitle:SULocalizedString(@"Downloading update...", @"Take care not to overflow the status window.") maxProgressValue:0.0 statusText:nil];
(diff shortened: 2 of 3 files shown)


zorgiepoo:
Thanks for the PR.

There may be a user expectation however that the user shouldn't be prompted for an update if it's not possible to update the application (the "Install Update" button is a lie). Likewise, the same can be said about establishing a connection to the update server. Probed updates could be an exception to that though.

I do see the appeal of not moving an application until necessary. But at the same time I'm not sure if we should develop the framework for those non-standard needs. Apple does not want apps to remain on a read-only volume, although they don't currently offer a great way around this.

jhammer:
I agree with you, but this PR doesn't change the user experience in a meaningful way from the current way of doing things—it just provides the delegate (and user) with more information:

1. For updates that are initiated by the user, they currently see the "can't update; read-only" alert right up front when they try and check for updates. This PR would change that so they see it only after an update is offered. Either way, the user is still seeing the same "can't update; read-only" alert in response to their request to update. I would argue this is better experience if only because it gives the user more information to base a decision on.

2. For updates that are not initiated by the user (like a background update), the user is not bothered by any alerts (same as before). However, the delegate at least is made aware of an update that is available and that it cannot be applied because the volume is not writable. What the app developer wants to do with that information is up to them. In most cases, the answer is probably "nothing".

Regarding your last paragraph: I may have been unclear in my PR, but I am not proposing that Sparkle do anything related to the actual moving of the application. The scope of this PR is just tweaking when that error is propagated so that the delegate has more information with which to make decisions.

It might help if I explain a possible use case for this PR:

1. User downloads app (as a .zip archive) from website on macOS 10.12
2. Unzips and launches. App is translocated.
3. Sparkle update check is made
4. Delegate learns that an update is available but that it could not be downloaded since disk is not writable
5. App (not Sparkle framework) uses Security APIs to determine that app has been translocated
6. App (not Sparkle framework) informs the user that an update is available and prompts them to move the app to /Applications (or does it for them)

Thanks for considering!

zorgiepoo:
> For updates that are initiated by the user, they currently see the "can't update; read-only" alert right up front when they try and check for updates. This PR would change that so they see it only after an update is offered. Either way, the user is still seeing the same "can't update; read-only" alert in response to their request to update. I would argue this is better experience if only because it gives the user more information to base a decision on.

A "can't update; read-only" is shown immediately because the application doesn't have any hope of updating, which is why a connection to the server isn't established, as well as why the user isn't supplied with a falsified install option. Showing the release notes in detail is debatably better.

> For updates that are not initiated by the user (like a background update), the user is not bothered by any alerts (same as before).

I don't think you accounted for non-user initiated UI scheduled updates.

> However, the delegate at least is made aware of an update that is available and that it cannot be applied because the volume is not writable. What the app developer wants to do with that information is up to them. In most cases, the answer is probably "nothing".

But a user may expect the application to not check for updates (no internet connectivity) if the app has no hope of updating.. which is something that needs to be well considered if changed.

jhammer:
Sounds like a more acceptable PR for now would be to just restrict this behavior to `SUProbingUpdateDriver` so that `-[SUUpdater checkForUpdateInformation]` is functional even on read-only volumes, but the rest of the update machinery is left unchanged. What do you think?

zorgiepoo:
That sounds reasonable to me, wonder what @pornel thinks. Really, I'm not a fan of this translocation stuff :(.

Sparkle/SUUIBasedUpdateDriver.m
```diff
@@ -114,6 +114,10 @@ - (void)updateAlertFinishedWithChoice:(SUUpdateAlertChoice)choice
     [self.host setObject:nil forUserDefaultsKey:SUSkippedVersionKey];
     switch (choice) {
         case SUInstallUpdateChoice:
+            if (![self verifyHostVolumeWritable]) {
+                return;
```
pornel: Do I read this correctly that if user presses the Install button, nothing will happen at all?

Sparkle/SUBasicUpdateDriver.m
```diff
@@ -43,14 +43,20 @@ @implementation SUBasicUpdateDriver
 @synthesize tempDir;
 @synthesize relaunchPath;
 
+- (BOOL)verifyHostVolumeWritable
```
pornel: Probably a better name for it would be `abortIfOnReadOnlyVolume`

pornel:
I think it's an improvement when a *manual* check can still tell the user that there are no new updates. This way the user can know whether it's worth moving the app.

However, I'm concerned that showing errors from regular automatic updates would annoy users. Maaaybe only show the error if the update found is marked as critical.

And finally, showing the regular update prompt with an Install button, and then showing an error, is a bad UI. If we know the update will fail, let's not ask user an invalid question. The prompt could be replaced with a message along lines "There is an update available, but to install it you first have to move the app to the Applicaitons folder and restart from there".

zorgiepoo:
On second thought with having just probe checks working - a developer that performs a probe check may expect a scheduled UI update to work after.

This PR seems to be defer wanting to move the application away until a new update is out, which will inevitably happen sooner or later. However it may be easiest to handle this independently from Sparkle, or if Apple later provides a way to opt out of translocation. For example, if the user has run your application on a read-only volume for a "long" period of time, you can alert the user (or do it for them) to move the app elsewhere.

pornel:
I'm going to close this.  

As far as I understand this implementation, it would cause error messages to pop up when user is using an app from a read-only volume and an update is available. I think the current behavior of Sparkle staying silent and out of the way until user checks for update is a better experience.

Issue #861, opened 2016-08-10
Does Sparkle use <channel><link> for anything?


moubry:
I'm sorry in advance — this is pretty pedantic.

In the example [SampleAppcast.xml](https://github.com/sparkle-project/Sparkle/blob/master/Resources/SampleAppcast.xml) the value for channel’s `<link>` is this:

``` xml
<?xml version="1.0" encoding="utf-8"?>
<rss version="2.0" xmlns:sparkle="http://www.andymatuschak.org/xml-namespaces/sparkle" xmlns:dc="http://purl.org/dc/elements/1.1/">
    <channel>
        <title>Your Great App's Changelog</title>
        <link>http://you.com/app/appcast.xml</link>
        ...
```

However, [if you look at the RSS 2.0 spec](https://validator.w3.org/feed/docs/rss2.html#requiredChannelElements) for this value:

> Element: link
> Description: The URL to the HTML website corresponding to the channel.
> Example: http://www.goupstate.com/

Should I do what the example appcast does (link to the XML file) or should I do what the RSS spec says to do (link to the app’s HTML page) or does it even matter? Does Sparkle use the channel link value for anything?

pornel:
Sparkle doesn't use that link. It's an error in the example.

PR #862, opened 2016-08-10
Remove <link> from SampleAppCast.xml

moubry:
Sparkle doesn’t use the channel’s link element. This resolves issue https://github.com/sparkle-project/Sparkle/issues/861.


pornel:
Thanks

Issue #864, opened 2016-08-15
Bug: Calling installUpdatesIfAvailable will disable checkForUpdates


denniss:
Hi guys,

I am building a background app and in `applicationDidFinishLaunching`, I am doing the following: 

```
let updater = SUUpdater.sharedUpdater()
if updater.automaticallyChecksForUpdates && updater.automaticallyDownloadsUpdates {
    updater.installUpdatesIfAvailable()
}
```

However, I think doing this will put the updater in a bad state. When it does not find any update to download or install, calling `checkForUpdates:_` will just just return because this line in `checkForUpdatesWithDriver` will always return true:

`if ([self updateInProgress]) { return; }`

denniss:
My current hacky solution to this:

```
class UpdaterController : NSObject, SUUpdaterDelegate {
    
    static let sharedInstance = UpdaterController()
    
    func updaterDidNotFindUpdate(updater: SUUpdater!) {
        if let driver = updater.valueForKey("driver") {
            driver.performSelector(Selector("abortUpdate"))
        }
    }
}
```

It looks like the behavior I mentioned above happens because of this method

```
- (void)didNotFindUpdate
{
    if ([[self.updater delegate] respondsToSelector:@selector(updaterDidNotFindUpdate:)])
        [[self.updater delegate] updaterDidNotFindUpdate:self.updater];
    [[NSNotificationCenter defaultCenter] postNotificationName:SUUpdaterDidNotFindUpdateNotification object:self.updater];

    if (!self.automaticallyInstallUpdates) {
        NSAlert *alert = [[NSAlert alloc] init];
        alert.messageText = SULocalizedString(@"You're up-to-date!", "Status message shown when the user checks for updates but is already current or the feed doesn't contain any updates.");
        alert.informativeText = [NSString stringWithFormat:SULocalizedString(@"%@ %@ is currently the newest version available.", nil), [self.host name], [self.host displayVersion]];
        [alert addButtonWithTitle:SULocalizedString(@"OK", nil)];
        [self showAlert:alert];
        [self abortUpdate];
    }
}
```

I am setting `automaticallyInstallUpdates` to true and because of that, `[self abortUpdate];` does not get called.

zorgiepoo:
Looks like a bug that `abortUpdate` is not called. Good catch!

[...]

PR #865, opened 2016-08-15
[BUG-864] Call abort update after updater cannot find valid update

denniss:
For fixing this:
https://github.com/sparkle-project/Sparkle/issues/864


zorgiepoo:
This should be good to merge for now, even though I don't like supporting `-installUpdatesIfAvailable`

pornel:
Thanks!

PR #872, opened 2016-08-27
Menu item validation

zorgiepoo:
This just alters the menu item validation to match what Andy must have originally intended for.

Which is that if an update is downloaded automatically in the background, the user can initiate an update check that will abort that downloaded update, and check for the latest update from the server. The upside is that the newest update will be retrieved, but the downside is that the downloaded update will be wasted in this scenario (my [fork](https://github.com/zorgiepoo/Sparkle) makes the opposite decision).

Anyway, this is just a simple fix. I also made the menu item validation work on the test app's window controller.


zorgiepoo:
Oh, and note I did not change the implementation of -updateInProgress because interruptible updates are still updates in progress.

PR #873, opened 2016-08-27
Dsa existence

zorgiepoo:
Two changes:

1. Alert developers from making a mistake of specifying a filename to a DSA public key in their bundle's Info.plist but not having the file actually exist in the bundle's resources.
2. Prevent upgrades that would result in a security downgrade by removing the public DSA key (this is a security related change).

I have seen developers make a mistake of not supplying a public key in their new update.

### DIFF
--- a/Sparkle/SUUpdater.m
+++ b/Sparkle/SUUpdater.m
@@ -127,16 +127,23 @@ static NSString *const SUUpdaterDefaultsObservationContext = @"SUUpdaterDefaults
     BOOL servingOverHttps = [[[feedURL scheme] lowercaseString] isEqualToString:@"https"];
 
     if (!hasPublicDSAKey) {
-        if (!isMainBundle) {
+        // If we failed to retrieve a DSA key but the bundle specifies a path to one, we should consider this a configuration failure
+        NSString *publicDSAKeyFileKey = [self.host publicDSAKeyFileKey];
+        if (publicDSAKeyFileKey != nil) {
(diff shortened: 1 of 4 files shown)


Sparkle/SUUpdater.m
```diff
@@ -127,16 +127,23 @@ -(void)checkIfConfiguredProperly {
     BOOL servingOverHttps = [[[feedURL scheme] lowercaseString] isEqualToString:@"https"];
 
     if (!hasPublicDSAKey) {
-        if (!isMainBundle) {
+        // If we failed to retrieve a DSA key but the bundle specifies a path to one, we should consider this a configuration failure
+        NSString *publicDSAKeyFileKey = [self.host publicDSAKeyFileKey];
+        if (publicDSAKeyFileKey != nil) {
             [self showAlertText:SULocalizedString(@"Insecure update error!", nil)
-                informativeText:SULocalizedString(@"For security reasons, you need to sign your updates with a DSA key. See Sparkle's documentation for more information.", nil)];
+                informativeText:[NSString stringWithFormat:SULocalizedString(@"For security reasons, the file indicated by the '%@' key needs to exist in the bundle's Resources.", nil), SUPublicDSAKeyFileKey]];
```
pornel: It'd be helpful to print `publicDSAKeyFileKey` as well

zorgiepoo: Done.

PR #874, opened 2016-08-27
Remove NTSynchronousTask

zorgiepoo:
Changes taken from my fork, this removes a dependency on NTSynchronousTask which I I have found to be unnecessary. DMG unarchiving code and code signing verification tests used this dependency, but uses of it can be replaced with NSTask.


Sparkle/SUDiskImageUnarchiver.m
```diff
@@ -89,12 +88,26 @@ - (void)extractDMGWithPassword:(NSString *)__unused password
...
+            task.standardOutput = outputPipe;
+            
+            [task launch];
+            
+            [inputPipe.fileHandleForWriting writeData:promptData];
+            [inputPipe.fileHandleForWriting closeFile];
+            
+            [task waitUntilExit];
+            
+            output = [outputPipe.fileHandleForReading readDataToEndOfFile];
```
pornel: I've found that such pattern can deadlock. If the task tries to write to stdout, but stdout buffer is full, it will wait forever, because your task is not reading yet.

zorgiepoo: Good catch! I re-ordered it such that the standard output is read before waiting until the task exists. Does this address that issue?

pornel: Yup. That should be good.

jkbullard:
A single read before testing for the end of the task is very fragile because it is subject to race conditions. If, for example, the CPU is heavily loaded and memory is low, the task may not output anything until after the read is done, thus the read will not read anything and the task will stall if it later fills up its output buffer. Or if the task outputs immediately, fills its buffer and stalls, and then the read is done, the task will stall the next time its buffer fills.

One solution is to loop, monitoring the output and emptying it as needed, until the task is done. At one time this was the approach used by [EtreCheck](https://github.com/etresoft/EtreCheck); I don't know if that is still true.

But I have seen that approach fail, too (long story; possible OS X bug), so [Tunnelblick](https://tunnelblick.net)  ([source code on GitHub](https://github.com/Tunnelblick/Tunnelblick)) sends the task's output to a temporary file and then reads from the file after waiting for the task to exit.

pornel:
@jkbullard My understanding is that `readDataToEndOfFile` is *not* a single `read()`, but a loop internally that blocks and keeps reading until EOF. If it's called before the process starts, it'll just block and wait.

The only potential I see here for deadlock is if the process never closed stdout. However, because we're setting custom pipe and not inheriting handles, I don't think it could outlive the process.

zorgiepoo:
Hm, not sure about all of this, but looking at this [thread](https://groups.google.com/forum/#!topic/des-moines-cocoaheads/IygQulT48VA) I think using `-readDataToEndOfFile` may be okay here. The data returned back shouldn't be *that* big too, and `hdiutil` should be closing stdout, not spawning another process that outlives it or anything, if that's relevant.

PR #876, opened 2016-08-27
Update mdimporter modification dates after delta extractions

zorgiepoo:
Developers are supposed to [update the modification date](https://developer.apple.com/library/mac/documentation/Carbon/Conceptual/MDImporters/Concepts/Troubleshooting.html) of their mdimporter bundles regardless of whether or not the update is distributed over Sparkle or on their website. The one edge case is if we do a delta extraction where we may update a file inside the mdimporter bundle but not update the modification date of the spotlight importer.

So to be on the safe side we update the modification dates after delta extractions, and remove our invocation to `mdimport` during the final step of installation.


pornel:
Previous implementation called mdimporter with no args (rather than `-r` as the doc you've linked to). I think that was slightly different behavior - it updated data about the app as a search result in Spotlight, which is different than addition of app's bundled importers.

zorgiepoo:
Didn't realize that article referenced `-r`, however that section is if the importer is *not* part of an application bundle. Other references in the article show running `mdimport` with no arguments (besides some debug logging ones).

If you download an application or update online, nothing in the system is going to automatically run `mdimport`. All we have to do is register with launch services (which we do when we touch the app bundle), have the application run once, and update the mod date for mdimporter bundles after delta extractions.

Part of the reason I removed mdimport in my fork is beaches it's not runnable under root.

pornel:
Makes sense.

Issue #863, opened 2016-08-14
Sparkle uses host app identity to verify bundle updates


sergeche:
I’m building a TextMate 2 custom plugin (e.g. bundle) and want to add plugin autoupdate feature.

Here’s init code for updater from my bundle:

```swift
let bundle = NSBundle(forClass: self.dynamicType)
updater = SUUpdater(forBundle: bundle)
updater.checkForUpdates(self)
```

Update check works fine: it displays my bundles’ icon, name and version; downloads and extracts update. The problem is in “Install & Relaunch” phase: instead of checking codesign identity of my bundle, it uses identity of host app (e.g. TextMate 2) which, of course, differs from mine and there’s no way to install bundle update.

Am I missing something? How can I tell Sparkle to use bundle’s identity?

zorgiepoo:
I think you can get by if you use DSA signatures described on the [documentation page](https://sparkle-project.org/documentation/), although it does point out a flaw in the verification process.

sergeche:
I did, but it doesn’t help: Sparkle still uses host app’s codesign identity to verify my bundle

zorgiepoo:
If DSA keys are available and match, it should then [not match](https://github.com/sparkle-project/Sparkle/blob/master/Sparkle/SUBasicUpdateDriver.m#L335) the Apple code signatures.

sergeche:
Yes, it worked, but only after I disabled bundle codesigning. If bundle is codesigned it still tries to compare identities.

Thanks!

zorgiepoo:
@sergeche we just merged in changes that should resolve this by using your bundle instead of the host app.

PR #877, opened 2016-08-27
Improve Host Signing Verification

zorgiepoo:
The code signature verification methods no longer assume the host is the application invoking the methods. This is mostly taken from my fork + I just added some unit tests. This should help resolve issues like #863

### DIFF
--- a/Sparkle/SUCodeSigningVerifier.m
+++ b/Sparkle/SUCodeSigningVerifier.m
@@ -59,6 +59,8 @@
     
     // Note that kSecCSCheckNestedCode may not work with pre-Mavericks code signing.
     // See https://github.com/sparkle-project/Sparkle/issues/376#issuecomment-48824267 and https://developer.apple.com/library/mac/technotes/tn2206
+    // Aditionally, there are several reasons to stay away from deep verification and to prefer DSA signing the download archive instead.
+    // See https://github.com/sparkle-project/Sparkle/pull/523#commitcomment-17549302 and https://github.com/sparkle-project/Sparkle/issues/543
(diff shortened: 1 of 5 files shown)


pornel:
Looks good, but the change is scary, because a bug in it could cause confused deputy problem.

zorgiepoo:
Feel free to hold off it. I'm just porting some of my changes back here. I'm always iffy on these security type APIs. I wrote it based on how the other functions were written. The confidence I can give is just that I've tested the changes and stubbed in the framework into a random 3rd party app and had it update.

I'm not realllyyy familiar with the confused deputy problem, but I wonder if you're talking about a bug in my implementation, or an attacker trying to swap a bundle.

Sparkle/SUCodeSigningVerifier.m
```diff
@@ -13,53 +13,62 @@
...
+        SULog(@"Failed to load NSBundle for original");
+        result = -1;
         goto finally;
     }
-
-    result = SecCodeCopyDesignatedRequirement(hostCode, kSecCSDefaultFlags, &requirement);
+    
+    result = SecStaticCodeCreateWithPath((__bridge CFURLRef)[oldBundle bundleURL], kSecCSDefaultFlags, &oldCode);
+    if (result == errSecCSUnsigned) {
+        return NO;
```
pornel: Other branches set `result` and `goto finally`, so it'd be good to do it here for consistency

pornel:
BTW, bundle creation here could be avoided if `codeSignatureAtPath:matchesSignatureAtPath:` worked on bundle URLs instead of paths. The temp bundles are loaded only to get their URLs, but we could ask for a URL earlier.

I'll make a PR with this.

PR #878, opened 2016-08-28
Improve host writability check by testing ownership/group changes

zorgiepoo:
So an update isn't attempted to be installed automatically/silently when authentication is necessary. Related: #795

### DIFF
--- a/Sparkle/SUHost.m
+++ b/Sparkle/SUHost.m
@@ -11,6 +11,7 @@
 #import "SUSystemProfiler.h"
 #include <sys/mount.h> // For statfs for isRunningOnReadOnlyVolume
 #import "SULog.h"
+#import "SUFileManager.h"
 
 #if __MAC_OS_X_VERSION_MAX_ALLOWED < 101000
 @interface NSProcessInfo ()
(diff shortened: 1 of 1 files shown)


Sparkle/SUHost.m
```diff
@@ -74,7 +75,34 @@ - (BOOL)allowsAutomaticUpdates
...
+    if (![[NSFileManager defaultManager] isWritableFileAtPath:bundlePath.stringByDeletingLastPathComponent] || ![[NSFileManager defaultManager] isWritableFileAtPath:bundlePath]) {
+        return NO;
+    }
+    
+    // Just because we have writability access does not mean we can set the correct owner/group silently
+    // Test if we can set the owner/group on a temporarily created file
+    // If we can, then we can probably perform an update without authorization
+    // One place where this matters is if you copy and run an app from /tmp/
+    
+    NSString *tempFilename = @"permission_test" ;
```
pornel: I suggest using a dot-file to avoid Finder (over)reacting to it. Perhaps `.sparkle_update_permission_test`?

zorgiepoo: Is using a dot file needed? I'm creating it in a requested temporary directory (not related to me trying to run the app in `/tmp/`)

Changing owner/group is the first likely thing to happen that might require authentication. I suppose it doesn't have to require authentication, and it may be possible for another operation like writing into the parent directory to require authentication though. Or vise versa like shown here.

PR #881, opened 2016-09-02
Separate validation

zorgiepoo:
This pull request requires binary delta patches to be validated before attempting extraction, rather than validating them after doing extraction. Delta patches are our own archive format, and I'd feel safer if they were not extracted first. This is also an efficiency saver if the patch is not validated properly. If a delta patch fails, a full update can always be attempted, so failing if a developer tries to change DSA keys is OK.

This also refactors the validation for bundles and packages so the check isn't in the same function.

A UI display bug is also fixed if a download is attempted more than once (i.e, delta patch fails).

Performed some test cases with package updates, app updates, and delta updates where the signature is valid and invalid.


zorgiepoo:
Merged latest changes from master

pornel:
The code looks ok, but the mutable state of `validatedDsaSignatureBeforeUnarchiving` worries me a bit. I don't see any bug with it now, but I'm worried that such indirect way of reporting security-sensitive flag could cause problem in the future.

zorgiepoo:
I made the rename change (unsafe does make more sense), and refactored the validation into its own class to address the mutability concerns. I used "prevalidate" as opposed to "validate" (or post validate) to distinguish between validation before and after unarchiving.

pornel:
That's very good

PR #882, opened 2016-09-02
Use bundle URLs

pornel:
Use bundle URLs instead of paths to avoid creation of temp bundle objects

### DIFF
--- a/Sparkle/SUBasicUpdateDriver.m
+++ b/Sparkle/SUBasicUpdateDriver.m
@@ -337,18 +337,17 @@
             return NO;
         }
     }
-    
-    BOOL updateIsCodeSigned = [SUCodeSigningVerifier bundleAtPathIsCodeSigned:installSourcePath];
+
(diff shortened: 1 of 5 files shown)


pornel:
I hope this change doesn't touch Swift tests, because they need syntax changes in the latest xcode :)

zorgiepoo:
Good idea. `SUCodeSigningVerifierTest.m` has a lot of mismatched paths/url code (eg: naming a variable a URL and assigning it a NSString, or invoking URL method on a NSString). I'm guessing it's probably why the tests are failing. Other than that I think it looks good. Don't think swift is involved here.

zorgiepoo:
You can use -[NSURL checkResourceIsReachableAndReturnError:] instead of -[NSFileManager fileExistsAtPath:] to remove the nullability warnings in SUCodeSigningVerifier.m

pornel:
Thanks. I didn't know it existed.

PR #884, opened 2016-09-04
Xcode8

pornel:


zorgiepoo:
Xcode 8 is now out, so this and #889 should be merged.

PR #885, opened 2016-09-05
Add more documentation to SUUpdater.h

zorgiepoo:
Add docs and re-order some methods based on order of most used/importance.

[edit]: I realize that some of the docs are also duplicated on the website.. That is kind of unfortunate.


pornel:
👍

Issue #759, opened 2016-02-18
SUScheduledCheckInterval set to 1 minute in latest version?


taoeffect:
We updated our app's Sparkle and a user told us that it's now checking for updates every minute now. We confirmed this. Our app doesn't set `SUScheduledCheckInterval` in the Info.plist and just goes with what should be the default of 86400.

Is this possibly a bug in the latest version?

Manually setting the prefs using `defaults` fixes the problem.

zorgiepoo:
https://github.com/sparkle-project/Sparkle/blob/master/Sparkle/SUConstants.m#L21 - looks like for debug mode, it's set at 1 minute. Did you build a debug version of Sparkle yourself?

This does make me think that having this kind of code is not a good idea.

taoeffect:
Looks like that's probably the culprit, thanks @zorgiepoo!

PR #886, opened 2016-09-06
Warn a developer if they're using a debug version of Sparkle

zorgiepoo:
Due to different values for min/default update check intervals, production apps should not use a debug version of Sparkle

---

Realized I had this branch I created long ago, and think it'd be a good idea to help issues like #759

### DIFF
--- a/Sparkle/SUUpdater.m
+++ b/Sparkle/SUUpdater.m
@@ -59,6 +59,15 @@ NSString *const SUUpdaterAppcastNotificationKey = @"SUUpdaterAppCastNotification
 static NSMutableDictionary *sharedUpdaters = nil;
 static NSString *const SUUpdaterDefaultsObservationContext = @"SUUpdaterDefaultsObservationContext";
 
+#ifdef DEBUG
++ (void)load
+{
+    // Debug builds have different configurations for update check intervals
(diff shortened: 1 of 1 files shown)


vslavik:
FWIW, this has some less pleasant consequences to those of us who don’t use CocoaPods or prebuilt Sparkle binaries, but build everything, Sparkle included, as part of a workspace:

Xcode configurations are global, so Sparkle is inevitably build in Debug when debugging the app (but also guaranteed to be built as Release when archiving, so the nanny warning is not helpful in this case). If you have Xcode configured to show the output view only if there actually is output, it’s now always going to show up and eat valuable screen space.

It’s easy enough to revert in a private fork, of course, but still something to be aware of.

zorgiepoo:
Yeah, I'm not actually a fan of the debug build having such runtime consequences. If it didn't, we wouldn't have to log out anything. They exist for the test application primarily, so I suppose we could add an option for changing the minimum update check interval (`SUDefaultUpdateCheckInterval`already doesn't need to depend on `DEBUG` because we have a update check interval option).

pornel:
Maybe use something like this?

     #if defined(DEBUG) && !defined(SPARKLE_DONT_WARN)

vslavik:
I think @zorgiepoo’s approach, i.e. removing the need for differing builds in the first place, makes more sense.

PR #889, opened 2016-09-15
Swift 3

zorgiepoo:
Updates unit tests to swift 3, and updates travis to use xcode8 OS X image.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: objective-c
-osx_image: xcode7
+osx_image: xcode8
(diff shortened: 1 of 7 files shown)


zorgiepoo:
I had to also remove the defunct coveralls, which means we can now have passing pull requests! :)

pornel:
Wow, Travis has new Xcode available already. That was quick!

zorgiepoo:
They should have to if they want people to use them :p. Less so for updating their OS X environment though.

PR #906, opened 2016-10-30
Fix flickering download progress status text

vslavik:
This PR fixes status text during download (e.g. "10MB of 22MB") so that it no longer distractingly jumps around and flickers. This was caused by two things:
1. Flicker on macOS >= 10.8 due to the use of NSByteCountFormatter's   default mode which isn't optimized for stable display and is not   suitable for rapidly updated values.
2. Smaller movements of text on macOS >= 10.11 due to the use of San   Francisco font's default variant, where digits don't have fixed   width.

The fix is to use appropriate APIs, i.e. use monospaced digits variant of the system font and make `NSByteCountFormatter` output more stable values (pre-10.8 manual formatting code was OK already and behaved like `setZeroPadsFractionDigits:YES`).

Let me show you:

_Before:_

sparkle_before
(Rapid jumping around when the text changes from “10.9” to “11” and then “11.1” rapidly, significantly shortening the label for a fraction of a second, wasn’t captured due to its short duration, but it’s actually the _more_ noticeable problem for a human eye.)

_After:_

sparkle_after

### DIFF
--- a/Sparkle/SUStatusController.m
+++ b/Sparkle/SUStatusController.m
@@ -11,6 +11,7 @@
 #import "SUAppcast.h"
 #import "SUAppcastItem.h"
 #import "SUVersionComparisonProtocol.h"
+#import "SUOperatingSystem.h"
 #import "SUStatusController.h"
 #import "SUHost.h"
 
@@ -28,6 +29,7 @@
 @synthesize host;
 @synthesize actionButton;
 @synthesize progressBar;
+@synthesize statusTextField;
 
 - (instancetype)initWithHost:(SUHost *)aHost
 {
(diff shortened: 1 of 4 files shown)


Sparkle/SUStatusController.m
```diff
@@ -51,6 +52,11 @@ - (void)windowDidLoad
     [[self window] center];
     [[self window] setFrameAutosaveName:@"SUStatusFrame"];
     [self.progressBar setUsesThreadedAnimation:YES];
+
+    if ([NSFont respondsToSelector:@selector(monospacedDigitSystemFontOfSize:weight:)])
```
zorgiepoo: Use `+[SUOperatingSystem isOperatingSystemAtLeastVersion:]` instead and suppress the `-Wpartial-availability` warnings (like in `-[SUUIBasedUpdateDriver localizedStringFromByteCount:]`). Apple doesn't in general recommend doing runtime checking anymore (with some exceptions..) because private and incomplete implementations can exist in prior OS versions.

zorgiepoo:
Looks great! 👍

Sparkle/SUStatusController.m
jakepetroules: Yes, please fix and we can merge.

vslavik:
Updated, thanks for the review!

PR #908, opened 2016-11-01
Use optimized icon for Autoupdate.app

### LINKED ISSUE #803: Compress icon

pornel:
20 times smaller Sparkle.icns file + a Makefile that updates it from PNG files (it's not run during normal builds, because it needs to download and build libicns, which would be an overkill)

Closes #803


pornel:
Is that OK? @zorgiepoo @jakepetroules?

I think that's the last thing stopping the next release.

jakepetroules:
I hate to hardcode "binaries". Maybe you could make libicns a submodule and then only run the optimized icon build process in release mode?

jkbullard:
If I understand this correctly, pornel's PR adds an optimized icon binary that he says was generated from AppIcon.appiconset, and includes the script he says was used to generate the optimized icon binary from AppIcon.appiconset using an external tool (libicns). The binary will not be generated during a "normal" build.

Question: when _is_ it built?

Only manually, I assume, so if/when AppIcon.appiconset changes the PR that includes the change should also include a new Sparkle.icns that is an optimized version.

That seems like a reasonable tradeoff to me; I'd rather not have my Sparkle builds run code that is outside of Sparkle (that is, libicns). Or am I missing something and that is already done via other submodules?

(And kudos and thanks to pornel for including the Makefile-icon to show how to use libicns to do this!)

Anyone who "trusts" pornel's binary can just use it as is, and anyone who doesn't can use the script themselves to build the icon binary. (Just as anyone who "trusts" Sparkle can use the framework binary instead of building it themselves.) Frankly, I'd rather trust the binary than run libicns; for me it is not the time involved in running it, it is that it is an additional dependency that I would want to examine before using. Providing a binary makes that unnecessary. If it weren't basically images, and was a binary that executed code, I might have a different view, of course.

zorgiepoo:
I kind of agree integrating it in the build process may be another point of failure so I don't mind the current approach. On another note, I'd like to envision in the future that we'd want to remove including the icon in the .framework.

Btw, I currently get an error trying to execute that Makefile, probably because I need to install some other dependency to compile it:

```
msp@Slim ~/D/P/Sparkle> make -f Makefile-icon 
git clone https://github.com/pornel/libicns.git
Cloning into 'libicns'...
remote: Counting objects: 1620, done.
remote: Compressing objects: 100% (4/4), done.
remote: Total 1620 (delta 0), reused 0 (delta 0), pack-reused 1616
Receiving objects: 100% (1620/1620), 762.52 KiB | 0 bytes/s, done.
Resolving deltas: 100% (1112/1112), done.
Checking connectivity... done.
( cd libicns && autoreconf -vif && ./configure )
autoreconf: Entering directory `.'
autoreconf: configure.ac: not using Gettext
autoreconf: running: aclocal --force -I m4
aclocal: warning: couldn't open directory 'm4': No such file or directory
autoreconf: configure.ac: tracing
autoreconf: configure.ac: not using Libtool
autoreconf: running: /usr/local/Cellar/autoconf/2.69/bin/autoconf --force
configure.ac:27: error: possibly undefined macro: AC_PROG_LIBTOOL
      If this token and others are legitimate, please use m4_pattern_allow.
      See the Autoconf documentation.
autoreconf: /usr/local/Cellar/autoconf/2.69/bin/autoconf failed with exit status: 1
make: *** [libicns/Makefile] Error 1
```

pornel:
I don't think there's a problem of trust here, since Sparkle.icns is not code, and it's easy to verify it contains images it should.

I acknowledge this is committing *derived* content instead of just the source files. I usually avoid that too, but in this case the trade-offs are:

- do nothing, and have 780KB icon file. That's unacceptable for me personally, since that icon alone is 50% size of my own app's entire archive. I can deal with amount of wasted bandwidth it generates, but not with the absurdity of the bloated file size :)

- Add libicns, pngquant to the build process, along with their dependencies of autoconf, automake, pkg-config and libpng. That avoids having redundant Sparkle.icns in git, but makes building of Sparkle itself much more complicated and fragile. For unlucky users who don't have the right version of autotools or libpng include paths this can be a hair-pulling nightmare to work with.

- or use this PR, which adds a 37KB file to git, which we change about once a year maybe.

pornel:
@zorgiepoo `brew install libtool`

jkbullard:
I apologize if anyone took my comments to mean I don't/didn't trust pornel, that was not my intention.

Forcing everyone who builds Sparkle to install those dependencies seems like a lot to ask and makes building much more problematic and fragile. (As far as I know, Sparkle doesn't require anything beyond Xcode, but maybe I'm just lucky and already have installed anything else it needs to build.)

Including the binary means that anyone who doesn't want to build the binary doesn't have to, and including the script makes it easy to build the binary for those who are paranoid (even about images) and are willing to do the work.

zorgiepoo:
Yeah, we're all on the same page, except for maybe Jake.

jakepetroules:
It was just a suggestion. Doesn't bother me *that* much. Go ahead. 👍

pornel:
@jkbullard No offense taken. I get that "trust" here is a technical term.

zorgiepoo:
@pornel Does the test app use the optimized icon, or should we care?

pornel:
The test app uses the fat iconset, because we don't have to care about the size there.

PR #911, opened 2016-11-05
Master

erikbaarstad:
Use optimized icon for Autoupdate.app


pornel:
This looks like a mistake. We shouldn't be merging master into an old release branch.

PR #914, opened 2016-11-09
Rename Sparkle.icns to be consistent with ASSETCATALOG_COMPILER_APPICON_NAME

pornel:
That was the previous name of the file


zorgiepoo:
Sure

Issue #409, opened 2014-08-18
Don't SKIP_INSTALL (Revert d888705fd8376985e0b090d4a2582d7c838988f2)


jeremyhu:
I'm updating from andymatuschak's final commit (d2b7fd6236234ca65f273cb65e309920a52dd092) to 1.6.1.  In doing so, I noticed that SKIP_INSTALL was set to YES, so my install obviously doesn't install anything.

Please revert this.

commit d888705fd8376985e0b090d4a2582d7c838988f2
Author: Jorge Leandro Perez jorge@lantean.co
Date:   Tue Jan 28 01:15:14 2014 -0300

```
Enables Skip Install flag
```

diff --git a/Sparkle.xcodeproj/project.pbxproj b/Sparkle.xcodeproj/project.pbxproj
index 3aa696a..cb02b9e 100644
--- a/Sparkle.xcodeproj/project.pbxproj
+++ b/Sparkle.xcodeproj/project.pbxproj
@@ -1368,6 +1368,7 @@
                                COMBINE_HIDPI_IMAGES = YES;
                                OTHER_CFLAGS = "-DDEBUG";
                                SDKROOT = macosx;
-                               SKIP_INSTALL = YES;
                      };
                      name = Debug;
              };
  @@ -1377,6 +1378,7 @@
                      buildSettings = {
                              COMBINE_HIDPI_IMAGES = YES;
                              SDKROOT = macosx;
-                               SKIP_INSTALL = YES;
                      };
                      name = Release;
              };

Issue #315, opened 2014-01-06
Update error!


BaseHead:
Hey all and Andy!

I saw this error many times talked about on the web but nothing seemed to help my situation.

Update Error!
An error occurred while installing the update. Please try again later.

Getting this after it downloads the update and I click to install it.

It all used to work fine but it's been about 4 months since we released a version on auto-updater using Sparkle via Feeder.
seeing it on 10.7 and 10.8 and same with all other users of my software.
Other Sparkle apps on my mac seem fin

Any ideas what to try or check?

thx!

Steve

Issue #194, opened 2012-07-28
Don't recognize new version


Togo:
Hi,

the UpdateWindow from Sparkle says me every time that i have the newest version installed. But i set the version in the xml to 1.1.0 and in the actual test build to 1.0.0. 

```
<?xml version="1.0" encoding="utf-8"?>
<rss version="2.0" xmlns:sparkle="http://www.andymatuschak.org/xml-namespaces/sparkle"  xmlns:dc="http://purl.org/dc/elements/1.1/">
   <channel>
      <title>Your Great App's Changelog</title>
      <link>https://s3.amazonaws.com/hotkeyeve/EVEAppcast.xml</link>
      <description>Most recent changes with links to updates.</description>
      <language>en</language>
         <item>
            <title>Version 2.0 (2 bugs fixed; 3 new features)</title>
                        <sparkle:releaseNotesLink>
                            http://you.com/app/2.0.html
                        </sparkle:releaseNotesLink>
            <pubDate>Wed, 09 Jan 2006 19:20:11 +0000</pubDate>
            <enclosure url="https://github.com/downloads/Togo/eve/EVE V1.1.app.zip" sparkle:version="1.1.0" length="903539" type="application/octet-stream" sparkle:dsaSignature="BAFJW4B6B1K1JyW30nbkBwainOzrN6EQuAh" />
         </item>
      </channel>
</rss>

Is there a way to debug the framework to see whats going on? 

Thank for your help
```

PR #915, opened 2016-11-10
Fix CFBundleShortVersionString format

grahammiln:
Apple's expected version format is three digit version numbers. We were inserting git hashes into this field. This change moves the useful git version information into the previously unused copyright line.

See https://developer.apple.com/library/content/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html#//apple_ref/doc/uid/20001431-102364

> The build version number should be a string comprised of three non-negative, period-separated integers with the first integer being greater than zero—for example, 3.1.2. The string should only contain numeric (0-9) and period (.) characters. Leading zeros are truncated from each integer and will be ignored (that is, 1.02.3 is equivalent to 1.2.3).


zorgiepoo:
So I don't think the short version string is actually problematic or bad. I think this is one of those cases where Apple is being misleading and docs not being trustful :/

That quotation and link in particular is referencing CFBundleVersion though which needs strict requirements because it's parsed in a meaningful way. The short version string on the other hand isn't and is for human readable purposes. A sensible versioning system ought to distinguish between a machine parsesable "build number" and a human readable "version string".

The App Store could have "requirements" for the short version string but that would be purely for visual or conventional purposes since the system doesn't care. Note the short version string is also localizable.

Lastly I think we would be misusing the copyright field were this to change.

grahammiln:
At the moment, Sparkle inserts inappropriately formatted information into the Info.plist. Should we not follow Apple's documentation?

The `git` hash information is useful but not appropriate for these two version key pairs where Apple provide clear formatting guidelines. An extract of the documentation is appended below.

Maybe Sparkle should follow Apple's example and prefix the version information to `CFBundleGetInfoString`; see Preview.app for an example of this. This way the hash will be accessible in the Get Info panel.

I am happy to improve the script to do what is needed. The script could do with additional error checking that could go in at the same time.

> ## CFBundleShortVersionString
>
>CFBundleShortVersionString (String - iOS, macOS) specifies the release version number of the bundle, which identifies a released iteration of the app.
>
>The release version number is a string composed of three period-separated integers. The first integer represents major revision to the app, such as a revision that implements new features or major changes. The second integer denotes a revision that implements less prominent features. The third integer represents a maintenance release revision.
>
>The value for this key differs from the value for CFBundleVersion, which identifies an iteration (released or unreleased) of the app.
>
> This key can be localized by including it in your InfoPlist.strings files.

> ## CFBundleVersion
>
>CFBundleVersion (String - iOS, macOS) specifies the build version number of the bundle, which identifies an iteration (released or unreleased) of the bundle.
>
>The build version number should be a string comprised of three non-negative, period-separated integers with the first integer being greater than zero—for example, 3.1.2. The string should only contain numeric (0-9) and period (.) characters. Leading zeros are truncated from each integer and will be ignored (that is, 1.02.3 is equivalent to 1.2.3). The meaning of each element is as follows:
>
> - The first number represents the most recent major release and is limited to a maximum length of four digits.
> - The second number represents the most recent significant revision and is limited to a maximum length of two digits.
> - The third number represents the most recent minor bug fix and is limited to a maximum length of two digits.
> If the value of the third number is 0, you can omit it and the second period.
>
> While developing a new version of your app, you can include a suffix after the number that is being updated; for example 3.1.3a1. The character in the suffix represents the stage of development for the new version. For example, you can represent development, alpha, beta, and final candidate, by d, a, b, and fc. The final number in the suffix is the build version, which cannot be 0 and cannot exceed 255. When you release the new version of your app, remove the suffix.

zorgiepoo:
Looks like the copyright key replaces 'CFBundleGetInfoString' in which case I think it looks okay to use the copyright key.

Using the git hash for CFBundleVersion is sure wrong. I suppose you're right it may be better in normal cases to stick the git hash in the copyright key than the short version string.

Though I want to clarify that guideline for the short version string is for aesthetics / convention / user experience. That field doesn't need to follow a particular format (here is where I think their docs are misguiding) else the field wouldn't be localizable.

It may also be important to note that Autoupdate is not a normal app. For example, if a troubleshooting user wanted to report the version info from the app, we may be better off making it in a more obvious to get to field rather than worry about the visualness to an ordinary user.

grahammiln:
I have had an application rejection from the Mac App Store because of a malformed version field – it included a git hash just as Sparkle does. Apple does apply these format guides but, of course, Sparkle and the Mac App Store are mutually exclusive.

Sparkle should follow Apple's documentation. The `git` hash can be stored in the copyright line or elsewhere less constrained but visible.

zorgiepoo:
Yeah, as I've mentioned the App Store could add additional restrictions, not to be confused with what the system intended to permit.

Whilst I don't think Sparkle has to follow the convention, I wouldn't be opposed to this change.

ksuther:
Here's one reason to consider not modifying the Info.plist at build time:

https://github.com/sparkle-project/Sparkle/pull/917

Writing it to a header that can then be shown somewhere would speed up incremental builds a bit because changing the Info.plist forces Autoupdate.app to get copied into Sparkle.framework, then Sparkle.framework to get copied into the parent app. Both of these then need to be re-signed every build which takes an extra second or two (instead of just doing nothing).

zorgiepoo:
Well, this pull request has no change on modifying or not modifying the Info.plist at build time, so it's not entirely relevant here.

The norm though is to not build Sparkle every time you try building the app that depends on the framework, so I am curious what kind of workflow you're using, unless you have a reason to build Sparkle frequently? Also, Autoupdate is not really supposed to be signed due to the location we're placing it at.

ksuther:
It seems relevant to me in that the phase `Run Script: Set Git Version Info` modifies the Info.plist every build in order to set the short version string. If I'm misunderstanding then feel free to ignore me :)

We have Sparkle.xcodeproj added as a child project to our main xcodeproj, then reference the framework from there.

You were correct about Autoupdate. What it does do is cause Audoupdate.app to get copied into Sparkle.framework, which then causes Sparkle.framework to get copied into the parent app (and then re-signed). So it's one extra signing step, not two. This gets slightly worse in your `ui-separation-and-xpc` fork, as there are the additional XPC services that all have to be re-signed on every build. Avoiding the extra copying and signing steps becomes a bigger win as there are more and more separate binaries.

zorgiepoo:
> It seems relevant to me in that the phase Run Script: Set Git Version Info modifies the Info.plist every build in order to set the short version string. If I'm misunderstanding then feel free to ignore me :)

Relevant as in we currently have a set git version info script, and we would still have it even if this PR is approved (i.e, no changes to what you're concerned with).

> We have Sparkle.xcodeproj added as a child project to our main xcodeproj, then reference the framework from there.

Ah okay, I didn't know Xcode had such a feature. I just follow the standard setup by copying a built product of Sparkle's framework into my project, which avoids building the framework more than necessary. I'm not that familiar with Cocoapods, but I think it should work similarly considering they recommend developers to store their pods under source control.

By the way, I intended to make inline comments regarding feedback for this pull request, but I'm noticing that this is using GitHub's new review system which I'm not familiar with and that my comments show up as "pending" - not sure if they are actually visible to anyone else.

Configurations/set-git-version-info.sh
```diff
@@ -14,15 +14,19 @@ if [ -z "$SRCROOT" ] || \
...
 
-# and use it to set the CFBundleShortVersionString value
+# and use it to set the NSHumanReadableCopyright value
 export PATH="$PATH:/usr/libexec"
-PlistBuddy -c "Set :CFBundleShortVersionString '$version'" \
+PlistBuddy -c "Set :NSHumanReadableCopyright '$gitversion'" \
+    "$BUILT_PRODUCTS_DIR/$INFOPLIST_PATH"
+
+# CFBundleShortVersionString requires simpler version format
+PlistBuddy -c "Set :CFBundleShortVersionString '$CURRENT_PROJECT_VERSION'" \
```
zorgiepoo: I don't think we have to use PlistBuddy for the short version string. Should be fine with just ${CURRENT_PROJECT_VERSION} in the Info.plist files.

zorgiepoo: I don't think we have to use PlistBuddy for the short version string. Should be fine with just ${CURRENT_PROJECT_VERSION} in the Info.plist files.

zorgiepoo:
>> It seems relevant to me in that the phase Run Script: Set Git Version Info modifies the Info.plist every build in order to set the short version string. If I'm misunderstanding then feel free to ignore me :)

> Relevant as in we currently have a set git version info script, and we would still have it even if this PR is approved (i.e, no changes to what you're concerned with).

Ok, I see. If we don't include the git hash at all in the Info.plist file, we won't have to modify the Info.plist, which would be nice. But we would then be throwing out that information or moving it to some obscure place. I guess you may have been suggesting to have a script write out a header or resource file or some such.

ksuther:
Yeah, that's what I meant. Not having the hash in something that's visible in Get Info doesn't seem like a huge loss since it's a framework.

Alternatively, I thought there was a way to force Xcode to execute build phases in order instead of parallelizing them, but I can't seem to find it now (not Parallelize Builds for building multiple targets).

zorgiepoo:
The git script applies to the framework as well as the Autoupdate app which is later copied to a location outside of the app and could potentially be useful to have the hash available if something goes wrong.. I do agree though having less run scripts is nice.

The following is just my opinion -- I can't help but feel that having your application so coupled with how an external dependency like Sparkle is built is not a great idea. This may or may not be saying that using a subproject here is not good, I don't know. As far as I understand, most people rely on already compiled dependencies when using Xcode (via our website documentation or via Cocoapods) and therefore would not run into this. Feel free to correct me if I'm wrong, but there are a number of other issues (#584, #409, #886) people run into that would be much less of a problem if they weren't concerned with how we build Sparkle. The effect is that a minor internal issue (if one at all) suddenly turns into a significant one, and also sounds quite fragile to me (what happens when a future Xcode version suddenly makes Sparkle not compile - then you wait for a fix or patch it yourself?). That, and how we debug Sparkle may not necessarily align with how people may want to debug their apps.

pornel:
FWIW I use Sparkle as a subproject too, but I've never had any trouble with the version string, so I don't understand what this ticket is about.

zorgiepoo:
> FWIW I use Sparkle as a subproject too, but I've never had any trouble with the version string, so I don't understand what this ticket is about.

It's a matter of formatting correctness - the short version string should be like "x.x.x" according to some division (App Store) from Apple. We use "x.x.x git-hash" but we could move the git-hash to the copyright field. It's a minor problem because it doesn't impact the system differently either way and we're dealing with frameworks & helper apps.

We went off on a related topic of not appending the hash to the Info.plist file via a build run script, which would more negatively affect developers that use Sparkle as a subproject which I'm clearly not being a fan of ;). I admit sub-projects may work slightly better if you're a maintainer/contributor though.

pornel:
But is the formatting requirement just theoretical? What does it break? Sparkle is banned from AppStore on principle anyway.

pornel:
BTW, I've just noticed that the script getting the hash isn't quite correct either - it gets hash of master, rather than hash of checked out branch.

Regardless of formatting, it may be better to get "git hash" equivalent from:

    git describe --tags

which will return a tag on exact match or a closest tag + hash.

zorgiepoo:
> But is the formatting requirement just theoretical? What does it break? Sparkle is banned from AppStore on principle anyway.

Theoretical, doesn't break anything. It's a requirement for the app store and I suspect the main docs on that Info.plist key was influenced from the app store folk. It's more of a matter of following convention.

pornel:
Chrome does not conform either: 53.0.2785.80. Spotify even has a hash: 1.0.41.375.g040056ca. Sublime went freestyle with Build 3126.

I'm not convinced. If the problem is that the hash doesn't conform to strict interpretation of the documentation, then the hash in `NSHumanReadableCopyright` *also* does not strictly conform to the documentation.

zorgiepoo:
Chrome, Spotify, and Sublime all look out of place and show a "worse" user experience by having a strange version string ;). It's merely aesthetics IMO, not that it matters a whole lot for Sparkle anyway.

`NSHumanReadableCopyright` is technically suppose to obsolete/replace `CFBundleGetInfoString` which was described as following:

> CFBundleGetInfoString (String) provides a brief description of the bundle. For an application bundle, this key provides a short description of the application or the current release that the build or release version number cannot convey; for example, the date of the release. This key can be localized.

jkbullard:
Historically, `CFBundleShortVersionString` contained a string which was described as the "marketing version". It is the string that is displayed in "Get Info" for the application, along with the copyright info, and its contents were not constrained.

This was discussed in the first comment by Andy on Issue #194 and the fourth comment by KenThomases on Issue #315.

It's possible that the reason that Apple wants to restrict the contents of `CFBundleShortVersionString` for apps used in the App store is to avoid someone putting malicious Javascript (or similar) into it and causing problems when it is displayed as part of a web page without proper filtering. If it is constrained to digits and periods, that problem is avoided. Sparkle avoids that problem by not displaying it as part of a *web page*, displaying it as part of the "update is available" *window*.

Because:
 * Sparkle cannot be used in apps in the App store;
 * Sparkle is a framework, not an app, so Sparkle's `CFBundleShortVersionString` does not show up except when specifically asking for the version of Sparkle -- that is, "Get Info" on an app that uses Sparkle doesn't show the Sparkle version, it shows the app's version;
 * Putting a git hash into `NSHumanReadableCopyright` is not correct either, so we would just trade one "problem" with another; and
 * The git hash would need to be inserted in all localized versions of `NSHumanReadableCopyright.

I think this PR should be rejected and  the git hash should continue to be in `CFBundleShortVersionString`.

pornel:
I suggest alternative solution. Use `git describe --tags` to obtain the hash. For tagged versions it will *not* include a hash, so the version will match Apple's desired pattern. That should be good enough for people using stable release versions of Sparkle as a subproject.

zorgiepoo:
> Putting a git hash into NSHumanReadableCopyright is not correct either, so we would just trade one "problem" with another; and

That's not really true. The documentation for that key replaces `CFBundleGetInfoString` which could be suited for such a task. But one could argue the short version string could be appropriate enough for our use case.

> The git hash would need to be inserted in all localized versions of `NSHumanReadableCopyright`.

This should be no different than how `CFBundleShortVersionString` can be localizable.

I like @pornel's alternative solution. Although this is overall a pretty nitpicky problem. I think the most important thing pointed out was that the git hash is not retrieved from the current branch.

> ..That should be good enough for people using stable release versions of Sparkle as a subproject.

Will there be any problems if the Info.plist file is only altered *some* of the time, with respect to having to do a clean build, or some such?

pornel:
#919

grahammiln:
I am concerned the desired effect of this change is being lost.

At the moment, Sparkle places an incorrectly formatted version number in the `Info.plist` of the framework and AutoUpdate.app.

Following Apple's documentation will best place Sparkle for use with tools and schemes that expect the version format. I do not believe what other products or projects do is useful, given the clear and precise nature of Apple's guidelines.

Regarding the git hash, this can be considered a separate problem to solve. It should not be in the two CF version values; that is clear.

zorgiepoo's comment about potentially avoiding the run script and using `${CURRENT_PROJECT_VERSION}` within the `Info.plist` seems hopeful. I hope to set aside time this week to see if this is possible.

zorgiepoo:
> At the moment, Sparkle places an invalidly formatted version number in the Info.plist of the framework and AutoUpdate.app.

> Following Apple's documentation will best place Sparkle for use with tools and schemes that expect the version format. I do not believe what other products or projects do is useful, given the clear and precise nature of Apple's guidelines.

I think we're going in circles. It's not invalid. As @jkbullard mentioned, it historically is a marketing version and not constrained. Apple's documentation is *misleading* and tailored for developers shipping to the App Store. It is *not supposed* to be parsed (heck, it is localizable!) -- except by an arbitrary App Store validator for approving apps. It's equivalent to someone else like MacUpdate designing their own repository of apps and suddenly saying only "x.x" should be used even though this restriction is purely arbitrary. At best, you can argue it's inappropriate, but it's not invalid.

I don't think we ever used the git hash for the CFBundleVersion key by the way.

> zorgiepoo's comment about potentially avoiding the run script and using ${CURRENT_PROJECT_VERSION} within the Info.plist seems hopeful. I hope to set aside time this week to see if this is possible.

My comment was just regarding the short version string, since you already had appended the hash to the copyright key. The run script would thus still be needed.

grahammiln:
OK, as is, this pull request does not improve Sparkle.

PR #916, opened 2016-11-10
Move swift and other build settings into xcconfig files

zorgiepoo:
Note it's important that swift settings should not be included in target configurations that don't use swift. Feel free to suggest if this can be done better.


pornel:
Will removal of `LD_RUNPATH_SEARCH_PATHS` cause any problems?

zorgiepoo:
That was removed for the UI tests, but the UI tests xcconfig has `LD_RUNPATH_SEARCH_PATHS` defined so I don't think so.

PR #917, opened 2016-11-10
Set output path for Set Get Version Info script phase

ksuther:
This slightly reduces the amount of work during incremental builds. Without this Autoupdate.app would always have to get recopied into the framework. This would then cause the Sparkle framework to get recopied into the parent app.


ksuther:
Looks like this doesn't work 100% of the time. Having the output file specified will sometimes cause Xcode to try to run the build phase before Info.plist gets copied, which then causes a build failure.

Two workaround options are to manually copy Info.plist in the shell script or to embed this in a constant rather than Info.plist, but closing this for now.

PR #919, opened 2016-11-12
Don't append hash to versions of builds with a tag

pornel:


zorgiepoo:
Any way for this to not break if we suddenly go to 2.x?

If you want to improve the experience for people that use Sparkle as a subproject, you may have to use `PlistBuddy` to retrieve the current short version string, and only alter the Info.plist file if the string differs. I'm afraid that "only altering Info.plist when it's not an exact matching tag" may lead into problems.

grahammiln:
Will this still create invalid version numbers for those building from source, as a sub-project?

pornel:
@grahammiln it will use conforming version if you build it from source using checkout of a released version. If you checkout a commit that is not tagged with a version, then it'll add a hash to the version string.

pornel:
@zorgiepoo what problems do you envision?

zorgiepoo:
@pornel The key getting out of sync due to some iteration of commits, or switching branches, or not doing a clean build, or something along those lines.

jakepetroules:
This makes sense; it should be a tag OR a git hash. Not both.

Issue #419, opened 2014-09-04
Using the framework inside an Xcode project


DidierMalenfant:
I'm wondering how people are integrating the sparkle framework inside their projects. Basically I'm assuming I must be doing it wrong and wanted to sanity check.

I use a git submodules to bring the Sparkle master branch in to a sub directory of my project. I then add sparkle.framework as an entry in the 'Linked Frameworks and Libraries' and also as a item in my copy phase for frameworks (which copies into the Frameworks directory of my app bundle).

This works great until install time. I can't build a release version without a) changing the install path in the Sparkle project from ```@rpath``` to ```@executable_path/../Frameworks``` b) turning on 'Skip Install' to avoid a warning (using the install path with ```@executable_path``` but with 'Skip Install' off.

I'm assuming the project must build as a useable framework right out of the box for others so I'm wondering where I'm going wrong in setting this up.

PR #927, opened 2016-12-02
Allow the framework to be archived from Xcode and appear in the Organizer

nivekkagicom:
Using SKIP_INSTALL = NO allows one to keep a copy of the archived framework in Xcode's Organizer (Archives) directory as a more "permanent" archive of the framework and to export the built products for use in other applications as a stand-alone framework.


zorgiepoo:
There has been discussion about this in the past about whether this flag should be enabled or disabled (#409, #419). I personally think subproject integration is bad and should not be catered to, but some people may disagree with me ;), and decisions about it have been made in the past.

nivekkagicom:
Interesting. My app uses several third-party frameworks and this is the only one that uses SKIP_INSTALL = YES. I am not a fan of subproject integration either so clearly I am biased.

PR #933, opened 2016-12-10
Minor tweaks

pornel:


zorgiepoo:
I disagree that initializing the local variables and having early returns for this code is better. This is easily verifiable. Comment out one of the variable assignments, and the compiler will spit out a warning about a path that the variable has not been assigned -- that is very useful. Additionally the current code is more immutable (single assignment form) and doesn't short circuit logic anywhere, which could lead to bugs down the road.

pornel:
OK, I did not trust the warning.

I think short-circuiting logic — in small functions — is safer. This is because `return NO` is final, but a mistake in nesting of `if`s could assign `result = NO` and then fall into codepath that assigns `result = YES`.

zorgiepoo:
Unfortunately we don't have an attribute such as Java's `final` or Swift's `let` that would prevent multiple assignments. I'm actually indifferent to the changes to `validateWithUpdateDirectory:` but still for not initializing locals in the init method.

zorgiepoo:
Your latter code snippet is actually a poor example because the compiler will catch uninitialized reads, if you were to access res afterwards.

pornel:
OK, I've removed changes to init.

pornel:
I'm assuming the other changes are OK

PR #938, opened 2016-12-21
Unarchiving blocks

pornel:
I don't like delegates, so I've replaced it with callbacks.

I've added reporting of specific unarchiving error.

### DIFF
--- a/Sparkle/SUAutomaticUpdateDriver.h
+++ b/Sparkle/SUAutomaticUpdateDriver.h
@@ -13,7 +13,7 @@
 #import "SUBasicUpdateDriver.h"
 #import "SUAutomaticUpdateAlert.h"
 
-@interface SUAutomaticUpdateDriver : SUBasicUpdateDriver <SUUnarchiverDelegate>
+@interface SUAutomaticUpdateDriver : SUBasicUpdateDriver
 
 @end
 
--- a/Sparkle/SUBasicUpdateDriver.h
+++ b/Sparkle/SUBasicUpdateDriver.h
@@ -15,7 +15,7 @@
 #import "SUAppcast.h"
 
 @class SUAppcastItem, SUHost;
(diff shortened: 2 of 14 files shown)


zorgiepoo:
Hm ok.. I feel a bit indifferent about delegation vs blocks (latter where retain cycles need to be carefully avoided, but this code in question is fine..)

If you feel strongly about this, I would much appreciate a patch or merging work to be done against my fork :). I really think the fork should be pulled in a remote branch or something, be maintained to be in sync or so, else we deal with future costs straying away indefinitely from a modern architecture (with changes more impactful than minor refactoring).

My recommendation for nullability attributes, if you want to add them to a header, is to always use the `NS_ASSUME_NONNULL_BEGIN` and `NS_ASSUME_NONNULL_END` macros, and assume non null as the default.

pornel:
Great suggestion with `NS_ASSUME_NONNULL_BEGIN`. I'll update it.

I'm writing a tool to automatically build an appcast from a directory of archives, and for such case callbacks are a bit easier to use than delegates.

I'll see if I can use your fork.

zorgiepoo:
The appcast code in my fork should be pretty identical.

The unarchiving code is similar as well, though I've moved from subclassing -> protocols, and removed the dynamism code that registered unarchiver implementations using the objc runtime and global mutable state.

zorgiepoo:
Btw, not relevant here, but if you work with XPC, a reply block cannot be invoked more than once so that is one case where a delegate sort of object may be necessary (in the 'delegate' vs 'callback' discussion).

pornel:
I've change nullability syntax.

Added basic progress for DMG and delta.

zorgiepoo:
That's neat. If you want me to merge the changes in my fork, or if you want to try, let me know.

pornel:
Since you wanted to bring unarchiver changes in sync, I'll leave merging to you, as otherwise I'd just be getting in your way.

PR #942, opened 2016-12-21
Unarchiver protocol

zorgiepoo:
Here is my unarchiving code using protocols merged with your code using blocks @pornel 

I also had to fix a regression introduced recently due to how subclass method implementations not being called (this is one of the reasons my fork avoids subclassing entirely). But this also illustrates refactoring can be a little dangerous.. Manual testing is just as important as automated testing. In the test app, you'd see that after extraction, the app would immediately restart rather than prompt the user to install & relaunch using the UI scheduled driver.

Unfortunately the UI automated tests we have may only work sometimes and not be that stable at the moment..


zorgiepoo:
I tested this build in VLC.app, which uses a dmg for its feed. Having the determinate progress bar with more steps at the extraction stage is nice 👍.

Sparkle/SUBasicUpdateDriver.h
```diff
@@ -39,6 +40,10 @@
 - (void)extractUpdate;
 - (void)failedToApplyDeltaUpdate;
 
+// Needed to preserve compatibility to subclasses, even though our unarchiver code uses blocks now
+- (void)unarchiver:(id)ua extractedProgress:(double)progress;
+- (void)unarchiverDidFinish:(id)ua;
```
pornel: is that for 3rd party unarchivers?

Sparkle/SUBinaryDeltaUnarchiver.m
```diff
@@ -6,16 +6,25 @@
...
+
+@property (nonatomic, copy, readonly) NSString *archivePath;
+@property (nonatomic, copy, readonly) NSString *updateHostBundlePath;
+
+@end
 
 @implementation SUBinaryDeltaUnarchiver
 
+@synthesize archivePath = _archivePath;
+@synthesize updateHostBundlePath = _updateHostBundlePath;
```
pornel: BTW, can we use authosynthesized properties now? I don't remember what system requirements they had.

Sparkle/SUBasicUpdateDriver.h
zorgiepoo: No, 3rd party / custom update driver subclasses.

Sparkle/SUBinaryDeltaUnarchiver.m
zorgiepoo: I'd love to without breaking anything ;). The first step would to make sure all code is @synthesize a = _a first before moving to autosynthesizing which would mean some ivars need name changing.

Sparkle/SUBasicUpdateDriver.h
zorgiepoo: 1st party subclass too, actually.

PR #944, opened 2016-12-23
Simplified extension matching

pornel:
Dumb code is shorter, and doesn't make selectors from string.


zorgiepoo:
The changes in this PR is obsolete by #942 =P.

zorgiepoo:
Well what I mean to say is if you want to make a similar change, make it after #942 is in.

PR #945, opened 2016-12-23
Sorted project file

pornel:
I've used sort-Xcode-project-file.sh. It might help mergeability of the project file.


zorgiepoo:
The script is from [here](https://github.com/WebKit/webkit/blob/master/Tools/Scripts/sort-Xcode-project-file) right? I feel like the other outstanding pull requests should be merged in first, and once I merge those changes into my fork, then this can go in. And I should probably run this script on my fork rather than attempt a merge from this change.

pornel:
Yes, it's from webkit, and the idea is that if you sort it on your fork then the diff will be smaller.

zorgiepoo:
I merged the recent changes to my fork miraculously.

I don't think this is really going to help actually. My fork moves files around in appropriate places and I'm just afraid of trusting a project file merge. For example, the unarchiving code in master is in the Sparkle.framework target, but it is in the Autoupdate target in my fork -- and I wouldn't want a project merge to accidentally add files to the wrong targets. I also have to remove files that are master-specific from merges sometimes.

PR #946, opened 2016-12-23
Tool to generate appcasts automatically

pornel:
One-stop-shop with convention-over-configuration to generate (and update) appcasts.

You give it private key and a directory of tarballs, and it'll figure out all versions from plists, generate delta updates, and dsa-sign everything.

* assumes all updates can be unarchived and contain .app bundles
* assumes `SUFeedURL` is present in the plist, and archives are served from the same directory

It's not clever architecturally, I just wanted get stuff done.

I'm compiling everything in instead of linking Sparkle.framework, since not all symbols are public and mixing and matching was a pain.

### DIFF
--- a/generate_appcast/main.swift
+++ b/generate_appcast/main.swift
@@ -1,0 +1,51 @@
+//
+//  main.swift
+//  Appcast
+//
+//  Created by Kornel on 20/12/2016.
+//  Copyright © 2016 Sparkle Project. All rights reserved.
+//
+
+import Foundation
+
+func main() {
+    let args = CommandLine.arguments;
+    if args.count < 3 {
+        let command = URL(fileURLWithPath: args[0]).lastPathComponent;
+        print("Generate appcast from a directory of Sparkle updates\nUsage: \(command) <private key path> <directory with update archives>\n",
+            " e.g. \(command) dsa_priv.pem archives/\n",
+            " Appcast files and deltas will be written to the archives directory.\n",
+            " Note that pkg-based updates are not supported.\n"
+        )
+        exit(1);
+    }
+
+    let privateKeyPath = URL(fileURLWithPath: args[1]);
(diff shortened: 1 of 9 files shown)


generate_appcast/main.swift
```diff
@@ -0,0 +1,51 @@
...
+        print("Generate appcast from a directory of Sparkle updates\nUsage: \(command) <private key path> <directory with update archives>\n",
+            " e.g. \(command) dsa_priv.pem archives/\n",
+            " Appcast files and deltas will be written to the archives directory.\n",
+            " Note that pkg-based updates are not supported.\n"
+        )
+        exit(1);
+    }
+
+    let privateKeyPath = URL(fileURLWithPath: args[1]);
+    let archivesSourceDir = URL(fileURLWithPath: "/Users/porneL/ImageOptim/imageoptim/archive", isDirectory:true);
```
zorgiepoo: I'm guessing this isn't supposed to be hardcoded.

zorgiepoo:
Pretty cool. I might consider using it =).

Few things I noticed:

1. You have a hardcoded path for `archivesSourceDir` (likely a testing/debugging mistake)
2. My first thought was not including the `pubDate` field, although I guess older versions of Sparkle may need it..
3. The filenames generated from delta updates are based on the short version string, so what happens when you have multiple builds (more than two) that have the same short version string but differing build numbers.

zorgiepoo:
Oh yeah, how (or does this) handle release notes?

pornel:
I kept `pubDate` mostly because it was easy enough to implement and happened to just work :) I was positively surprised how well original date was preserved by OS X and my rsyncing of tarballs.

Good point about the deltas. Fixed.

It handles release notes by updating/merging existing appcast file instead of overwriting it. So you can edit the file and add notes manually.

PR #948, opened 2016-12-24
Add UI tests to travis CI and add 10.12 as latest SDK to check

zorgiepoo:
This may or may not work on Travis depending on it if allows the UI accessibility stuff. I just hope it doesn't hang..

Also, I noticed we have hardcoded versions for the SDKs and 10.12 wasn't included. Is it possible to do something a bit better so we don't forget to add new versions?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,9 +31,14 @@ ci:
 			( rm -rf build && xcodebuild -sdk "macosx10.$$i" -scheme Distribution -configuration Coverage -derivedDataPath build ) || exit 1 ; \
 		fi ; \
 	done
-	for i in {10..11} ; do \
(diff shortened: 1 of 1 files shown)


zorgiepoo:
Ack,

> Unable to run UI Tests because Xcode Helper does not have permission to use Accessibility. To enable UI testing, go to the Security & Privacy pane in System Preferences, select the Privacy tab, then select Accessibility, and add Xcode Helper to the list of applications allowed to use Accessibility If you believe this error represents a bug, please attach the log file at

kornelski:
I'm assuming travis as a product is dying, so there's no point waiting for that.

PR #950, opened 2016-12-25
Updater delegation

zorgiepoo:
Separate updater delegate methods into its own file, make optional delegate methods be able to return optional values, add nullability annotations.


pornel:
Organizing delegate methods is a good idea.

However, casting of the updater doesn't look nice. If the delegate methods require specific class, then the protocol abstraction is a lie from typesystem perspective.

zorgiepoo:
I agree it's not nice, but there are other factors that play into this. My fork doesn't have this issue because it's re-arcitectured to pass everything in the internal drivers instead of invoking methods on SUUpdater, but I don't want to do those changes or break compatibility _here_.

I also consider it a flaw the compiler can't do this:

```
@class SUUpdater; // don't #import "SUUpdater.h"
SUUpdater<SUUpdaterPrivate> *updater = ...;
updater.delegate; // error, forward-declaration class
```

Else I would have used `SUUpdater *` instead of `id`

Lastly, the decision to pass `SUUpdater *` to our delegate methods is *also* a [bad decision](https://zgcoder.net/ramblings/avoid-passing-the-delegator) that's not revertible now.

And then you'd have to ask if we benefit from restricting what we expose from SUUpdater, and perhaps that's useful even though the restriction is in between rather than exposing everything vs exposing nothing.

zorgiepoo:
Ok, I can do this without typecasting by working around the compiler defect by first using a temporary `id<SUUpdaterPrivate>` variable.. Hang on.

PR #951, opened 2016-12-26
Decouple and simplify installation code using protocols

zorgiepoo:
This also fixes a bug where the status progress window would be visible during interactive (non-guided) package installations. Possibly the last major refactor change to sync with my fork.


pornel:
Nicely untangles the code!

PR #953, opened 2016-12-27
Add xcconfig file for generate_appcast target

zorgiepoo:
Also remove swift configurations from Common config files, and change swift optimization from fast to whole module.

For some reason unknown to me, generate_appcast fails to build in Xcode 'All' or 'Distribution' target. This can be reproduced in master too, so I don't think it's related to this PR. `make release` works fine however.


zorgiepoo:
Ah, I think it's the target dependency ordering. Place `generate_appcast` after the other Sparkle targets for the Distribution and All targets; I trust you can make that change @pornel

pornel:
I was unable to reproduce the build failure. All and Distribution worked fine for me.

zorgiepoo:
The errors I am running into is symbols like SUError and SUStandardVersionComparator being unresolved (Swift code not finding those in the obj-c bridging).

One workaround for me I found I believe is adding Sparkle.framework as a target dependency to generate_appcast, and adding 'import Sparkle' in Appcast.swift and DSASignature.swift, to get both `make release` to work for me and building under Distribution target in Xcode. But this is kind of strange.. @pornel

pornel:
I've seen that briefly, but I think it's an Xcode bug, because it goes away after Clean Build Folder.

zorgiepoo:
Yeah, I think you're right..

PR #963, opened 2016-12-29
Remove the old app/bundle instead of moving it to the trash

### LINKED ISSUE #827: Old apps get sent to trash?
My users are noticing that every time they update the app, the previous version gets sent to the trash. Is this a recovery mechanism by design? The trash is growing daily everyday, which is not something I would want since it's going to eat their hard disks too easily. I just want to remove it completely.

zorgiepoo:
Fixes #827


pornel:
So much deleted code, yay :)

Issue #962, opened 2016-12-28
Remaining leftover changes from my fork


zorgiepoo:
Here I'll keep an update a track of a list of some "small" changes from my [fork](https://github.com/zorgiepoo/Sparkle) that *could* possibly be made. I may not have much time left so this is what I want to focus on. Let me know what you think should or should not be addressed in the near future.

* [x] Expect the `length` attribute for appcast items to exist.
   * Warn if the attribute is not specified (a fair number of developers leave it out)
   * Use this field value instead of value from http header (what if that value is unknown?)
   * Warn if the attribute != expected content value from http header (if both are available)
   * Note or be aware that the download did receive responses can be received multiple times in rare cases for a single download according to the docs.
* [x] Make guided packages the default rather than interactive packages being the default for packages.
   * Opt into `foo.sparkle_interactive.pkg` or `foo.sparkle_interactive.mpkg`
   * Interactive installations have real problems (lack of success status handling, strange task spawning code) and worse user experience.
   * Some developers unpleasantly ship e.g. `foo.sparkle_guided.pkg` for their website distribution.
* [x] Remove the old application instead of moving the old application to the trash (#827)
   * Not seamless: clutters user's Trash; Sparkle is the only updater that does this.
   * More inefficient across volumes potentially, strange issues (#752)
   * Moving old app to trash may not always be possible/easy in my fork (I think, regarding running as root).
   * Could make atomic-safe directory replacements more difficult in the future with APFS
   * Would cause removal of some code
* ~~Don't ask for update permission after 3 (or x) number of hours, instead of 2nd launch~~
   * ~~2nd launch is a poor metric; user could launch application more than once in a short period of time.~~
   * ~~In my fork where support for updating *other* bundles is improved, the lifetime of the updater and bundle to update are distinct, and the updater be started up repeatably and not ran for long (e.g. `sparkle-cli`)~~
* [x] Rewritten `SULog()` usage
   * Let us stick with reporting errors for the time being. We shouldn't be logging information unless something unexpected happen (i.e, don't always log when extracting update).
   * A global file location (in master) to write logs is not a good idea. It is rewritten when multiple applications are using Sparkle simultaneously, and when multiple processes are using `SULog()` simultaneously (very common in my fork). Always writes to disk.
   * Should take leverage of Apple's logging APIs; my fork uses ASL APIs currently. 10.12 introduced a new logging API that I haven't looked into yet.
[...]

PR #964, opened 2016-12-29
Make guided package installs the default for updating packages

zorgiepoo:
Interactive package installs now need to be opted into (foo.sparkle_interactive.pkg or foo.sparkle_interactive.mpkg)

The website documentation may need to be updated later, and this should be noted as an important point in the change log.


pornel:
You've mentioned that people were publicly serving the installer under the same `sparkle_guided` name as it's used for Sparkle. Given this, we could expect people to reuse the same URL the other way too.

Given that, could we use another, more neutral name for this? (`*.installer.pkg`?) Maybe another opt-in method?

zorgiepoo:
> You've mentioned that people were publicly serving the installer under the same sparkle_guided name as it's used for Sparkle. We could expect people to reuse the same URL the other way too.

It's not the URL, but the filename after being unarchived that's important. People would still need to serve `sparkle_guided` for a while I guess for older Sparkle versions.

> Given that, could we use another, more neutral name for this? (*.installer.pkg?) Maybe another opt-in method?

Are you suggesting to rename `sparkle_interactive` to `installer`? Not sure if that would make more sense, or if I'm not following.

pornel:
Sorry, `s/URL/filename/`. I mean that it ends up being a user-visible name.

Yes, I've meant renaming `sparkle_interactive`. This name is odd outside context of the appcast. A name like `foo.installer.pkg` is a bit more neutral.

But overall maybe we should change the flag to use an appcast element or an attribute? e.g. `sparkle:install="interactive"`?

zorgiepoo:
> Yes, I've meant renaming sparkle_interactive. This name is odd outside context of the appcast. A name like foo.installer.pkg is a bit more neutral from user perspective.

That makes sense from a user-perspective. From a developer perspective, it's guided vs interactive. So, sure, we could rename it.

> But maybe instead of using filename at all, we should change the flag to use be appcast element or an attribute? e.g. sparkle:install="interactive"?

Already have this in my fork for different reasons explained in #962. However it's not a substitute for the unarchived filename due to considering the feed to be untrusted by itself for this.

pornel:
> Already have this in my fork for different reasons explained in #962. However it's not a substitute for the unarchived filename due to considering the feed to be untrusted by itself for this.

What's the attack scenario, especially when the default is guided? I don't see any serious attack here, because the archive contents itself is trusted, so the attacker can't do anything other than small nuisance.

PR #968, opened 2016-12-31
Hide automatic updates checkbox for information only updates

Bi11:
It's a bit of odd to show 'Automatically download and install updates in the future' checkbox if an update has only information but no downloads.
Even the checkbox can be hidden by SUAllowsAutomaticUpdatesKey, I think it would be better to be determined by appcast, instead of disabling auto downloads in host app permanently.


pornel:
Yes, that makes sense. Thanks!

PR #973, opened 2017-01-03
Resume pending automatic update when possible

### LINKED ISSUE #970: Show automatic update alert from checkForUpdates if an update is ready
We don't have a good indication when an automatic update has been downloaded and is waiting to be installed.

Currently invoking `checkForUpdates` aborts previous update and checks again, showing UI as if the update was discovered for the first time. This is misleading and makes users think auto-updates didn't work: https://github.com/ImageOptim/ImageOptim/issues/181

Using `[self.driver showUpdateAlert]` would display a more appropriate message, and as a bonus, it wouldn't throw away already-downloaded update.

pornel:
Fixes #970


zorgiepoo:
Looks nice 👍. My fork solves this by resuming updates in a different way, but this is good short term.

PR #980, opened 2017-01-05
Updated README

pornel:
Writing a home-grown updater that downloads a zip file isn't too hard, but doing it securely and reliably *is* hard, so I've tweaked the readme to emphasise this.


zorgiepoo:
On another note, I find it kind of weird we still highlight "changes from 1.5b" as part of the Read Me. That was kind of a while ago (and the security vulnerability helped obsoleting it too), and we have a change log file. Don't even use the "finish_installation" name anymore, eg.

pornel:
Yes, I've removed the finish_installation line, and moved that section lower.

PR #983, opened 2017-01-07
Upgrade SULog to use logging APIs that Apple provides built in

zorgiepoo:
With this change, we ensure distinct per-app, per-process logging.
Note that the global file ~/Library/Logs/SparkleUpdateLog.log is no longer written into.

On 10.12 and above, SULog uses the os_log APIs. You may want to use log(1) to view logs in addition to Console.app. See the WWDC video on the new logging API and how the Console app has changed.

On older systems, this uses ASL (Apple System Log).

We don't want to reinvent logging, or rely on 3rd party libraries, thus we use what the system provides to us.

Two logging levels were added: Default and Error. The code documents their intended purpose. Now when you call SULog you must make a decision which level to use rather than rely on an implicit level.


pornel:
Looks good. I'd wait with merging this until 1.16 is released, to avoid changing too much at once

PR #986, opened 2017-01-07
Switch to LZMA for the distribution archive

pornel:
7.3MB -> 4.3MB

Previous Sparkle distributions were around 4MB. The archive got much larger in 1.16 due to generate_appcast tool having a swift runtime.

### DIFF
--- a/Configurations/make-release-package.sh
+++ b/Configurations/make-release-package.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -e
+set -eo pipefail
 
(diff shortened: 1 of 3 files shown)


zorgiepoo:
I don't know how to extract the `xz` archive using Archive Utility. We can't expect people to use 3rd party unarchiver utilities. Not sure how Cocoapods is going to handle this format either; that would have to be tested.

pornel:
Oh my. Indeed. I assumed it'd work.

PR #990, opened 2017-01-10
Tell OS when networking happens without UI

pornel:
The docs say it helps OS schedule priority of the download and choose between wifi and cellular.


zorgiepoo:
Seems good.

PR #993, opened 2017-01-11
Delete trailing spaces

mathbunnyru:


zorgiepoo:
An innocent looking change, but this may cause me pain to merge to my re-architecture fork.

Also, contributors are going to make leading whitespace changes even if this becomes merged. That, and I don't think this is a big deal :/.

mathbunnyru:
@zorgiepoo thanks. I've just started working on this project. Trailing spaces are pain to me when navigating through the code. Could you check please, does this change is really difficult to merge? It seems to me, like all or almost all lines are empty (with spaces) in the end and it may be easy to merge.

PR #999, opened 2017-01-15
Adds test cases for "sparkle:criticalUpdate" tag.

tempelmann:
Please note that I only actually tested the one added to the Test App (which would bail due to an assert() call if the tag is not correctly identified), while someone else please test if the swift based XCTest works as well.

### DIFF
--- a/TestApplication/SUTestApplicationDelegate.m
+++ b/TestApplication/SUTestApplicationDelegate.m
@@ -10,11 +10,41 @@
 #import "SUUpdateSettingsWindowController.h"
 #import "SUFileManager.h"
 #import "SUTestWebServer.h"
+#import "SUUpdater.h"
+#import "SUUpdaterDelegate.h"
+#import "SUAppcast.h"
+#import "SUAppcastItem.h"
+
+#pragma mark - Updater Delegate
+
+@interface SUTestUpdaterDelegate : NSObject  <SUUpdaterDelegate>
+@end
+
+@implementation SUTestUpdaterDelegate
+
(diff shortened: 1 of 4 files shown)


zorgiepoo:
Thank you for the PR. Travis says the Swift one works.

I would prefer to not add these changes to the TestApplication, and to keep the assertions in the unit tests which is more of an appropriate place. I see the reason you added it to the Test Application is because you didn't verify the unit tests.

In the unit tests, could you add assertions for the other items, to verify that they shouldn't be critical updates? (is that right?)

tempelmann:
Fair enough. I just needed the one in the test app for my own testing, and you can use the swift code, which is the cleaner place, indeed.
I rather not make more changes to the Unit Test because I am currently on OSX 10.10, with an Xcode version that doesn't run the tests at all, hence I could not verify them.

zorgiepoo:
> I rather not make more changes to the Unit Test because I am currently on OSX 10.10, with an Xcode version that doesn't run the tests at all, hence I could not verify them

You can verify them by pushing the code here. If it doesn't work, Travis (our continuous integration) will spit out an error and GitHub will indicate the error appropriately.

tempelmann:
Alright. Will do in a few minutes. Do I have to remove the code from the test app as well or can you selectively skip those?

tempelmann:
Alright, I've added the other checks.

zorgiepoo:
> Do I have to remove the code from the test app as well or can you selectively skip those?

It'll be easier on our end if you remove changes from sparkletestcast.xml and SUTestApplicationDelegate.m

tempelmann:
Done.

PR #1000, opened 2017-01-16
Add Obj-C generics where applicable

zorgiepoo:
Just some small generic changes I'm sync'ing. We've already been using generics in private files.

In headers that are exported, we test to see if generic support is available.


pornel:
Looks OK, but I think number of ifdefs could be reduced by using a typedef, e.g. `SUDictonaryOfStrings`.

zorgiepoo:
Well, the preprocessor conditional is used when exporting an API that a developer may use. If we use a define or typedef, then they may go "what the heck is this?" when they look up the method signature, see it in the auto-type completion, etc.

PR #1003, opened 2017-01-17
Add forward declaration for Xcode7

Bi11:
`+ (instancetype)buttonWithTitle:(NSString*)title target:(id)target action:(SEL)action;` 
belongs to `NSButton`.


Bi11:
I found the SDK checking style in Sparkle looks like
`#if __MAC_OS_X_VERSION_MAX_ALLOWED < 101200`
which follows the Apple's Guide [Conditionally Compiling for Different SDKs](https://developer.apple.com/library/content/documentation/DeveloperTools/Conceptual/cross_development/Using/using.html#//apple_ref/doc/uid/20002000-1114741-CJADDEIB)
In my recent added SUTouchBarForwardDeclaration.h, I used 
`#if !defined(MAC_OS_X_VERSION_10_12_1)` style
which can be found in Chromium project.

Should I change it?

zorgiepoo:
Yeah, sure, for consistency and following guidelines if nothing else.

PR #1004, opened 2017-01-18
Test if master can still build successfully

zorgiepoo:


pornel:
https://www.traviscistatus.com/

they have backlog of about 300 jobs now

Bi11:
I added a build matrix for both xcode 8 and 7
```
matrix:
  include:
    - osx_image: xcode8.2
    - osx_image: xcode7.3
```
Travis will create multi jobs for a build. See an example
https://travis-ci.org/Bi11/Sparkle/builds/192947694

But according to the build log, Travis failed on xcode 7, blocked by compile swift of genearate_appcast.

Should we fix the build for xcode 7, or drop support of xcode 7, or just ignore it.

zorgiepoo:
Xcode 7 won't be build to run generate_appcast and run the unit tests. Somehow we'd have to skip those if we wanted to add Xcode 7 to our continuous integration.

PR #1006, opened 2017-01-18
Use app-independent subsystem identifier

pornel:
I've noticed that macOS doesn't customzie subsystem identifiers for each app, e.g. `com.apple.defaults `, `com.apple.securityd` are used for all processes.


zorgiepoo:
Ok, seems fair. I think the particular the process that generated a log message can still be retrieved/filtered.

PR #1010, opened 2017-01-20
Termination autoupdate

zorgiepoo:
I set the base branch to 1.17.

This is the termination listener my fork uses. It is not based on AppKit (so it works for non-regular processes) and it is not polling / timer based. Complexity is increased a bit. The termination listener accepts an optional/number instance but doesn't really have to -- but that isn't a high priority I want to untangle related to the code in my fork at the moment.


pornel:
Looks great

PR #1016, opened 2017-01-24
Report installation progress from Installer

pornel:
Installer has a machine-parseable output. I've dragged it all the way up to the UI.

### DIFF
--- a/Sparkle/Autoupdate/Autoupdate.m
+++ b/Sparkle/Autoupdate/Autoupdate.m
@@ -136,7 +136,7 @@ static const NSTimeInterval SUTerminationTimeDelay = 0.5;
         self.statusController = [[SUStatusController alloc] initWithHost:host];
         [self.statusController setButtonTitle:SULocalizedString(@"Cancel Update", @"") target:nil action:Nil isDefault:NO];
         [self.statusController beginActionWithTitle:SULocalizedString(@"Installing update...", @"")
-                                   maxProgressValue: 0 statusText: @""];
+                                   maxProgressValue:100 statusText: @""];
         [self.statusController showWindow:self];
     }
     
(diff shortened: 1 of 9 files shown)


zorgiepoo:
I haven't reviewed / looked into the code in detail yet, but what about the 2.0 branch? Is this worth merging this feature in there too? Undoubtedly it's going to be a little different of a scenario. Remember, that branch needs to be maintained too now =).

zorgiepoo:
Err, that's not meant to scare you off, by the way. I don't think it will be that difficult to apply the same changes in a new PR to 2.0 (it will just be a little different, perhaps even easier because there's no intermediate fileop tool. The UI progress is in a different agent though, but shouldn't be hard to communicate.).

In the future, that kind of needs to be considered though. If a not-small enhancement should be prioritized, which may require creating a little more effort and testing.

pornel:
Yup. I haven't tried merging it into 2.0 yet, but in general changes like this are why I would like to bring logical structure of both branches as close as possible.

zorgiepoo:
From my point of view, they are as close (or almost) as possible without creating a huge amount of wasted effort, or compatibility breakage, or debugging, or much extra testing work in between =).

But this is why I think you have to look at both branches before deciding to work on a feature, or before pulling in a feature, and the first decision should be to prioritize whether it's worth working on at this time.

pornel:
It would be very unfortunate if incompatibility of the other branch prevented Sparkle from getting improvements now.

zorgiepoo:
It doesn't, but it makes you need to be aware of creating changes in an older branch, for the maintenance / costs, or extra effort that needs to be put in sync, that is involved.

In my experience, It's not all that different when developing the next major version of an application, and still maintaining the stable version.

pornel:
Apart from merging issues, is that code OK?

zorgiepoo:
I'm not really sure about the line parsing code. Maybe even aside from merging issues (which isn't that great, but effort needs to be put in), the feature should be deferred (greater complexity of code with benefits that doesn't affect most use cases).

Btw, I think `performFinalInstallationProgressBlock` should be renamed to `performFinalInstallationWithProgressBlock`

pornel:
Yeah, the merging code is bad. I wanted to avoid naive solution that's potentially quadratic, but this one is potentially buggy instead :)

Issue #1017, opened 2017-01-27
Delta update verification fail 


GoodMorningCall:
## Where
CFBundleName (CFBundleDisplayName) is localized.
e.g. エンジン

## Reason
The expected .app folder name is different from its actual file system representation.
e.g. 
CFBundleName: エンジン
 -> UTF8 Hex: 0xE30x820xA80xE30x830xB30xE30x820xB80xE30x830xB3
File system representation: エンジン
-> UTF8 Hex: 0xE30x820xA80xE30x830xB30xE30x820xB70xE30x820x990xE30x830xB3

## Source File
- Sparkle/Unarchiving/Binary Delta/SUBinaryDeltaCommon.m
### Method
- hashOfTreeWithVersion()
### Code
line#175

`       NSString *relativePath = pathRelativeToDirectory(path, stringWithFileSystemRepresentation(ent->fts_path));
`

## Update Suggestion
`        NSString *relativePath = pathRelativeToDirectory(                                                         stringWithFileSystemRepresentation(pathBuffer), stringWithFileSystemRepresentation(ent->fts_path));
`

zorgiepoo:
Thank you for pinpointing this issue. Would you like to create a pull request for this issue, or would like us to handle making the changes you suggested?

I think we may be able to just get away with increasing the minor version, instead of the major version for the binary delta format, considering the unit tests will still pass and it may not affect many apps.

GoodMorningCall:
Hi Mayur,

Thank you for your quick response.

I would like you to handle it because I don't know the way to create a pull request. I've never done it before. If I do it asking for Google, it will take a whole day and I think I'll definitely make a mistake ^^;

Kind regards,
- Takayuki Ogino

pornel:
Oh, great find! That may be the problem I've been seeing #998  — I do have localized name in my app too.

pornel:
I wasn't able to reproduce it just by adding decomposable Unicode characters to delta filenames. But just in case I've added normalization you suggested, and another when getting CFBundleDisplayName. #1018

Does that fix the issue for you?

GoodMorningCall:
Yes, it fixes my issue.

Delta filename is not my case. My delta filename is ASCII, maybe it comes from archive name(mine is Engine1.00.dmg which I named when I created it).

My problem is localized app name which is used for the app menu title.
I use InfoPlist.strings for localization, and define CFBundleName/CFBundleDisplayName in it.
[...]

PR #1018, opened 2017-01-27
fileSystemRepresentation normalization

pornel:
#1017


zorgiepoo:
Increase binary delta minor version?

Issue #969, opened 2017-01-02
Use alternative method of configuring interactivity of installer updates


pornel:
Requirement to have `*.sparkle_*.pkg` filename means that authors can't use the same file download for both user-facing downloads as well as updates, without exposing Sparkle-specific filename to users.

We could use another method, e.g.

* Extract pkg and parse `Distribution` file to find whether there's anything for user to configure, and show installer UI only if there are any configuration options available
* Use a tag or attribute in the appcast, e.g. `sparkle:install="interactive"` 

Related #964

zorgiepoo:
To answer your question in #964, it's not something I thought about that much, but I'll let you know what are the implications.

First off, there are three installation types: "application", "package", "interactive-package" (these are the names my fork uses, but these could be renamed..). For the optional `sparkle:installationType` attribute, "application" is the default.

Under a modern architecture (my fork), guided package installations implies that authorization will always be needed up front, before launching Autoupdate. And "interactive" implies the opposite, authorization will never be needed up front (because the pkg will request it). For a regular "application", needing authorization depends on other factors.

So, if the appcast says it's a guided pkg (implying to authorize Autoupdate), and after extracting the archive, what is found is actually a regular application, what should we do? My fork bails out an error. I think that's reasonable.

If the appcast says it's a guided pkg, (implying to authorize Autoupdate), and after extracting the archive, what is found is actually an interactive pkg installer, what should we do? Still just execute it as a guided pkg anyway? What will happen if there are configuration options available, will it just fail gracefully?

If the appcast says it's an interactive pkg (implying to not authorize Autoupdate), but its actually a guided pkg, what should we do? Still just execute it as an interactive pkg anyway? I suppose I can't anticipate issues arising from that.

Btw, is there documentation available describing that guided installation won't work with special configuration? I see the man page on `installer` has some kind of config flag.

zorgiepoo:
I do think the option to use interactive vs guided may want to be in the developer's control, in case one such installation method is fault (re: identifying Distribution to pick when to show UI idea).

pornel:
I was referring to your comment that the appcast is untrusted, so was wondering if attacker being able to change installation type can exploit something. I don't see much potential for attacks:

- attacker could change guided to interactive, which is merely annoying
- attacker could change interactive to guided. This, I presume, will cause installation to proceed with default options or fail. That doesn't allow any arbitrary code execution.

So I think it's safe to allow interactivity to be controlled from the appcast.

[...]

PR #1022, opened 2017-02-04
Add support for updates with flat packages

### LINKED ISSUE #54: Support for flat .pkg files that are not compressed
_**martoche (martin-ottenwaelter)_\* [reported (on Launchpad)](https://bugs.launchpad.net/sparkle/+bug/342901) on' 2009-03-14:*

Mac OS X 10.5 introduced flat .pkg files that are already compressed. It
doesn't make sense to .zip or .tar.gz a flat .pkg file, they can be
downloaded as-is.

Sparkle assumes that all downloaded files should be unarchived, which is
not the case for a flat .pkg file.

Attached to this report is a patch that solves the problem. The patch
consists in one new class "SUNoOpUnarchiver" which is a SUUnarchiver
subclass that registers itself as the unarchiver for flat .pkg files,
but does nothing when asked to unarchive the file.

ylatuya:
Flat packages are already archived and compressed so they
can be distrubuted as they are without the need to apply
and extra tar/zip archiving.
This commit adds a new unarchiver implementation that does
nothing with udpates using the .pkg or .mpkg extension

Fixes #54

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -5,6 +5,7 @@ Copyright (c) 2015-2017 Mayur Pawashe.
 Copyright (c) 2014 C.W. Betts.
 Copyright (c) 2014 Petroules Corporation.
 Copyright (c) 2014 Big Nerd Ranch.
+Copyright (c) 2017 Andoni Morales Alastruey.
 All rights reserved.
 
 Permission is hereby granted, free of charge, to any person obtaining a copy of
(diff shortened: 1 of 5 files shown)


zorgiepoo:
Thank you for the pull request.

One potential issue is that we carry an extra bit of information after extraction, which is telling if the package is interactive or guided. That bit of information may only be trusted inside the archive, since the name of the archive may not be trustworthy. Related #969

As the PR already does, we will still want to keep compatibility with archiving flat pkg's.

`-unsafeIfArchiveIsNotValidated` returns `NO`currently for the new unarchiver implementation, but may ought to should return `YES` in my [branch](https://github.com/sparkle-project/Sparkle/tree/ui-separation-and-xpc).

ylatuya:
It's my first time integrating Sparkle into an application and I still don't know much the codebase, but if I understood correctly Sparkle's flows, application developers could still choose between guided or interactive mode by naming their .pkg in the same way they'd do when archiving the flat package and with the same implications. We'd like to avoid having to provide 2 different files for each release, archived for Sparkle and flat package for our website :) In both scenarios, after the unarchiving step, Sparkle will see the same output results, myapp.pkg or myapp.sparkle_interactive.pkg. It's the author's responsability to provide correct links and filenames, in the same way it has to be done when publishing an archive since the the unarchiver only relies in the filename extension.

Is there anything that needs to be changed to the get this PR merged?

zorgiepoo:
I'm thinking of the scenario where an attacker comes in and changes the pkg filename, such that Sparkle is fooled into thinking a pkg should be ran guided even though the developer wanted it to be ran interactively -- although I'm not sure of the potential consequences. (If the pkg is archived, then pkg filename is trusted).

Perhaps one way is to decide that interactive pkg's shouldn't be possible if they aren't archived. I agree  in the end that shipping non-archived pkg's (when developers are able to, have proper MIME support on their server) should be possible. On another thread that may be affected, I'm also trying to push for #1023 :).

Issue #1032, opened 2017-02-10
generate_appcast should add `sparkle:shortVersionString` to the enclosure


digitalmoksha:
This is more apparent when you use a version string that is a build number, like `310` and a short version like `1.5.4`

Without the `sparkle:shortVersionString`, the update dialog will show 

   "My App 310 is now available - you have 1.5.4"

pornel:
Definitely. Would you be able to make a pull request for this?

PR #1034, opened 2017-02-10
added `sparkle:shortVersionString` to the enclosure, #1032

digitalmoksha:
added `sparkle:shortVersionString` to the enclosure, #1032


pornel:
Thank you

PR #1036, opened 2017-02-14
Disable the keyboard shortcut for the install button for scheduled updates to avoid accidental installs.

gnachman:
Scheduled updates open a key window at unpredictable times. Users find that they sometimes press <kbd>Enter</kbd> just as the window appears, causing an interruption in their work.

Example: https://gitlab.com/gnachman/iterm2/issues/5549

This commit removes the keyboard shortcut for scheduled updates, but preserves it for user-initiated updates.


pornel:
Thank you

PR #1039, opened 2017-02-20
Fix pt_BR translations

fujisoft:
Small update from the last fix.


pornel:
Are you sure about change of … to ...?

pornel:
Thanks

PR #1041, opened 2017-02-22
Added Swift friendly method. Please double check that I didn't create…

jaspws:
… a retain cycle - it's been a very long time since I've used objc blocks


pornel:
I'm not convinced we should add such method. It's a small improvement, but the the existing method should be usable from Swift.

PR #1044, opened 2017-02-28
Fix Japanse localization

1024jp:
Generalize message to user as the previous setence was more like "Anonymous system profile information is used to help **our company** plan future development work."


pornel:
Thanks

PR #1076, opened 2017-05-11
Prevent updating from running on first launch

willco007:
On first launch Sparkle will run if the shouldPrompt value is NO as a result of calling scheduleNextUpdateCheck.

### DIFF
--- a/Sparkle/SUUpdater.m
+++ b/Sparkle/SUUpdater.m
@@ -221,7 +221,7 @@ static NSString *const SUUpdaterDefaultsObservationContext = @"SUUpdaterDefaults
             [self resetUpdateCycle];
         }];
         // We start the update checks and register as observer for changes after the prompt finishes
-    } else {
+    } else if (hasLaunchedBefore) {
         // We check if the user's said they want updates, or they haven't said anything, and the default is set to checking.
         [self scheduleNextUpdateCheck];
     }


pornel:
Thanks for the PR!

The shouldPrompt setting strictly speaking refers only to the update check permission prompt (where Sparkle asks if it's OK to contact server to check for updates), rather than all update prompts.

I realize it could be useful if Sparkle was completely silent on the first launch, and didn't even prompt to update if an update is found, so we might to change that.

@jakepetroules what do you think?

This particular change would also prevent silent autoupdate on exit from running, but currently this wouldn't be a problem, because user needs to tick "allow" box first. If we ever allow app authors to have completely silent autoupdates, then we'd need an exception for silent autoupdates too.

willco007:
Ah, sorry about that. I was unaware of this subtle behavior. It is sort of nice to avoid interrupting on boarding on first launch which is what I was avoiding.

jkbullard:
Would this PR make it _impossible_ to have an app check for updates on its first launch?

Tunnelblick checks on the first launch because it is an app with parts that execute as root. For many years, Tunnelblick has set the "SUHasLaunchedBefore" default to cause Sparkle to allow a check on the first launch. Will that still work?

jakepetroules:
Most of the world has moved more to a model of automatic, silent updates vs explicit prompts so I think it's fine to potentially have an update prompt on the first launch - developers want users to be updated.

It's very unlikely that users will run an old version of an app for the first time. For the most part, a user will download an app and run it immediately.

willco007:
If you set SUHasLaunchedBefore to yes it will work as it always has.

jkbullard:
@willco007 - Thanks. That's what I needed to know.

@jakepetroules - Yes, but our app is often distributed by others (i.e., not from our website) and thus it is very easy for a user to download an out-of-date version of the app.

kornelski:
I'm going to close this.

We already avoid asking for permission on the first run.

We do show update, if available, on the first run. 

* This won't happen for most users who download the latest version from official source.
* If users got an outdated copy of the app from another source, there are both pros and cons of showing the update. 
* Not always showing the update may make it harder for developers to test if updates are working.

PR #1081, opened 2017-05-27
Prepend app name to Autoupdate.app name

pornel:
SPARKLE_RELAUNCH_TOOL_NAME config requires recompilation of Sparkle, so it's a lot of hassle.

I've realized that we don't need it. Since we're launching a copy of the Autoupdate.app, not the original in Sparkle.framework, we can rename it on the fly.

I've went for `$AppBundleName (Autoupdate).app`. It looks OK in the auth dialog.


jkbullard:
👍

PR #1084, opened 2017-06-01
Don't assume that the progress block exists (the parameter is nullable)

ksuther:
SUPlainInstaller was crashing when using a nil progress block.
This also affects the ui-separate-and-xpc branch.


pornel:
Thank you. That's an important fix!

I'm surprised that the compiler did not catch that :(

PR #1085, opened 2017-06-02
Fix crash by accessing nil callback when replacing binary

naan:
Ref https://github.com/sparkle-project/Sparkle/issues/363

I was testing `ui-separation-and-xpc` branch and found one crash issue when AutoUpdate tries to update downloaded binary. 

I'm not exactly sure where I should send those progress value to UI, so I just made an empty callback to avoid the crash at this moment.


naan:
I revert changes since the method is annotated `nullable`, so it should accept nil. Instead of passing empty callback, add nil check before calling progress callback.

naan:
ah, someone caught same issue in #1084. Closing this :)

PR #1086, opened 2017-06-04
Updated translation to reflect change on 55be13c

vitu:
Updated translation to reflect change on “Update error message to mention temp folders rather than optical drives”.


pornel:
Thank you

PR #1087, opened 2017-06-04
Updated Japanese translation to reflect change on 55be13c

1024jp:
The same update as #1086 but for Japanese localization.


pornel:
Thank you

PR #1089, opened 2017-06-07
Add support for the sparkle:os attribute in the enclosure tag

Memphiz:
This allows to use the same appcast xml file for winsparkle and sparkle. This feature is also already stated in the sparkle wiki here:

https://sparkle-project.org/documentation/publishing/#publishing-an-update

under "Alternate download locations for other operating systems"

But it was not implemented yet.

Behavior is as follows.
1. If no sparkle:os attribute was found -> assume its an osx update item
2. If sparkle:os attribute was found -> check if it contains "osx" as value. If it hasn't this item is not assumed to be an osx update item


pornel:
Thank you

PR #1090, opened 2017-06-09
Added initWithUpdaterDelegate:userDriverDelegate: to SPUStandardUpdaterController

ksuther:
SPUStandardUpdaterController didn't have a way to cleanly create it outside of `awakeFromNib`.

This adds `initWithUpdaterDelegate:userDriverDelegate:` as an alternate to relying on `awakeFromNib`.

If an app is distributed to the Mac App Store you can't easily put SPUStandardUpdaterController in the nib (Apple rejects apps for referencing Sparkle in the app or nib). This would probably get through approval initially since this isn't the main Sparkle branch, but that isn't a good assumption for the future.


pornel:
Hi. Thanks for the PR. This looks quite sensible.

PR #1091, opened 2017-06-10
Remove unused escapeURLComponent

ksuther:
`escapeURLComponent` migrated to `SPUUpdater`, it doesn't need to be in `SUUpdater` anymore.


pornel:
Thanks

PR #1093, opened 2017-06-16
Modifications to Allow Sparkle to Integrate with Versions & Kaleidoscope

jimrutherford:
Made changes to the AutoUpdate target to let Xcode manage signing on this target.  Without this setting, our custom Build Phase Script title "Resign Everything" was failing on with an error that "AutoUpdate was never signed so cannot be resigned".  This simple change solved the error for both the Kaleidoscope and Versions project.

How to test:  Pull down the branches where this project is integrated into Versions and Kaleidoscope and verify that the projects build and run.


pornel:
Thanks. This is obviously very specific to your setup. 

I do enjoy the similarity of works in re-signing and resigning ourselves to Xcode's fussy signing.

PR #1094, opened 2017-06-19
Include bin files in CocoaPods installation

keith:
This updates the podspec to include the bin scripts in Pods directory
when using CocoaPods. This way users don't also have to download the
Sparkle repo out of band to sign releases.


pornel:
Thank you

keith:
Thanks!

PR #1097, opened 2017-06-22
Update German localization for informal style and changed terminology

Eitot:
Apple switched to an informal style in its German localisations in macOS Sierra ([source](https://www.mactechnews.de/news/article/macOS-Sierra-redet-den-Nutzer-mit-Du-an-164371.html)).

Apple also seems to be using the words "Update" instead of "Aktualisierung" and "laden" instead of "herunterladen" (as seen in the App Store section in System Preferences). This might have been a recent change too, I am not sure.

I also made some slight changes to match the source strings better.


pornel:
Thank you

PR #1098, opened 2017-06-23
Update Dutch localization for informal style and changed terminology

Eitot:
As in #1097, Apple seems to have changed the form of address in Dutch too. I also made some stylistic changes that reflect Apple's own localisations a bit better.


pornel:
Thank you

PR #1099, opened 2017-06-23
Use "class" property attribute where available

Eitot:
Clang in Xcode 8+ supports this attribute to declare class accessors as class properties. They provide consistent semantics with Cocoa singletons and other accessors in both Objective-C and Swift.

More information: https://useyourloaf.com/blog/objective-c-class-properties/.

The preprocessor macro can be found here: [llvm-mirror/clang/lib/Lex/PPMacroExpansion.cpp#L1169](https://github.com/llvm-mirror/clang/blob/a38ba9770a70056f9fdd6c71f819e5db45a105e4/lib/Lex/PPMacroExpansion.cpp#L1169).

### DIFF
--- a/Sparkle/SUFileManager.h
+++ b/Sparkle/SUFileManager.h
@@ -21,7 +21,11 @@ NS_ASSUME_NONNULL_BEGIN
 * Creates a file manager that will not authorize for file operations
 * @return A new file manager instance
 */
+#if __has_feature(objc_class_property)
+@property (class, readonly) SUFileManager *defaultManager;
+#else
 + (instancetype)defaultManager;
+#endif
 
 /**
  * Creates a file manager that allows authorizing for file operations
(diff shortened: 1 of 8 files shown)


pornel:
I'm not sure about this. The `#if/#else` duplicate code. I don't have Xcode 7, which makes the alternative code path less likely to be kept up to date. The changes to `archivePath` seem unnecessary.

What is the benefit of declaring class properties?

Eitot:
It is just syntactic sugar from what I understand. In Swift, you can drop the brackets and in Objective-C, you can use dot notation. It is simply for consistency with Cocoa APIs, e.g. [NSFileManager](https://developer.apple.com/documentation/foundation/nsfilemanager/1409234-defaultmanager?language=objc), as well as with instance properties.

I made the changes to `archivePath` to avoid the nullability warning in Xcode. The alternative would have been a forced cast to `(NSString *_Nonnull)` to suppress the warning.

PR #1115, opened 2017-08-07
Add optional updaterDidRelaunchApplication: method on SUUpdaterDelegate

AppTyrant:
Added optional updaterDidRelaunchApplication: method on SUUpdaterDelegate


pornel:
Thank you

PR #1117, opened 2017-08-10
Added a timer invalidation in a asynchronous callback.

drewmccormack:
The scheduleNextUpdateCheckFiringImmediately: method includes an asynchronous callback. It invalidates the timer before invoking the method with the callback (retrieveNextUpdateCheckInterval:), but doesn’t invalidate again inside the callback block.

The problem with this is that it is possible the timer is started during the delay. In release versions of an app, crashes have been seen due to an assertion that the NSTimer be nil when scheduling.

The proposed fix is to invalidate the timer in the callback, ensuring the NSTimer is nil before scheduling.


pornel:
Thank you for the pull request. That's a good catch.

PR #1120, opened 2017-08-16
Fix for Skip This Updates button not working in XPC branch

ksuther:
includesSkippedUpdates was set to YES in SPUScheduledUpdateDriver
This was causing the Skip This Version button to not work when a background update check ran.
This was introduced in 47f08be1a.


pornel:
Thank you

PR #1123, opened 2017-09-06
Fixed progress bar not updating while downloading

willco007:
This file still uses NSURLDownload callbacks:

- (void)download:(NSURLDownload *)__unused download didReceiveResponse:(NSURLResponse *)response
and 
- (void)download:(NSURLDownload *)__unused download didReceiveDataOfLength:(NSUInteger)length

But the actual download is now handled by NSURLSession so the progress bar is never updated.  Had to implement session callback.

- (void)URLSession:(NSURLSession *)__unused session downloadTask:(NSURLSessionDownloadTask *)__unused downloadTask didWriteData:(int64_t)__unused bytesWritten totalBytesWritten:(int64_t)totalBytesWritten totalBytesExpectedToWrite:(int64_t)totalBytesExpectedToWrite

### DIFF
--- a/Sparkle/SUUIBasedUpdateDriver.m
+++ b/Sparkle/SUUIBasedUpdateDriver.m
@@ -169,11 +169,17 @@
     [super downloadUpdate];
 }
 
-- (void)download:(NSURLDownload *)__unused download didReceiveResponse:(NSURLResponse *)response
+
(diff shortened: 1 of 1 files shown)


willco007:
NM, this is our fork only. Sorry for the troubles.

PR #1124, opened 2017-09-11
Restore Xcode 7 and macOS 10.11 SDK compatibility

### LINKED ISSUE #1122: Sparkle 1.18.1 cannot be compiled with macOS SDK smaller than 10.12
Due to recently added support of os_log in SULog.m, Sparkle is now able to log to the 10.12 native logging framework.

Those new API was introduced into the newest 10.12 SDK only. While the file has runtime fallbacks to older mechanisms, the included header "<os/log.h>" is unguarded.
This means that sparkle cannot be compiled with any older SDK than 10.12.

This should be fixed by guarding the header to allow compilation with 10.11. Alternatively, documentation should be updated if sparkle needs to be compiled with 10.12 now (which would be very sad).

dfuhrmann:
This pull request resolves #1122, by restoring 10.11 compatibility for the current sparkle code base.
* SULog was modified to only use the new os_log APIs if compiled with a 10.12 SDK.
* Some forward declarations for touch bar were modified to compile with 10.11 SDK
* all xib files are changed to Xcode 7 format, as only this format can be read and compiled with Xcode 7

### DIFF
--- a/Sparkle/SULog.m
+++ b/Sparkle/SULog.m
@@ -7,15 +7,29 @@
 //
 
 #include "SULog.h"
+
 #include <asl.h>
+#include <Availability.h>
+
+#if __MAC_OS_X_VERSION_MAX_ALLOWED >= 101200
 #include <os/log.h>
+#else
+
+typedef struct os_log_s *os_log_t;
+#define os_log_create(subsystem, category) nil
+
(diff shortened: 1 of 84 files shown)


pornel:
That's great. Thank you.

PR #1128, opened 2017-09-19
Merged master into ui-separation-and-xpc

ksuther:
Merged the latest commits into ui-separation-and-xpc. There were some merge conflicts because of all the xibs being updated for Xcode 7 compatibility.


pornel:
Thank you

Issue #835, opened 2016-07-06
Sparkle clients polling frequently


herf:
We use Sparkle to update f.lux (thanks!) Our appcast is hosted at https://justgetflux.com/mac/macflux.xml - only SSL.

A minority of clients (perhaps one in 100,000) seem to poll in a loop. So everyday we have a bunch of IPs we have to drop at the firewall, because otherwise they send 10 requests/sec all the time. (Our nginx server also has all the abuse controls dialed now.)

We are using Sparkle 1.14.0, and the issue happened on 0.6.1 also (upgraded before reporting). Latest versions of Mac OS affected (including 10.11.5 and 10.12 in my logs today). This issue has appeared in the last year or two.

Is there anything that would allow Sparkle to poll (or retry) at a rate like this?

Normally this could be spammers, but in a few cases, the logs suggest these are normal users (who've been checking for updates once a day for a long time), and all of a sudden they start the rapid-polling behavior.

zorgiepoo:
Hey,

The main thing I can think of is our background scheduling code (in SUUpdater.m):

```objc
    // How long has it been since last we checked for an update?
    NSDate *lastCheckDate = [self lastUpdateCheckDate];
	if (!lastCheckDate) { lastCheckDate = [NSDate distantPast]; }
    NSTimeInterval intervalSinceCheck = [[NSDate date] timeIntervalSinceDate:lastCheckDate];

    // Now we want to figure out how long until we check again.
    NSTimeInterval delayUntilCheck, updateCheckInterval = [self updateCheckInterval];
    if (updateCheckInterval < SUMinimumUpdateCheckInterval)
        updateCheckInterval = SUMinimumUpdateCheckInterval;
    if (intervalSinceCheck < updateCheckInterval)
        delayUntilCheck = (updateCheckInterval - intervalSinceCheck); // It hasn't been long enough.
    else
        delayUntilCheck = 0; // We're overdue! Run one now.
    self.checkTimer = [NSTimer scheduledTimerWithTimeInterval:delayUntilCheck target:self selector:@selector(checkForUpdatesInBackground) userInfo:nil repeats:NO];
```

But it's not quite obvious to me how it can come from here. The update check interval is constrained by `SUMinimumUpdateCheckInterval` which means the info.plist key or user default read there can't be the reason. The last update check date is read from the user defaults and set before poking at the appcast.

[...]

PR #1135, opened 2017-10-05
Fix #1061 & #835: Sparkle clients polling too frequently

### LINKED ISSUE #1061: Clients polling frequently
Bug #835 is still occurring.

We rolled a f.lux update with Sparkle 1.16.0 and the frequent polling is still happening. 
(Our script just blocked a client polling 13 times a second.)

jkbullard:
The SULastCheckTimeKey preference (a/k/a "default"), which is stored
in ~/Library/Preferences/….plist, was used to store the date of the
last update check.

That causes a problem if the storage is read-only (for example, if the
.plist file is “locked” (has the uchg or schg flag set) or if the
SULastCheckTimeKey key "exists in a domain that precedes the
application domain the search list"). The date would never be modified,
and the old date would cause a new update check to be queued at the end
of an update check. If application or bundle was up-to-date, or the
user had chosen to skip the update that was available, this would cause
update checks to happen as often as the computer and network would
allow, many times each second.

The new code uses an instance variable to store the value. The variable
is initially loaded from the “SULastCheckTimeKey” preference, and each
time the variable is modified its new value is stored to that
preference. However, the variable, not the preference, is used when
calculating whether enough time has passed since the last update check
to warrant another check. The variable will always (after the first
update check) have the "correct" value, and update checks will not be
done more frequently than desired.

### DIFF
--- a/Sparkle/SUUpdater.m
+++ b/Sparkle/SUUpdater.m
@@ -266,9 +266,8 @@ static NSString *const SUUpdaterDefaultsObservationContext = @"SUUpdaterDefaults
 
 - (void)updateLastUpdateCheckDate
 {
-    [self setUpdateLastCheckedDate:[NSDate date]];
-    
     [self willChangeValueForKey:@"lastUpdateCheckDate"];
+    [self setUpdateLastCheckedDate:[NSDate date]];
     [self.host setObject:[self updateLastCheckedDate] forUserDefaultsKey:SULastCheckTimeKey];
     [self didChangeValueForKey:@"lastUpdateCheckDate"];
 }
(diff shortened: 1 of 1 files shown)


Sparkle/SUUpdater.m
```diff
@@ -253,13 +256,20 @@ - (void)updateDriverDidFinish:(NSNotification *)note
...
+    {
+        [self setUpdateLastCheckedDate:[self.host objectForUserDefaultsKey:SULastCheckTimeKey]];
+    }
+    
+    return [self updateLastCheckedDate];
 }
 
 - (void)updateLastUpdateCheckDate
 {
+    [self setUpdateLastCheckedDate:[NSDate date]];
```
pornel: I think this should be after willChangeValueForKey, so that before/after values in observers work

pornel:
Thank you!

PR #1141, opened 2017-10-21
Update Sparkle.strings

stefanhp:
Translation of "Description text for SUUpdateAlert when the update informational with no download."


pornel:
Thank you

PR #1154, opened 2017-11-25
Update Japanese localization

1024jp:
Apple tweaked the rule of Japanese localization on macOS Sierra.
This update conforms to the modern rule.


pornel:
Thank you

PR #1160, opened 2017-12-07
Shell script fixes

ksuther:
Fixed an error in set-agent-signing.sh
Don't call codesign in set-agent-signing if the identity is "-" (this means no signing identity is set)
Fixed some warnings flagged by shellcheck


pornel:
Did you mean to check for "-" only in one of the two cases?

ksuther:
I wasn't checking the Sparkle Test App Info.plist because it seems to get signed by default, but I added a similar conditional there.

pornel:
Thank you

PR #1161, opened 2017-12-08
Run codesign when signing identity is a hyphen

ksuther:
I'm sorry, my last shell script change wasn't totally correct (#1160). It was building in my main projects but then failed when building the UI test targets.

It turns out that using a hyphen (-) to sign creates an adhoc signature rather than not signing (this is new to me, oops!). The last fix for the incorrect conditional in the shell script is still correct, but the extra check for "-" needs to be removed.


pornel:
Thanks for the correction.

PR #1165, opened 2017-12-20
[WIP] Port SPUDownloader to 1.x w/NSURLSession & NSURLDownload

### LINKED ISSUE #1131: Replace NSURLDownload use
[NSURLDownload is now printing out a console warning on 10.13.](https://stackoverflow.com/questions/46088843/macos-10-13-scheduling-the-nsurldownload-loader-is-no-longer-supported)

Still works, but the console spam is annoying.

Edit: Building from 1.18.1 source using Xcode 9/10.13 SDK. I checked head and NSURLDownload is still used.

```
(lldb) b [NSURLDownload init]
Breakpoint 1: where = CFNetwork`-[NSURLDownload init], address = 0x0000000000238543
(lldb) r
...
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = breakpoint 1.1
    frame #0: 0x00007fff3536e543 CFNetwork`-[NSURLDownload init]
    frame #1: 0x00007fff3536e5f1 CFNetwork`-[NSURLDownload initWithRequest:delegate:] + 48
    frame #2: 0x0000000107f67ad6 Sparkle`-[SUAppcast fetchAppcastFromURL:inBackground:completionBlock:] + 799
    frame #3: 0x0000000107f6cf46 Sparkle`-[SUBasicUpdateDriver checkForUpdatesAtURL:host:] + 594
```

Deadpikle:
Heyo, this PR attempts to port `SPUDownloader` from the `ui-separation-and-xpc` branch to `master` in order to fix #1131. I will note this as WIP because I'm not sure of best practice for handling all of the `XYZ is only available on 10.X` warnings (see especially `SPUURLRequest.m`), and I'm also guessing there's a better way to implement the "abstract" `SPUDownloader`. Do we need some sort of `#if`/`#else` for not compiling `NSURLSession` stuff on older versions?

I've done _minimal_ testing myself, and release notes/downloads seem to work OK on my machine running 10.12.

Please let me know what I can do to improve this PR for potential merging into `master` if that's something project maintainers want to see! Thanks.

### DIFF
--- a/Sparkle/SPUDownloaderDeprecated.m
+++ b/Sparkle/SPUDownloaderDeprecated.m
@@ -60,36 +60,19 @@
 
 - (void)download:(NSURLDownload *)__unused d decideDestinationWithSuggestedFilename:(NSString *)name
 {
-    if (self.mode == SPUDownloadModeTemporary)
-    {
+    if (self.mode == SPUDownloadModeTemporary) {
         // Files downloaded in temporary mode should not last for very long,
         // so it's ideal to place them in a system temporary directory
         NSString *destinationFilename = NSTemporaryDirectory();
-        if (destinationFilename)
-        {
+        if (destinationFilename) {
             destinationFilename = [destinationFilename stringByAppendingPathComponent:name];
             
             [self.download setDestination:destinationFilename allowOverwrite:NO];
         }
-    }
-    else
-    {
-        // Remove our old caches path so we don't start accumulating files in there
-        NSString *rootPersistentDownloadCachePath = [[SPULocalCacheDirectory cachePathForBundleIdentifier:self.bundleIdentifier] stringByAppendingPathComponent:@"PersistentDownloads"];
+    } else {
+        NSString *tempDir = [super getAndCleanTempDirectory];
         
-        [SPULocalCacheDirectory removeOldItemsInDirectory:rootPersistentDownloadCachePath];
-        
-        NSString *tempDir = [SPULocalCacheDirectory createUniqueDirectoryInDirectory:rootPersistentDownloadCachePath];
-        if (tempDir == nil)
-        {
-            // Okay, something's really broken with this user's file structure.
-            [self.download cancel];
-            self.download = nil;
-            
-            NSError *error = [NSError errorWithDomain:SUSparkleErrorDomain code:SUTemporaryDirectoryError userInfo:@{ NSLocalizedDescriptionKey: [NSString stringWithFormat:@"Can't make a temporary directory for the update download at %@.", tempDir] }];
-            
-            [self.delegate downloaderDidFailWithError:error];
-        } else {
+        if (tempDir != nil) {
             NSString *downloadFileName = self.desiredFilename;
             NSString *downloadFileNameDirectory = [tempDir stringByAppendingPathComponent:downloadFileName];
             
--- a/Sparkle/SPUDownloaderDeprecated.m
+++ b/Sparkle/SPUDownloaderDeprecated.m
@@ -174,4 +174,9 @@
     [super cleanup];
 }
 
+- (void)cancel
+{
+    [self cleanup];
+}
+
 @end
--- a/Sparkle/SPUDownloaderDeprecated.m
+++ b/Sparkle/SPUDownloaderDeprecated.m
@@ -13,6 +13,8 @@
 #import "SPULocalCacheDirectory.h"
 #import "SUErrors.h"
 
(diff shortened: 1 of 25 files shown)


Sparkle/SPUDownloaderDeprecated.m
```diff
@@ -0,0 +1,184 @@
...
+        if (destinationFilename)
+        {
+            destinationFilename = [destinationFilename stringByAppendingPathComponent:name];
+            
+            [self.download setDestination:destinationFilename allowOverwrite:NO];
+        }
+    }
+    else
+    {
+        // Remove our old caches path so we don't start accumulating files in there
```
pornel: The cache cleanup code is duplicated in both classes. Would you be able to move it to a method in the base class?

Sparkle/SPUDownloaderSession.m
```diff
@@ -0,0 +1,181 @@
...
+    
+    if (self.mode == SPUDownloadModePersistent && bytesWritten >= 0) {
+        [self.delegate downloaderDidReceiveDataOfLength:(uint64_t)bytesWritten];
+    }
+}
+
+- (void)URLSession:(NSURLSession *)__unused session task:(NSURLSessionTask *)__unused task didCompleteWithError:(NSError *)error
+{
+    self.download = nil;
+    if (error) {
```
pornel: Why is that? Is this method called without an `error`?

I'm worried that if download fails without notifying the delegate it could make the delegate stuck waiting forever.

pornel:
Thanks for the PR. The approach is good.

It would be good to have tests, especially for both types of downloaders, so that we don't accidentally break the less used one later.

I could upload some files to sparkle-project.org, and you could download them and check their contents in a unit test (using each downloader subclass). What do you think?

pornel:
(oops, wrong button, I didn't mean to close it)

Sparkle/SPUDownloaderSession.m
Deadpikle: Yeah, the method is called without an error. See [documentation](https://developer.apple.com/documentation/foundation/nsurlsessiontaskdelegate/1411610-urlsession). `If an error occurred, an error object indicating how the transfer failed, otherwise NULL.` My guess is that this is _always_ called for `NSURLSession`, since a session can be called with multiple tasks, and `URLSession:downloadTask:didFinishDownloadingToURL` is also called so you can track tasks as they complete. A better solution may be to refactor the logic from the task finishing to the session completing so that there aren't any weirdo race conditions or anything. Is there any case where more than one task would be running at once?

IIRC from my testing, sending the `nil` error on to the delegate caused errors (which, granted, should likely be fixed). If I recall, in the `ui-separation-and-xpc` branch, `delegate` was `nil` (race condition?) by the time this was called. I don't have the time this instant to retest things, so I'll take a closer look and report back "soon."

Deadpikle:
I refactoring the tmp directory clearing along with the error reporting.

I'd be more than happy to write up some unit tests. That's generally always a good idea 😁 Just let me know when you've got some files up!

pornel:
https://sparkle-project.org/unit_test/test.xml contains `"<test>appcast</test>\n"`
https://sparkle-project.org/unit_test/icon.png 133495 bytes large. You could check checksum of it (like sha1), or maybe just check file size and load it as NSImage and check whether it worked.

Deadpikle:
Hey, I've stuck in some unit tests. I opted for the hash-the-file method for checking download verification.

Re: Setting `NSOperationQueue` operation count to 1: in working on the unit tests for the `NSURLSession` downloader, I was often getting the `didCompleteWithError` _before_ `didFinishDownloadingToURL`, which seemed odd. [This post on Apple's dev forums](https://forums.developer.apple.com/thread/47252) seemed to indicate some fishiness with delegate callback order, and, indeed, setting the queue's operation count to 1 makes the callbacks called in the order of `didFinishDownloadingToURL` -> `didCompleteWithError`. I'm not sure this is the "real" solution, but it might clear up the `if (error) { ... }` delegate callback stuff. I need to research a bit more. (It's the holidays, so that might take a bit of time.)

pornel:
Thanks. 

Once thing I've noticed is that you're using custom/non-main operation queue for the download now. Does that mean the delegate callbacks are called from non-main thread?  If so, then `downloaderDidReceiveDataOfLength` delegate methods need to ensure UI is updated on the main thread.

pornel:
The tests are good. Thank you very much for implementing this.

I've added main thread changes in #1167

Issue #860, opened 2016-08-10
Doesn't build for distribution if cocoapods not installed


sanfords:
The offending line of make-release-package.js:

`spec_version=$(printf "require 'cocoapods'\nspec = %s\nprint spec.version" "$(cat "$SRCROOT/Sparkle.podspec")" | LANG=en_US.UTF-8 ruby)`

Thank you!

zorgiepoo:
@jakepetroules is it possible to print a warning if cocoa pods is not installed or some such instead of failing? `Gem::Specification::find_all_by_name('cocoapods').any?` may work as a test (from this [SO thread](http://stackoverflow.com/questions/1032114/check-for-ruby-gem-availability)).

kolpanic:
I'm trying to do a release build of the `ui-separation-and-xpc` branch, and I'm seeing this too. The culprit is "sanity check" [here](/sparkle-project/Sparkle/blob/master/Configurations/make-release-package.sh#L5).

kornelski:
`which pod` may also work on Bash level. 

Does anyone want to make a pull request for this?

kolpanic:
@kornelski According to [this answer](https://stackoverflow.com/questions/592620/check-if-a-program-exists-from-a-bash-script#677212), `which` should be avoided for this kind of thing. I created #1170 using the preferred approach mentioned there.

kornelski:
Thanks for the PR.

kolpanic:
@kornelski My pleasure. Thanks for you work.

BTW, my PR was to` ui-separation-and-xpc`. I don't know how you want to get this fix onto `master`—possibly by cherry-picking?

PR #1170, opened 2018-01-04
Check if cocoapods is installed

kolpanic:
Ensure cocoapods is installed before performing the release “sanity check” (#860)

Based on https://stackoverflow.com/questions/592620/check-if-a-program-exists-from-a-bash-script#677212


kornelski:
Thank you

PR #1181, opened 2018-02-02
Amend to avoid gender issue

vitu:
Amend to avoid gender issue

Instead of addressing the user, which might cause gender issues, string has been adapted to include and refer to “app” (masculine, in accordance with Apple's terminology).

Would translate back as “The app is up-to-date!”


kornelski:
Thank you. That is appreciated!

PR #1183, opened 2018-02-15
Fixing build error when some enclosing folder contains a space.

charlieMonroe:
I personally have my projects on iCloud Drive which contains a space in the path (`~/Library/Mobile Documents/...`) and the build fails running these two scripts.


kornelski:
Oh, bash. 

Thank you for the PR.

PR #1184, opened 2018-02-16
Fixed SPUDownloader leak

ksuther:
SPUDownloader was getting leaked because its NSURLSession wasn't ever invalidated. If NSURLSession isn't invalidated then its delegate keeps a strong reference.

Looks like the same fix will also need to be applied to master.


kornelski:
Nice find. Thanks.

PR #1189, opened 2018-03-09
Fix crash when abortUpdateWithError is called on the main queue

patters:


Deadpikle:
Correct me if I'm wrong, but couldn't you do the same and reduce complexity by using `[NSThread isMainThread]`?

kornelski:
Thanks for the PR. I've applied a version of that with `isMainThread`.

patters:
Thanks @kornelski !

@Deadpikle I didn't use `isMainThread` because I've read that only works for processes that are calling into `NSApplicationMain` (otherwise there can be some odd discrepancies, where running on the main thread does not necessarily mean the main queue, and vice versa).

PR #1192, opened 2018-03-15
Improve 'read-only' error message

adrianthomas:
Heya,

By far the most common cases we see for this error being shown are 
- running the app from ~/Downloads 
- running from a .dmg. 

I guess most apps no longer use physical media, so I'd like to update the error message to reflect that.

👋
Adrian

equinux


kornelski:
Thanks for the PR. We meant to change it before, but I changed localization key rather than the value :)

PR #1193, opened 2018-03-15
Patch 2

adrianthomas:
Heya,

By far the most common cases we see for this error being shown are 
- running the app from ~/Downloads 
- running from a .dmg. 

I guess most apps no longer use physical media, so I'd like to update the error message to reflect that.

This is the branch we're using, so it'd be great to get the updated strings here as well 👍 

👋
Adrian

equinux


kornelski:
Dupe of #1192

PR #1212, opened 2018-04-16
expose CodesigningVerifier, add codesign info api

sunuslee:
1. export CodesigningVerifier & CodesigningVerifier.h to Framework's Header
2. add api (NSDictionary *)codeSignatureInfoAtBundleURL:(NSURL *)bundlePath to main app to use for more **customized validate work** (teamid, etc), not just check if codesign in validate.
3. this change is base on **+ (void)logSigningInfoForCode:(SecStaticCodeRef)code label:(NSString*)label**.

Signed-off-by: sunuslee <sunuslee@gmail.com>


kornelski:
What is the use case for this?

kornelski:
In what cases does the team ID change? 

We're trying to check that the developer ID of the old and new bundle is the same. Is that not enough?

sunuslee:
i mean, we want to get more control & detail on codesign info. not just team id.

sunuslee:
@kornelski hello, i've send you email for the detail about this pull request:0

sunuslee:
@kornelski hello, any updates? :)

kornelski:
Use of Sparkle as a basis for an installer for new apps is very interesting, and for that purpose I see it does make sense to expose the code signing.

sunuslee:
@kornelski yeah! i am too busy right now, i will get back to u as soon as i got some free time!!

sorry for the delay reply:>

PR #1216, opened 2018-04-20
Fix closeCheckingWindow called from background thread

mallexxx:
Currently closeCheckingWindow is called from background thread (caused by [SPUDownloaderSession URLSession:downloadTask:didFinishDownloadingToURL] session delegate method called in background). Although it's surprisingly working well in most cases, sometimes it crashes.


kornelski:
Thank you

PR #1230, opened 2018-05-13
Support for getting the private key from keychain for signing deltas

tamasprezi:
It's not safe to have private keys sitting on the disk in plaintext files, but so far this has been a requirement* for `generate_appcast`. This PR adds support for reading the private key directly from a keychain by extending the tool's CLI.

To load the key from file (old):

```
$ generate_appcast private.pem all_my_dmgs/
```

To load the key from file (new):

```
$ generate_appcast -f private.pem all_my_dmgs/
```

To load the key from keychain (new):

```
$ generate_appcast -n ~/Library/Keychains/login.keychain -n "My Sparkle Update Private Key" all_my_dmgs/
```

<sub>* Note: there is a workaround which involves extracting the previously stored key from the keychain, writing it to file, and passing that path to `generate_appcast`, but IMO that's an unnecessary roundtrip. </sub>

### DIFF
--- a/generate_appcast/DSASignature.swift
+++ b/generate_appcast/DSASignature.swift
@@ -25,6 +25,30 @@ func loadPrivateKey(at privateKeyURL: URL) throws -> SecKey {
     return (cfitems! as NSArray)[0] as! SecKey;
 }
 
+func loadPrivateKey(named keyName: String, fromKeychainAt keychainURL: URL) throws -> SecKey {
+    var keychain: SecKeychain? = nil
+    
+    guard SecKeychainOpen(keychainURL.path, &keychain) == errSecSuccess, keychain != nil else {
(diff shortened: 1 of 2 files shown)


kornelski:
Keychain support is sweet! Thank you.

However, breaking existing usage is not so good. Could you keep old behavior when given 2 non-option args?

Can you auto-detect default login keychain path, so that it's not necessary to always specify it?

tamasprezi:
Keeping old behavior: sure thing 👍 
Default keychain: I don't see any API on the Security.framework for discovering the user's keychains (the functionality provided by `security list-keychains`), so I'd rather not hardcode anything there.

Wevah:
Doesn't SecItemCopyMatching search the user's default keychains by default, or am I misremembering?

tamasprezi:
@Wevah good point, I'll check

PR #1234, opened 2018-05-22
delegate callbacks for fininshed download and extraction related events

horcs:


kornelski:
What's the use case for it?

horcs:
In my use case when the app is updating I need to show to the app user the current "state". Wether the download did finish successfully and is it extracting.
When the user otp-outs for background update the Sparkle UI is not indicating this info but we want to show it on certain places in the app. Also this makes diagnostic logging cleaner and easier.

PR #1238, opened 2018-05-28
Download process string updated to match master

aonez:
Just updated the download progress string method to match the master fines progress.


kornelski:
Thanks

PR #1244, opened 2018-06-04
fix warnings when using modules

nivekkagicom:
correct warning when including Foundation and modules are enabled


RobFog:
Looks like a rebase would be in order…? ;-)

PR #1249, opened 2018-06-14
Updated Finnish language resources

JasonLPollack:
There was an older version of Finnish resources that did not include some inline xib localizations and was missing some strings from the .strings file. This adds those missing resources.


kornelski:
Thank you

PR #1250, opened 2018-06-14
Silence Xcode 10 "Messaging unqualified id" warning

ksuther:
Silence "Messaging unqualified id" warning that's new in Xcode 10 (-Wobjc-messaging-id). This warning is added only when compiling against the 10.14 SDK.

This change should also be applied to the ui-separation-and-xpc branch.

### DIFF
--- a/Configurations/ConfigCommon.xcconfig
+++ b/Configurations/ConfigCommon.xcconfig
@@ -87,5 +87,8 @@ CLANG_ENABLE_MODULES[sdk=macosx10.8] = NO
 // These should be removed once the conversion to ARC is complete
 WARNING_CFLAGS_EXTRA = -Wno-custom-atomic-properties -Wno-implicit-atomic-properties
 
+// Turn off -Wobjc-messaging-id on Xcode 10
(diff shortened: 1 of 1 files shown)


kornelski:
Is it possible to fix the code it's warning about, instead of disabling the warning?

ksuther:
That would involve casting every place where Sparkle calls a method on an id object, which happens in quite a few places (mostly where objectForKey is called). We could do this, although the correct way to fix most of the warnings would be to use generics on collections which would require increasing the minimum required version of Xcode.

kornelski:
I'm happy to bump minimum required Xcode version. We have hard constraint on supporting old macOS releases, but we can easily require developers to use newer Xcode.

ksuther:
I'll take a look and see how far I can get with that approach. This would mean requiring Xcode 7/10.11 to compile, which at this point is more than reasonable.

ksuther:
See #1251. It ended up not being too bad, and it looks like we're requiring Xcode 7 to build anyway.

PR #1251, opened 2018-06-14
Silence Xcode 10 "Messaging unqualified id" warnings by specifying types

ksuther:
Alternate fix for #1250. This explicitly casts the return types where the warnings are to silence them rather than disabling the warning.

If this approach is better I'll do the same thing in the ui-separation-and-xpc branch afterwards.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -30,7 +30,7 @@ Secure and reliable software update framework for Cocoa developers.
 ## Requirements
 
 * Runtime: macOS 10.7 or greater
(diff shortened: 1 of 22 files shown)


kornelski:
Thanks! That looks good. 

I'm disappointed that Xcode warns about `[self alloc]`. Seems like a big oversight in the warning implementation.

ksuther:
Yeah, that was my least favorite part. You'd think the compiler would be able to silence that since instancetype exists. Fortunately it only occurs a few times.

PR #1257, opened 2018-06-20
Fix compilation issues on Xcode 10 with new build system

LeoNatan:
During WWDC I found some compilation issues with Xcode 10 under the new build system. I sat with an Apple engineer in the labs, and we found that this change was necessary because the scripts access the Info.plist file, which in the new parallel build system may not be ready. These changes ensure the dependency system is aware and times tasks in the correct order.


kornelski:
That's cool. Thank you.

LeoNatan:
Ooh nice, thank you for the fast response!

Issue #1259, opened 2018-06-22
Issues with Spanish localisation


kenohori:
Hi,

There are a few issues with the Spanish localisation:
- Uses formal version of "you" (usted) and its related verb conjugations, contrary to the informal tú that is used throughout macOS.
- Weird abbreviations for B, KB, MB and GB. Not strictly incorrect, but not really the preferred form since the 90s.
- Sometimes doesn't say the same as the English counterpart. Extra text in some places, missing text in others.
- Key words that would make things clearer are missing here and there (e.g. currently).
- Inconsistent words for the same text in different strings.
- The App name is always referred to using the masculine form. It would be better to use always feminine since application (aplicación) is feminine.

Mind if I rewrite the Spanish localisation strings and submit a pull request?

kornelski:
Yes, please!

Our goals for localization are:

 - use the same terminology and phrases as Apple in macOS, where possible.
 - use gender-neutral forms when referring to the application user.

kornelski:
This has been released

PR #1261, opened 2018-06-28
New Spanish localisation

kenohori:
As discussed in #1259, I created new strings for the Spanish localisation.

I had to make some decisions. Hope these are okay with you:
- I refer to the app name and update using feminine pronouns. Makes most sense since application (aplicación), version (versión) and update (actualización) are all feminine nouns in Spanish.
- There's no ideal translation for relaunch, so I put it as "volver a abrir" (reopen).
- The most precise translations for feed (fuente) and parse (analizar sintácticamente) are technical terms that aren't used in common speech, so I used "información" (information) and analizar (analyse) instead.


kornelski:
Thank you!

nbpalomino:
I'm trying to use spanish localization, but I fail, where should be configured? Thanks

kornelski:
It's picked up automatically by macOS if the system is in Spanish. You don't need to do anything.

PR #1268, opened 2018-07-17
Fix binary delta create 1156

### LINKED ISSUE #1156: BinaryDelta create fails on network mounted NAS Device due to sym link permissions
Symbolic links created on a drive mounted (AFP) from a NAS device are always created with 777 permissions meaning that "BinaryDelta create" will always reject them and fail.  Since the real files and directories have the correct 755 or 644 permissions it seems to me the symlinks should be ignored as they will effectively take on those permissions.  If we do not change BinaryDelta so it ignores symbolic link permissions it is unusable for network mounted directories.

sagecook:
Closes  #1156


Sparkle/SUBinaryDeltaCreate.m
```diff
@@ -574,13 +586,28 @@ BOOL createBinaryDelta(NSString *source, NSString *destination, NSString *patchF
...
     xar_close(x);
 
-    unlink([patchFile fileSystemRepresentation]);
-    link([temporaryFile fileSystemRepresentation], [patchFile fileSystemRepresentation]);
-    unlink([temporaryFile fileSystemRepresentation]);
-
+    NSFileManager *filemgr;
+    filemgr = [NSFileManager defaultManager];
+    
+    [filemgr removeItemAtPath: patchFile error: NULL];
```
kornelski: This could be simplified to use `moveItemAtPath`. Also some stray NSLogs are left

Sparkle/blame.out
```diff
@@ -0,0 +1,586 @@
+08434edf7 (Zorg            2015-04-09 21:39:11 -0400   1) //
```
kornelski: Please remove this file

kornelski:
Exclusion of symlinks from permission checks seems sensible.

The replacement of file via hardlink was odd indeed.

Sparkle/blame.out
sagecook: Quite right.

Sparkle/SUBinaryDeltaCreate.m
sagecook: Will do.

sagecook:
Just committed/pushed requested changes with 97985747.

kornelski:
Great! Thanks

PR #1270, opened 2018-07-18
Fixed some typos

shayr1:
Fixed typos in the following keys:
"An error occurred during installation. Please try again later."
"An error occurred in retrieving update information. Please try again later."
Added missing key that fell back to English:
"Cancel Update" = "בטל עדכון"; (Inside alert when using a button to let user check for updates)


kornelski:
Thank you

PR #1273, opened 2018-07-25
Adds an option to turn off hideOnDeactivate in SPUStandardUserDriver for foreground apps

rob-patchett:
By default, SPUStandardUserDriver's update window disappears when moving focus to another app. Depending on your application, this might not be the desired behavior. This PR adds a boolean property that can be set to alter this behavior. Note that it still respects background apps, where the property has no effect.

Thank you for the time to review and please let me know if there might be a more appropriate way to address this issue.


kornelski:
Thanks

Issue #1277, opened 2018-08-19
Sparkle Test App crashing with assert(signUpdateTask.terminationStatus == 0)


tempelmann:
I just downloaded v1.20.0, built the Release scheme with Xcode 9.2 on macOS 10.12.6, and then launched the "Sparkle Test App.app" - that crashes with said assertion.
Same when I run the "Sparkle Test App" scheme from Xcode (in debugger).

Is that expected to be happening? If so, it would be nice if the documentation would mention that fact, and what has to be done to make this test app work properly.

tempelmann:
Ah, here's the output from the NSTask invocation:

> Error Signing Data
167:error:0606B06E:digital envelope routines:EVP_SignFinal:wrong public key type:/BuildRoot/Library/Caches/com.apple.xbs/Sources/OpenSSL098/OpenSSL098-64.50.7/src/crypto/evp/p_sign.c:99:

So it seems that the included .pem files are not valid.

tempelmann:
Okay, this is really strange. The current "sign_update" script uses this cmd line:

`$openssl dgst -sha1 -binary < "$1" | $openssl dgst -sha1 -sign "$2" | $openssl enc -base64`

I compared that to my older working signing script, and that uses "-dss1" instead of "-sha1" in the 2nd cmd:

`$openssl dgst -sha1 -binary < "$1" | $openssl dgst -dss1 -sign "$2" | $openssl enc -base64`

And guess what - that version works as expected.

So, what's up with that change, make in this commit: https://github.com/sparkle-project/Sparkle/commit/0cb0b2f2c7b07b65231528991ac98dc384e51146

It appears to break signing on 10.12.6.

tempelmann:
I've submitted a fix that checks the openssl version and uses the old method on pre-10.13 systems:

https://github.com/sparkle-project/Sparkle/pull/1278

PR #1278, opened 2018-08-19
Make "sign_update" work on pre-10.13 system

tempelmann:
Make "sign_update" work on pre-10.13 system by checking the OpenSSL version, using the older "-dss1" digest option again in that case. See https://github.com/sparkle-project/Sparkle/issues/1277 and https://github.com/sparkle-project/Sparkle/commit/0cb0b2f2c7b07b65231528991ac98dc384e51146


kornelski:
Thanks for the fix

PR #1280, opened 2018-08-26
Showing release notes with dark appearance according to the system theme

michelf:
This is done by applying a user stylesheet to get white text, making the background of the WebView transparent, and using a `NSBox` to display the dynamic dark background color.

Notes:

- All done in code so there is no need to change the IB files.
- Before this change, the spinner would appear invisible (white on white) while loading the release notes.
- The appearance of the release notes will unfortunately not adapt if the theme is changed while the window is open. That would probably involve adding a view in the hierarchy to handle calls to `effectiveAppearanceDidChange` by calling the method `adaptReleaseNotesAppearance` on the window controller.
- HTML files using custom colors might not appear at their greatest with this new background. Release note for some apps might need to be tweaked accordingly.

## Before

capture d ecran 2018-08-26 a 06 27 55

## After

capture d ecran 2018-08-26 a 15 56 00

### DIFF
--- a/Resources/DarkAqua.css
+++ b/Resources/DarkAqua.css
@@ -1,0 +1,10 @@
+html {
+    color: white;
+    background: transparent;
+}
+:link {
+    color: #419CFF;
+}
+:link:active {
+    color: #FF1919;
+}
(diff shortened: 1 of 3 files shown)


michelf:
That's why there's two layers:

* the `#if` is to enable this when compiling with the 10.14 SDK
* the `if (@available(macOS 10.14, *)` is the runtime check.

```
#if __MAC_OS_X_VERSION_MAX_ALLOWED >= 101400
     if (@available(macOS 10.14, *))
```

kornelski:
Oh, sorry. I've missed that line.

Issue #1275, opened 2018-08-08
French strings truncated for permission xib


jballance:
There is some text that is getting cut off in the French version of the panel that asks permission to automatically check for updates. In the ProfileInfo view, the button by the check box is too short. In the MoreInfoView, the last word in the line above the scroll view is cut off.

PR #1281, opened 2018-08-26
Fixing cut text in prompt window (French)

michelf:
Made the checkbox wider in French since there's plenty of space.

Shortened the text a bit in MoreInfoView so it does not get cut when the view is squeezed to fit the prompt window. A rough literal translation of the change is like this: "Below is shown the information ~~that will be~~ transmitted:".

This problem was reported in #1275.


kornelski:
Thank you

PR #1288, opened 2018-09-15
Dark Mode Release Notes Complement

michelf:
Continuing from #1280:

* Now reacting to system appearance changes using KVO on the `effectiveAppearance` property of the release notes view.
* Better future-proofing for dark mode detection using `bestMatchFromAppearancesWithNames` instead of matching current appearance by name.
* Fixing issue where `darkBackgroundView` wouldn't be recreated when entering dark mode because it wasn't set to nil before.


kornelski:
Thanks

PR #1293, opened 2018-10-08
Only observe effectiveAppearance on 10.14 or later

ksuther:
We're seeing crashes like this on Touch Bar Macs running 10.12:

```
*** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'An instance 0x60000033c980 of class WebView was deallocated while key value observers were still registered with it. Current observation info: <NSKeyValueObservationInfo 0x600000436700> (
<NSKeyValueObservance 0x60000185afa0: Observer: 0x60000038a0e0, Key path: effectiveAppearance, Options: <New: NO, Old: NO, Prior: NO> Context: 0x0, Property: 0x60000185aee0>
)'
```

I'm not entirely sure why removeObserver: isn't being called in dealloc. I don't have a 10.12 machine on hand right now, so not observing effectiveAppearance at all seems like the safest path.

### DIFF
--- a/Sparkle/SUUpdateAlert.m
+++ b/Sparkle/SUUpdateAlert.m
@@ -219,7 +219,12 @@ static NSString *const SUUpdateAlertTouchBarIndentifier = @"" SPARKLE_BUNDLE_IDE
     prefs.standardFontFamily = @"-apple-system-font";
     prefs.defaultFontSize = (int)[NSFont systemFontSize];
     [self adaptReleaseNotesAppearance];
-    [self.releaseNotesView addObserver:self forKeyPath:@"effectiveAppearance" options:0 context:nil];
+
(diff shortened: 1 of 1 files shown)


kornelski:
Thanks!

michelf:
> I'm not entirely sure why removeObserver: isn't being called in dealloc.

Because `releaseNotesView` is weak. The view gets deallocated before the window controller and thus calling `removeObserver:` in `dealloc` is too late.

This works for 10.14, but probably just by luck. `releaseNotesView` should be made `strong` since there is no guaranty about when a weak variable is deallocated, and thus no guaranty `removeObserver:` will be called in time.

ksuther:
Aha, I didn't even think about that because I practically never weakly reference views. Thank you :)

PR #1302, opened 2018-10-29
SPUDownloaderSession leaked because the NSURLSession wasn’t invalidated

merrysuperstar:
NSURLSession has a special handling for it's delegate... Now the session is explicitly invalidated to avoid a retain cycle.

From the Apple documentation:

Important

The session object keeps a strong reference to the delegate until your app exits or explicitly invalidates the session. If you don’t invalidate the session, your app leaks memory until it exits.


kornelski:
Thank you

PR #1304, opened 2018-11-07
Improved install instructions

danielpunkass:
Hi folks - I was digging into this branch and found it a little difficult to get my head around some of the things I needed to do. I think the included commits which comprise changes ranging from link fixes to format changes for readability to semantic rewrites for (I hope) better understandabilty, will help the average newcomer to the project as they try to take advantage of this branch.


kornelski:
Thanks

PR #1305, opened 2018-11-07
Define APPLICATION_EXTENSION_API_ONLY = NO

danielpunkass:
As long as we don't ourselves depend upon any Frameworks that are incompatible with Application Extensions, then we should allow frameworks that require compatibility to be able to link with us.


kornelski:
Thanks! I wouldn't guess that Sparkle works with that limitation.

PR #1311, opened 2018-11-10
Fix unit tests

danielpunkass:
This pull request includes a few fixes to quiet warnings I encountered on Xcode 10.1 running on Mojave 10.14.1, and includes a change to the behavior of SUFileManager's updateAccessTimeOfItemAtRootURL... to work around a subtle problem where the access time of a directory changes as a side-effect of asking for a child's fileSystemRepresentation. This fixes a breaking unit test "testUpdateFileAccessTime".


kornelski:
That's  good catch. Thank you.

PR #1317, opened 2018-11-20
2 tiny fixes

HofiOne:
- bogus debug build warning fixed (as DEBUG always defined from the .xcconfig files this warning always displayed, even in release builds)
- fixed docs build warning message to prevent execution of  embedded shell script line (xcode cannot display this way the correct error message as the underlying echo command will try to execute the embedded command samples)


kornelski:
Thanks!

PR #1325, opened 2018-12-10
Fix compile of TestApplication

Bi11:
Info.plist is missing but MainMenu.xib is duplicated in TestApplication/hr.lproj.


kornelski:
Thanks

PR #1327, opened 2018-12-11
Remove brew update to reduce CI time

Bi11:
By removing unrelated `brew update` which runs for about 5 minutes, Travis-CI time is now reduced to about 30 seconds.


Bi11:
It seems `cocoapods` and `doxygen` may be updated by `brew update`.
Sorry for that.

Issue #1307, opened 2018-11-09
XPC branch: Uncaught exception when closing update window in Mojave


teologov:
Hi, I've just migrated to the `ui-separation-and-xpc` branch. Everything works well expect the next failure when closing the update window:
`*** Terminating app due to uncaught exception 'NSRangeException', reason: 'Cannot remove an observer  for the key path "effectiveAppearance" from  because it is not registered as an observer.'`

It might be related to #1288.

michelf:
I think this was fixed in #1293 and #1294. Is your version recent enough?

kornelski:
I've hacked around it

teologov:
Cool, thanks!

PR #1328, opened 2018-12-13
Fix release note WebView's appearance in dark mode

Bi11:
Most changes are cherry-picked from PR #1280 and #1288.


michelf:
Maybe you should consider also picking PR #1293 & #1294 and commit 45d27cc5d07191320e32bf1813e63232db54f247 (which solves issue #1307).

michelf:
Oh, I see you're observing the window's appearance instead of the release note view's. That could make the other changes less necessary.

Bi11:
Umm, yes. `NSWindowController.window` is a strong reference.

kornelski:
Nice :)

PR #1329, opened 2018-12-13
Fix release note web view's contextual menu

Bi11:
I found "Reload" in release note's contextual menu does not reload contents, but opens the page in web browser app. I don't know how to fix it yet.

But I found the code added in 8ff6c147f26600c9dc15d6f76bed5730dda68af6 are intended to remove "Reload" from the menu. But it does not take effect nowadays due to missing assignment of UIDelegate.


Bi11:
The "Reload" behavior is changed by the code below
https://github.com/sparkle-project/Sparkle/blob/4b0a037685d1136d33f64f721a16cf8e63bc44b4/Sparkle/SUUpdateAlert.m#L335-L341
We can check it by `actionInformation[WebActionNavigationTypeKey]` if necessary.

PR #1333, opened 2018-12-15
Support App Store URL scheme in release note webview

Bi11:
`macappstore://` links will be opened with App Store directly.


kornelski:
Won't that open the link immediately, without user interaction?

Bi11:
I mean opening `<a href="macappstore://...">` links in the release notes by user clicking.
If a `SUAppcastItem.releaseNotesURL` is set to a macappstore link, webview will not open it.

A check of `[actionInformation[WebActionNavigationTypeKey] integerValue] == WebNavigationTypeLinkClicked` can be added before `NSWorkspace openURL` to ensure the link is clicked by user. See https://github.com/sparkle-project/Sparkle/pull/1329#issuecomment-447142241 . But I'm not sure if it is necessary.

kornelski:
Can you add the check and rebase? Thank

Bi11:
I think the check of `WebActionNavigationTypeKey` is unnecessary, because `<a href="macappstore://...">` link in release note can only be opened by user clicking.
Thus adding `macappstore://` scheme to the whitelist is sufficient.

PR #1343, opened 2019-01-15
Correct old vs new version statement

riaanvn:
The old syntax with single or double dash and no spaces on either side of the dash is very non-standard English (and Afrikaans; I cannot speak for other languages). I have changed it to bring it in line with the same string on WinSparkle, https://winsparkle.org/ (I was unable to locate the source file on Github, but this screenshot shows the difference in formulation.

### DIFF
--- a/Sparkle/en.lproj/Sparkle.strings
+++ b/Sparkle/en.lproj/Sparkle.strings
@@ -7,10 +7,10 @@
 "%@ %@ is currently the newest version available." = "%1$@ %2$@ is currently the newest version available.";
 
 /* Description text for SUUpdateAlert when the update is downloadable. */
-"%@ %@ is now available--you have %@. Would you like to download it now?" = "%1$@ %2$@ is now available—you have %3$@. Would you like to download it now?";
+"%@ %@ is now available (you have %@). Would you like to download it now?" = "%1$@ %2$@ is now available (you have %3$@). Would you like to download it now?";
 
 /* Description text for SUUpdateAlert when the update informational with no download. */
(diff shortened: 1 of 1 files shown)


kornelski:
This is a technicality, but you can't change the left side in the translation files.

It's not the text to show, but a text-looking-like identifier for the program used to search for the correct translation. Any change there breaks compatibility with the code and all other translations.

As for the style, I think it's valid in American English to have em-dash without spaces. A hair space around it is usually used as a compromise between this and spaces around dashes/pauses used in other languages.

riaanvn:
Hi Kornel. I was definitely not hoping that the PR in its current state would be merged, but I thought since we are discussing this, I would rather discuss it to conclusion.
 
a) would you consider a merge request to bring Sparkle strings in line with WinSparkle strings (specifically changing the convention e.g. moving from em-dash to round brackets for parenthesis, based on my motivations above)? If not, I will have to accept it. It is your project after all. 
b) If you would accept a PR, should I close this one and create a new one, or would it make sense to fix this PR to keep the context of the discussion.
c) If I made a convincing case for round brackets parenthesis, would you mind or prefer doing the PR yourself rather than me doing it? (I would not mind doing it but whatever would be the least amount of effort for you)

kornelski:
Yes, Sparkle/en.lproj/Sparkle.strings is the correct one for US English.

If you modify only that file, then you can only modify the right side.

Yes, it'll mean translators will see old string. If you want to change the left side, you have to change it in *all* the translations *and* the source code that searches for this string.

PR #1344, opened 2019-01-17
Notify user about uninstallable updates

simonwagner:


kornelski:
Thanks for the PR. 

Can you explain what's the use-case for it?

simonwagner:
Added some fixes for bugs that I found during testing:

* The assert in `[-SUUIBasedUpdateDriver alertForValidationError:]` did trigger when validationError was `nil` (which is the case when no newer update is available).
* I misunderstood the code in `bestItemFromAppcastItems:getDeltaItem:withHostVersion:comparator:`, I understood it as selecting the best item that is newer than the current version, but it did only select the newest item in the appcast. I changed it a little bit now so that it returns the newest item which is also newer than the current version of the host app. This is necessary to make sure that the first run does not select an item which is older as the newest app but is still supported by the OS.

zorgiepoo:
Handling this now in #1850. Some other appcast selection changes have been made since then which needs to be considered. It was nice referencing this PR.

PR #1347, opened 2019-01-18
Check that effectiveAppearance is being observed before calling removeObserver

pNre:
Issue #1307 is also present in the master branch when `SUShowReleaseNotes = NO`.
This is basically the same fix as in 45d27cc.


kornelski:
Thanks

PR #1351, opened 2019-01-26
Use contentsOfDirectory(at:...) instead of atPath:....

Wevah:
And rename `apps` to `bundles`.

This removes the need for filtering based on starting with "." (hidden files), as well as the string/URL back-and-forth. Also pre-fetches `.typeIdentifierKey`.


kornelski:
Thank you

PR #1353, opened 2019-01-30
fix #1352

### LINKED ISSUE #1352: crash in -[SUPipedUnarchiver extractArchivePipingDataToCommand:arguments:notifier:]
i've received a crash report about a crash in extractArchivePipingDataToCommand:
```

Couldn't posix_spawn: error 35 Couldn't posix_spawn: error 35 (null) (
0 CoreFoundation 0x00007fff42638ecd __exceptionPreprocess + 256
1 libobjc.A.dylib 0x00007fff6e700720 objc_exception_throw + 48
2 CoreFoundation 0x00007fff42638cff +[NSException raise:format:] + 201
3 Foundation 0x00007fff4492eac7 -[NSConcreteTask launchWithDictionary:error:] + 4586
4 Sparkle 0x0000000105991069 -[SUPipedUnarchiver extractArchivePipingDataToCommand:arguments:notifier:] + 863
5 Sparkle 0x0000000105990c69 __64-[SUPipedUnarchiver unarchiveWithCompletionBlock:progressBlock:]_block_invoke + 86
6 libdispatch.dylib 0x00007fff6f77fd53 _dispatch_call_block_and_release + 12
7 libdispatch.dylib 0x00007fff6f780dcf _dispatch_client_callout + 8
8 libdispatch.dylib 0x00007fff6f783240 _dispatch_queue_override_invoke + 767
9 libdispatch.dylib 0x00007fff6f78ed26 _dispatch_root_queue_drain + 325
10 libdispatch.dylib 0x00007fff6f78f4bd _dispatch_worker_thread2 + 90
11 libsystem_pthread.dylib 0x00007fff6f9c16dd _pthread_wqthread + 619
12 libsystem_pthread.dylib 0x00007fff6f9c1405 start_wqthread + 13
```

the error is apparent in 
SUPipedUnarchiver.m:116:

`            [task launch];
`

you should never call -[NSTask launch] without enclosing it in an @try block as its known to throw exceptions - its rare but it happens, e.g. also on systems that definitely need rebooting. i've often seen posix_spawn error 1 and 35.

the fix is easy:

```

    @try
    {
        [task launch];
    }
    @catch (NSException *e)
    {
        // insert error handling here
    }

```

core-code:


kornelski:
Thanks! There's one more thing - you can't just `return`, because the error variable is supposed to be used at the end of the function with `notifyFailureWithError`. You could put `goto fail` there or initialize error to `NULL`, and wrap the rest of the function in `if (!error)`

core-code:
sorry, should be better now.

the NSError was already initialised, and in ARC code this isn't even necessary.

i've moved 'archiveOutput' because else it would have been necessary to close that if the exception occurs.

kornelski:
Thanks!

PR #1356, opened 2019-02-04
Add new SUPKGUnarichiver

NoWiseMan:
Currently Sparkle only accept pkg files in zip archives, since pkg file are flat packages this is not needed anymore. This PR adds the handle of pkg files as direct downloads.
Everything it does is add a handler for a PKG Unarchiver which only notifies that it is finished.

### DIFF
--- a/Sparkle/SUPKGUnarchiver.m
+++ b/Sparkle/SUPKGUnarchiver.m
@@ -1,0 +1,55 @@
+//
+//  SUPKGUnarchiver.m
+//  Sparkle
+//
+//  Created by Thomas Schmitt
+//  Copyright 2013-2019 Thomas Schmitt. All rights reserved.
+//
+
+#import "SUPKGUnarchiver.h"
+#import "SUUnarchiverNotifier.h"
+#import "SULog.h"
+
+
+@interface SUPKGUnarchiver ()
+
+@property (nonatomic, copy, readonly) NSString *archivePath;
+
+@end
+
+@implementation SUPKGUnarchiver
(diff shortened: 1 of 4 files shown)


Sparkle/SUPKGUnarchiver.m
```diff
@@ -0,0 +1,55 @@
...
+
+
++ (BOOL)canUnarchivePath:(NSString *)path
+{
+	return [path.pathExtension isEqualToString:@"pkg"];
+}
+
++ (BOOL)mustValidateBeforeExtraction
+{
+    return NO;
```
kornelski: Isn't that a security vulnerability that bypasses Sparkle's signature checks?

NoWiseMan: Don't think so, because PKGs are checked later. Sparkle gave an error message when I return YES. It also gave an error when signature check fails (hmm .... this was in before the new EdDSA key ... I check this again)

kornelski: Can you add a unit test that proves its correct? (a test that passes with bare pkg signed, and a test that fails with bare pkg when signature is invalid)

RobFog:
"SUPKGUnarichiver" should probably read "SUPKGUnarchiver"?

And it seems the Travis build failed.

zorgiepoo:
Thanks for this PR. I used it as a reference for #1745 and #1746, only notable difference is I disallow interactive package installs this way and added support for 2.x too.

PR #1357, opened 2019-02-04
Update imports to use framework-style imports consistently

danielpunkass:
Starting in Xcode 10.3 a new warning CLANG_WARN_QUOTED_INCLUDE_IN_FRAMEWORK_HEADER allows situations where a framework-based import vs. local file based import is ambiguous because of the use of quoted header file name instead of import-style imports. This pull request both imposed the warning on Sparkle and fixes all the imports in the project to avoid the warnings both within the project and on behalf of any clients that import headers from Sparkle.


kornelski:
Interesting! I thought `<>` imports were not allowed for non-Apple frameworks.

kornelski:
Is there a risk of breaking other people's projects that use Sparkle's public headers, but don't set up framework include paths?

danielpunkass:
I don't think there is any risk of breaking clients, because the <> angle form just means "look for this in a standard place", and by linking against the Sparkle framework, the client will have Sparkle's Headers among its standard locations to look.

The angle bracket format isn't reserved for Apple, but just used to distinction between "local" header files and files that should be located in an external location. In the context of Sparkle's own headers, referencing them as "externally findable" in this way ensures that there won't be any confusion with a local copy of a header, and that Sparkle won't ever accidentally neglect to publish a header file that is intended to be referenced publicly in this way.

In short: instead of looking everywhere for these imports, with the angle brackets Xcode will only look among Sparkle's public header files.

PR #1363, opened 2019-02-22
Add delegate methods to suppress update alert.

gnachman:
## Background

Users want to know when an update is available, but some of them find a popup window to be too obtrusive [1]. I would like to offer users to option to receive a more subtle update notification. For inspiration, VS Code shows this when a new update is available:

image

## Implementation

This commit adds two new delegate methods:

```
- (BOOL)updaterShouldShowUpdateAlertForScheduledUpdate:(SUUpdater *)updater;
```

This is called only for scheduled updates. The delegate may choose to suppress the update, in which case the update is immediately aborted. That allows the "Check for Updates" menu item to remain enabled.

A delegate that wishes to present its own UI would likely use this notification to trigger the behavior, having used `updater:didFindValidUpdate:` to record any relevant information about the appcast item such as its version number.

The second new delegate method is:

```
- (void)updater:(SUUpdater *)updater didDismissUpdateAlertPermanently:(BOOL)permanently forItem:(SUAppcastItem *)item;
```

This is useful because it informs the application when the user wants to skip an update. It can then clear its own notification about the update in response to the user's expressed preference.

All update drivers other than the scheduled update driver remain unaffected by this change.

#### Footnotes

1. https://gitlab.com/gnachman/iterm2/issues/7502#note_143351094

### DIFF
--- a/Sparkle/SUScheduledUpdateDriver.m
+++ b/Sparkle/SUScheduledUpdateDriver.m
@@ -47,4 +47,15 @@
     return YES;
 }
 
+- (BOOL)shouldShowUpdateAlert {
+    id<SUUpdaterPrivate> updater = self.updater;
+    id<SUUpdaterDelegate> updaterDelegate = [updater delegate];
+
+    if ([updaterDelegate respondsToSelector:@selector(updaterShouldShowUpdateAlertForScheduledUpdate:)]) {
+        return [updaterDelegate updaterShouldShowUpdateAlertForScheduledUpdate:self.updater];
+    }
+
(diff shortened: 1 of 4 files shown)


kornelski:
Thanks for the PR.

I presume apps will use this to return `NO` and display their own UI instead. Should the first delegate method also give appcast item?

gnachman:
Sure, that makes sense. I've updated the PR.

kornelski:
Thanks!

gnachman:
😃Thank ou!

PR #1365, opened 2019-02-22
Eliminate Swift warnings by NSObject pointer reference to NSInvocation

fjaeger:
The proposed change would not break Objective-C consumers of the Sparkle framework, while eliminating compiler warnings in Swift environments due to nonexistent NSInvocation class.

Alternatively, we could also define a simple `SUInvocation` interface, which would even be accessible from Swift implementations. Thoughts?


kornelski:
Breaking existing code is not an option.

Perhaps for Swift it'd be better to add another delegate method that takes a block?

fjaeger:
I don't think this is breaking existing code. Objective-C implementations will still be able to use the passed pointer to an NSInvocation instance

PR #1369, opened 2019-03-02
use modules when feature is available to suppress warnings

nivekkagicom:
The header fails to check if modules are being used and always does #import which will generate a warning if modules are being used. Simply check if modules are being used and do @import if they are or fall back to #import of they are not.


kornelski:
It's merged now, thanks!

kornelski:
And sorry for the delay.

PR #1376, opened 2019-03-13
Use AVAILABLE macro to allow compiling with Xcode 8

hofman:
Use a macro for the @available keyword to allow compiling with Xcode 8


hofman:
BTW, I can confirm with these changes Sparkle builds in Xcode 8.

kornelski:
That's great. Thank you.

PR #1381, opened 2019-03-19
Allow to configure update groups for updating in phases

fjaeger:


kornelski:
Could you explain what does it do?

fjaeger:
Absolutely!

You can configure the phased updating mechanism by the following two keys (e.g. Info.plist):
* SUNumUpdateGroups : Defines how many groups should be considered when rolling out an update
* SUUpdateGroupInterval : Defines the time interval between rollouts for each group. 

The idea behind a phased rollout of an update is that you might want to evaluate certain behavior or crash reports or other indicators of the first group of users before the same update gets further rolled out to more groups. If you experience any issue with an update on one of the first update groups, you could still revoke the update by deleting the appcast item. Otherwise, the rollout would continue as usual - just a little delayed due to the update groups, which is simply a segmentation of your user base (based on their device serial number).

This feature is purely optional. If you do not provide SUNumUpdateGroups > 1 in your Info.plist, then the updating behavior would be the same as before.

fjaeger:
The proposed mechanism also respects critical updates. For critical updates, the rollout is performed immediately - skipping the grouping.

kornelski:
This is a neat idea. I've been meaning to add something like that.

However, that adds considerable complexity: dependence on IOkit, unique system identifier (so apps using sparkle could look like they're tracking users), to new keys and date parsing.

Would it work for you if it was simplified to update at random? In the appcast you'd insert a number that is a probability that an update check will result in a download, e.g. if you have checks set to one a day, and put 5%, about 5% of your users will upgrade every day.

fjaeger:
I understand your concerns.

For me it would also be okay to find some alternative way of assigning a user to an update group. My approach via IOKit was trying to get this pseudo-random assignment to an update group.

Your idea about defining the ratio of valid requests in the appcast is interesting, but it would also depend on the interval the updater checks for updates. I think we need some locally available information that decides when a user is allowed to get the update.

fjaeger:
If we just don't want to add the dependency on IOKit, we could also create a pseudo-random group identifier in-memory on framework initialization. The only disadvantage is that the user might be in a different group each time he/she launches the app using Sparkle ...  What do you think?

kornelski:
A random identifier would be better. It could be stored in nsuserdefaults the same way last check date is.

I think it should be controlled entirely from the appcast, not from Info.plist, because you might have different requirements/priorities depending on the update (e.g. release a big one very carefully).

michelf:
Some questions and observations:

- If a user checks for updates manually, will that shortcut the wait time or will it reports there is no update?
- Are all groups the same size?
- Shouldn't the deployment parameters (number of groups & interval) be part of the appcast instead? It seems weird to me that parameters in a previous version of the app would decide of the deployment schedule for the next version. Also, if you mess up these parameters with something silly (9999 groups with an interval of 9999 days) you can fix it server-side.
- You can store a randomly-generated group number in the user defaults upon first seeing an update with that feature, and reset it after a successful update to preserve privacy.

kornelski:
Yeah, manual check should probably bypass that feature. In case of ImageOptim sometimes sites like MacUpdate notice I've released a beta, and I've had people ask me why Check for Updates isn't seeing it.

kornelski:
I'm also concerned about usability of dates in the appcast, especially the US format with day of the week. ISO format YYYY-MM-DD might be more robust, but in general it's a thing I expect people to get wrong (backwards dates in US, different separators, leading zeros, typos).

Could we do something clever here instead? e.g. read Last-Modified HTTP header and count days from that?

fjaeger:
To answer the previous questions:
* manual update requests should behave normally and are not impacted by the grouping (I propose to only update the SUScheduledUpdateDriver)
* Assuming a uniform distribution of the random number generator, the groups should be more or less of the same size, yes

Alright, let's summarize what I am about to change:

* configuration of phased updating via appcast instead of Info.plist
* instead of IOKit, we generate a random number on first launch, store it in user defaults and use it to define the update group (random number is reset after each successful update)

kornelski:
Sounds good!

michelf:
Same-size groups is different from [how phased rollouts works on the App Store][1]. Not that I think either approach is wrong, just some difference to be aware of.

[1]: https://developer.apple.com/app-store-connect/whats-new/?id=31070842

kornelski:
Maybe explicit concept of groups isn't necessary, and it could be a percentage? (equivalent to hardcoding 100 groups)

fjaeger:
Interesting to see how the AppStore handles phased rollouts.

For the definition of the phases, we need two pieces of information, I think:
* the time period for each phase
* the number/percentage of each phase

Ideas, comments?

fjaeger:
If we want to define the rollout by one parameter, then it could be the time period of the full rollout, e.g. 7 days and we would just divide this into N (hardcoded) groups internally.

fjaeger:
Did some updates to the PR. Please let me know if it works for you.

If we want to have a more sophisticated definition of the update groups (e.g. similar to what Apple does now), we could add this easily later.

kornelski:
Thanks for the changes. Let's try that.

fjaeger:
What's the planned release schedule?

michelf:
Wouldn't it make more sense for the interval to be specified in days?

Funny timing: Apple [added the feature to the **Mac** App Store][1] yesterday. It was iOS-only previously.

[1]: https://developer.apple.com/app-store-connect/whats-new/?id=phased-release-macos

fjaeger:
I also thought about this, but other time intervals in Sparkle are also defined in seconds

PR #1382, opened 2019-03-22
Updated Turkish Translations and XIB Layout Fixes

bitigchi:


kornelski:
> "Install and Relaunch" = "Kur ve yeniden başlat";

This is a label on a button. Does Apple use lowercase words on multi-word buttons in Turkish?

bitigchi:
Not really. In Turkish, general practice is to use lowercase letters for imperative cases, but Apple keeps them all uppercase anyway. Do you want to keep it consistent with the rest of the system?

kornelski:
Yes, we try to blend in and follow whatever Apple is doing.

bitigchi:
Fixed.

kornelski:
I've merged it in fc3221c5d2ddedf443b0b46e2fe64efdefe32f2b

PR #1385, opened 2019-04-02
Phased rollout interval is an appcast item attribute and not its own element

fjaeger:


fjaeger:
Sorry for having missed this in the first place!

kornelski:
I think that highlights the lack of tests. Could you create a test for this?

fjaeger:
Just added a basic test for getting the phased rollout interval from an appcast item

kornelski:
Thanks

Issue #1371, opened 2019-03-05
ui-separation-and-xpc branch doesn't build


galad87:
I cloned the repository, switched to the ui-separation-and-xpc, and tried to build it, but got a lot of duplicated classes definition and similar errors.

kornelski:
I can't reproduce this. Are you sure it's not a problem with your Xcode or your project setup?

galad87:
It could be a Xcode 10.2 beta issue, I'll try again later with 10.1.

Here's the output anyway: https://gist.github.com/galad87/f7271123f596d98e8083ad87fa8306f6

svenemtell:
I get the same error when running `make release` in the latest `ui-separation-and-xpc branch`.
I also get a similar error when trying to build the "Distribution" scheme from within the Xcode project.
Building products (Sparkle.framework and the XPC services) one at a time works fine.
I am using `macOS 10.14.3` and `Xcode 10.1 (10B61)`.

xocialize:
I'm having issues building as well in both 10.1 and 10.2(beta).  Last of my make release output is:

** BUILD FAILED **


The following build commands failed:
	CompileC /var/folders/7q/2cq4mfm560g0cynz6t84zm1m0000gn/T/Sparkle.d4BQKk/Build/Intermediates.noindex/Sparkle.build/Release/SparkleCore.build/Objects-normal/x86_64/SPUAutomaticUpdateDriver.o Sparkle/SPUAutomaticUpdateDriver.m normal x86_64 objective-c com.apple.compilers.llvm.clang.1_0.compiler
(1 failure)
make: *** [release] Error 65

Let me know if there's any other info I can provide that might be of use.

kornelski:
That error message is not enough for me to diagnose the problem. Are there other errors?

xocialize:
Here's the complete output: https://gist.github.com/xocialize/7f5d8b119bf143e1cdc83f217a0bf90b

When I try and build the distribution scheme from xcode I get:

Duplicate interface definition for class 'SUAppcast'

I'm currently running this from a Mojave machine running the latest Beta but I've also reproduced this on a fresh install of release Mojave and Xcode 10.1 

I'm happy to provide any other info you can point me towards providing.  Thanks.

PR #1390, opened 2019-04-16
Fix header imports in `ui-seperation-and-xpc` branch

maarf:
> Allows building Distribution scheme with Xcode 10.2

I was building `ui-seperation-and-xpc` branch and ran into the same issues described in #1371.

I’m not sure why these duplicate deceleration errors are suddenly appearing when building with newer Xcode, but this fixed them.


kornelski:
Thanks

PR #1391, opened 2019-04-16
Updated XPC codesigning to enable hardened runtime.

michelf:
This adds the `-o runtime` flag when calling `codesign` in the `codesign_xpc` script, enabling hardened runtime for XPCs.

I don't think this to be sufficient for notarization in the general case when Autoupdate and Updater.app are included in the framework. It is sufficient for me since I specifically remove these two from the framework in my build process and rely solely on XPCs.

### DIFF
--- a/bin/codesign_xpc
+++ b/bin/codesign_xpc
@@ -10,7 +10,7 @@ def sanitize(argument):
     return ('"' + argument + '"' if ' ' in argument else argument)
 
 def _codesign_service(identity, path, entitlements_path=None):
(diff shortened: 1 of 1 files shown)


kornelski:
Thanks

michelf:
I just realized: this could break when `codesign` is from Xcode less than 12. Not sure if this is a problem.

kornelski:
Well, if hardened runtime becomes required, then Apple has effectively killed Xcode 11.

PR #1392, opened 2019-04-17
Improved error when running from translocated location

MichaelBuckley:
Hi all,

For many, many years now, we've been running a private fork of Sparkle for our products, and we've found that the further we've diverged, the harder it has become to integrate fixes from upstream. This was particularly difficult when integrating changes to enable hardened runtime for the updater app. As such, we would like to switch to using the official Sparkle if at all possible. This will require us to port a few features back from our fork. I am still working my way though all the differences, and hopefully won't need to submit many pull requests.

This first pull request adds an error message that will appear if a user attempts to update an app translocated app. Because app translocation runs the app from a read-only volume, users will currently see an error informing them that the app is running from a read-only location, but some users found this to be confusing.

The pull request also contains translations for the error string in the languages we localize our apps for. Unfortunately, this is a small fraction of the languages supported by Sparkle.

This is my first time submitting a pull request to Sparkle, and I read the code of conduct, but I could not find any other requirements about submitting pull requests. If it is deficient in any way, I will be more than happy to rectify any issues.

Thank you in advance for reviewing this request.

### DIFF
--- a/Sparkle/fr.lproj/Sparkle.strings
+++ b/Sparkle/fr.lproj/Sparkle.strings
@@ -4,6 +4,8 @@
 
 "%1$@ can't be updated, because it was opened from a read-only or a temporary location. Use Finder to copy %1$@ to the Applications folder, relaunch it from there, and try again." = "%1$@ ne peut pas être mis à jour quand il fonctionne à partir d’un volume en lecture seule, comme une image disque ou un lecteur optique. Déplacez %1$@ dans votre dossier Applications, relancez-le à partir de là et réessayez.";
 
+"Quit %1$@, move it into your Applications folder, relaunch it from there and try again. %2$@ can’t be updated if it’s running from the location it was downloaded to." = "Quittez %1$@ et déplacez-le dans le dossier Applications, relancez-le à partir de ce dossier et essayez à nouveau. %2$@ ne peut pas être mis à jour s'il est exécuté depuis le dossier dans lequel il a été téléchargé.";
+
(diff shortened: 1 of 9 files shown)


Sparkle/fr.lproj/Sparkle.strings
```diff
@@ -4,6 +4,8 @@
 
 "%1$@ can't be updated, because it was opened from a read-only or a temporary location. Use Finder to copy %1$@ to the Applications folder, relaunch it from there, and try again." = "%1$@ ne peut pas être mis à jour quand il fonctionne à partir d’un volume en lecture seule, comme une image disque ou un lecteur optique. Déplacez %1$@ dans votre dossier Applications, relancez-le à partir de là et réessayez.";
 
+"Quit %1$@, move it into your Applications folder, relaunch it from there and try again. %2$@ can’t be updated if it’s running from the location it was downloaded to." = "Quittez %1$@ et déplacez-le dans le dossier Applications, relancez-le à partir de ce dossier et essayez à nouveau. Transmit ne peut pas être mis à jour s'il est exécuté depuis le dossier dans lequel il a été téléchargé.";
```
michelf: Perhaps it should be `%2$@` instead of of "Transmit" in the localized string.

MichaelBuckley: Thanks for catching that. I have pushed a fix. Sorry for the inconvenience.

kornelski:
Thank you. It's great that it comes with at least some translations.

PR #1394, opened 2019-04-19
Custom updater icon

MichaelBuckley:
Hi again,

This pull request contains an option to set a custom icon for the Autoupdate app within Sparkle. We use this option internally to set different icons for different projects. Because the Autoupdate icon can appear in the dock during installation, we've found it has been helpful to users to associate the icon with our app, but not to use the exact same icon as our app, which can cause confusion.

Although we could modify the Autoupdate app target to use a different icon, this change change brings two benefits. First, because we use different Autoupdate icons for different projects, we would need to keep separate Sparkle branches for each project if we modified the Autoupdate target. Second, we would need to make sure to re-apply the custom icons every time we updated from upstream.

I'm wary of asking Sparkle Project to accept too many changes with optional feature flags, but solving these problems otherwise requires us to maintain multiple internal Forks of Sparkle, and we are trying to avoid using an internal fork at all. I also believe this could be useful to other users who are in a similar position.

Additionally, because this change required me to modify SUHost, I added more complete nullability annotations to that file. I would be happy to split these changes into separate pull requests if that makes things easier for you, or remove one of the commits if you do not wish to take it.

Thanks again!

### DIFF
--- a/Sparkle/Autoupdate/Autoupdate.m
+++ b/Sparkle/Autoupdate/Autoupdate.m
@@ -244,6 +244,19 @@ int main(int __unused argc, const char __unused *argv[])
             [application activateIgnoringOtherApps:YES];
         }
         
+        NSBundle *theBundle = [NSBundle bundleWithPath:[args objectAtIndex:1]];
+        SUHost *host = [[SUHost alloc] initWithBundle:theBundle];
+        NSString *customIconPath = host.updaterCustomIconPath;
+        
+        if (customIconPath) {
+            NSImage *customIcon = [[NSImage alloc] initWithContentsOfFile:customIconPath];
+            
+            if (customIcon) {
+                [[NSApplication sharedApplication] setApplicationIconImage:customIcon];
+            }
+        }
+
+        
         AppInstaller *appInstaller = [[AppInstaller alloc] initWithHostPath:[args objectAtIndex:1]
                                                                relaunchPath:[args objectAtIndex:2]
                                                             parentProcessId:[[args objectAtIndex:3] intValue]
--- a/Sparkle/SUConstants.h
+++ b/Sparkle/SUConstants.h
@@ -55,6 +55,7 @@ extern NSString *const SULastProfileSubmitDateKey;
 extern NSString *const SUPromptUserOnFirstLaunchKey;
 extern NSString *const SUDefaultsDomainKey;
 extern NSString *const SUEnableJavaScriptKey;
+extern NSString *const SUUpdaterCustomIconKey;
 extern NSString *const SUFixedHTMLDisplaySizeKey __attribute__((deprecated("This key is obsolete and has no effect.")));
 extern NSString *const SUAppendVersionNumberKey __attribute__((deprecated("This key is obsolete. See SPARKLE_APPEND_VERSION_NUMBER.")));
 extern NSString *const SUEnableAutomatedDowngradesKey __attribute__((deprecated("This key is obsolete. See SPARKLE_AUTOMATED_DOWNGRADES.")));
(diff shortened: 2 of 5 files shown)


kornelski:
Thanks for the PR.

I agree that it's better to use a dedicated icon related to the app than the generic autoupdate one.

I wonder if we should just replace it with app's own icon. I'm not sure what's wrong with using just an unmodified app's icon. We already show all update prompts and progress within the app, with the app's own icon, so from user perspective the app is updating itself.

I get that's just port of a feature from your app, but if it's an optional feature, that's more installation steps for new users, and in the end few apps will benefit. A change of defaults would improve the situation for everyone.

michelf:
Another option could be to compose an icon at runtime. For instance: you draw the app icon a bit smaller and surround it with spinning arrow-style icons.

kornelski:
Yeah, some overlay could work, too. 

I am not sure about cost/benefit of doing so. The autoupdater icon is seen rarely. It'd require some image manipulation code to be added to Sparkle. And authors would like to customize the overlay icon, so we'd be back to square one. 

I think just using host's icon is the 80/20 solution.

michelf:
Combining images at runtime is not that complicated and would provide a nice default for all apps:
``` objc
NSImage *appIcon = [[NSWorkspace sharedWorkspace] iconForFile:theBundle.bundlePath];
NSImage *overlay = [NSImage imageNamed:@"Overlay"];
NSImage *finalImage = [NSImage imageWithSize:appIcon.size flipped:NO drawingHandler:^BOOL(NSRect iconRect) {
    [appIcon drawInRect:NSInsetRect(iconRect, iconRect.size.width/5, iconRect.size.width/5)];
    [overlay drawInRect:iconRect];
    return true;
}];
[NSApplication sharedApplication].applicationIconImage = finalImage;
```
Personally, if I needed a custom updater icon I'd just add some post-build script in the app target tweaking the the icon resource in the updater. That would work for customizing the overlay too.

Adding API surface seems a bit unnecessary.

MichaelBuckley:
Thanks for the feedback!

For our purposes, you're right. It looks like we could replace the icon as a post-build step. I wasn't around when the decicison to implement this as a setting was made, but looking at the history, I suspect it had to do with a code signing problem that no longer applies.

As for why we didn't use the app icon for the Autoupdate app, while it's true that the app only shows up briefly in the Dock, having a duplicate icon still caused user confusion on more than one occasion.

As for this pull request, I'm more than happy to remove the second commit and just leave the nullability auditing if you think it's worthwhile. I could also rework it to incorprate a combination of the app and updater icon as michelf suggested. I also wouldn't be dissapointed if we just closed it. Just let me know what you'd prefer, and I'll be happy to accomodate.

michelf:
I made [a quick patch for composing dynamically the icon][1]. (Note: this was made from the ui-separation-and-xpc branch, not master.) The overlay icon was made from the arrows in the sparkle test app icon; it's not exactly at the correct scale but I think it's still good enough upscaled. I could turn that into a pull request depending on what is decided here.

[1]: https://github.com/michelf/Sparkle/commit/b2ad1613fa0616601d41c8be5c2141893b2e3f6b#diff-1463f643bb63ae1aaf38c08c482f6b52

zorgiepoo:
If I recall correctly, it is a goal of 2.x branch to 1. use the icon of the bundle that's being updated (it has a setApplicationIconImage: call), and 2. be more efficient (faster install) to minimize the number of cases this progress app shows up.

MichaelBuckley:
That's great! I was unaware of the 2.x branch when I submitted this pull request.

kornelski:
https://github.com/sparkle-project/Sparkle/pull/1394#issuecomment-601577934

PR #1403, opened 2019-05-04
Correction of Czech localization inconsistency

vojtakonarik:
The terminology was updated according to the new one used since macOS 10.5 Leopard.


kornelski:
Thank you

PR #1405, opened 2019-05-07
Fix various warnings

MichaelBuckley:
Hi again.

Since my last pull request, I've been successful in integrating the rest of the features from our Sparkle fork directly into our apps. No need to put them in the Sparkle code base. However, I noticed that the mainline Sparkle uses a submodule, ed25519, which has numerous warnings when compiled with Sparkle's warning flags. I submitted a [pull request](https://github.com/orlp/ed25519/pull/25) to that project a couple weeks ago to fix the warnings, but I haven't heard anything from the maintainer, so I disabled the warnings for those files only.

In the process, I also took care of warnings where Xcode asked to update to Swift 5 and base localization. There were no source changes required for the conversion to Swift 5, so I just updated the project file. However, this enabled more warnings. Apparently setting your project to use Swift 5 while using -Weverything will warn you about sending messages to `id`, so I cast the values to more specific types to silence the warning.

Since we're targeting 10.7 as the deploy target, I couldn't enable base localization. But I was able to edit the project file to add "Base" to the list of known regions. This disables the warning without affecting the build output.

Please let me know if you would like any changes to this. I can, for example, keep the Swift version unspecified and just change the `LastSwiftMigration` in the project.


zorgiepoo:
As this PR is going stale, please open a new one if you feel this needs to still be addressed. Some things have changed since then, including deployment target being increased, base localization being used now, and we have a slightly more active PR addressing Swift 5.

PR #1409, opened 2019-05-17
add delegate userDidSkipThisVersion

BobZombie:
I added it because I wanted a delegate when the user pressed the  "skip This Version" button. 
I am happy if you did merge it.
Thank you.

### DIFF
--- a/Sparkle/SUUIBasedUpdateDriver.m
+++ b/Sparkle/SUUIBasedUpdateDriver.m
@@ -163,7 +163,9 @@
             break;
 
         case SUSkipThisVersionChoice:
-            [[updater delegate] userDidSkipThisVersion:self.updater];
+            if ([[updater delegate] respondsToSelector:@selector(userDidSkipThisVersion:)]) { 
+                [[updater delegate] updater:self.updater userDidSkipThisVersion:self.updateItem]; 
+            }
             [self didDismissAlertPermanently:YES forItem:item];
             [self.host setObject:[self.updateItem versionString] forUserDefaultsKey:SUSkippedVersionKey];
             [self abortUpdate];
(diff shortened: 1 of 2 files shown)


kornelski:
Seems like a reasonable thing to add.  But please check if the method is implemented before calling it. See how other delegate methods do it.

Sparkle/SUUIBasedUpdateDriver.m
```diff
@@ -162,6 +163,7 @@ - (void)updateAlertFinishedWithChoice:(SUUpdateAlertChoice)choice forItem:(SUApp
             break;
 
         case SUSkipThisVersionChoice:
+            [[updater delegate] userDidSkipThisVersion:self.updater];
```
Neoster: One should check for responsiveness before calling the method on the delegate.

`if ([[updater delegate] respondsToSelector:@selector(userDidSkipThisVersion:)]) {
        [[updater delegate] updater:self.updater userDidSkipThisVersion:self.updateItem];
    }`

Neoster:
@BobZombie are you going to process the requested changes? Need any help with that?

kornelski:
Thanks

Sparkle/SUUIBasedUpdateDriver.m
```diff
@@ -162,6 +163,9 @@ - (void)updateAlertFinishedWithChoice:(SUUpdateAlertChoice)choice forItem:(SUApp
             break;
 
         case SUSkipThisVersionChoice:
+            if ([[updater delegate] respondsToSelector:@selector(userDidSkipThisVersion:)]) { 
+                [[updater delegate] updater:self.updater userDidSkipThisVersion:self.updateItem];
```
LeoNatan: How did this pass review? respondsToSelector is checked with one selector, while another is called inside. The inside selector is unknown, causing compile error.

https://github.com/sparkle-project/Sparkle/issues/1455

PR #1413, opened 2019-05-28
Minor fixes

tonyarnold:
As discussed in #363, I'll start small and bring over changes a piece at a time. 

This first PR is more of a tidy up: 

The vendor dependencies "bsdiff" and "ed25519" have been converted to static library targets, and the filesystem has been cleaned up to group them both under `$(PROJECT_ROOT)/Vendor`. 

Both were being included in multiple targets as raw source files, which can lead to maintenance issues down the road.


kornelski:
Great! Thank you

PR #1425, opened 2019-06-26
Make XPC service prefix "org.sparkle-project" customizable

DivineDominion:
I thought it'd be weird if we all switched to the XPC branch and then Activity Indicator would show dozens of entries for `org.sparkle-project.Downloader` and such. I assume y'all are changing the bundle Ids anyway in your apps. So I figured, why not make the change easier and limit it to a single line in the Config file?

All you need to do is change `XPC_SERVICE_BUNDLE_ID_PREFIX` in `ConfigCommon.xcconfig`  to something like `com.example.AmazingApp` and the XPC services will be recognizable by end users.

Single-line changes should also be easier to keep in sync with upstream updates to the branch.


kornelski:
Thanks

Issue #1411, opened 2019-05-20
Sparkle claims 'You’re up-to-date!' when in fact the feed is broken.


core-code:
This is not a bug report, but i want to encourage discussion of Sparkle's default behaviour of lying to its user and to developers alike. I'd like to believe this was a conscious design-decision but i think the negative sides far outweigh the positive ones and i believe the current implementation has led to the fact that (Sparkle) auto-updates are just plain broken and not working in a large number of apps.

Currently Sparkle works like this - you click 'Check for Updates' and there are 3 cases:
• there was an error loading the feed at all, which leads to an error message (good!)
• the feed was loading fine, and Sparkle found an update (even better!)
• the feed was loading fine, but Sparkle did not find find a version that is newer than the installed version in the feed. now this case is where the problems start. in this case Sparkle will always throw up a dialog 'You’re up-to-date! AppName x.y.z is currently the newest version available.' however, in 2 out of 3 options, this message is actually a lie. lets examine these cases:

1.) the installed version is actually identically to the newest version (found in the feed). this is good and actually the only case where Sparkle isn't lying.
2.) the installed version is actually newer than to the newest version (found in the feed) - i.e. the user has a beta version installed that is not yet official. claiming that the user has the latest version is wrong - he actually has a newer version. the facts should be stated as they are.
3.) Sparkle thinks the installed version is newer than the newest version because the feed is broken. technically this the same case as above (the installed version is numerically newer than feed version), however the cause is a different one: the user does not have something newer-than-the-latest installed, but the feed is just plain broken and all versions numbers listed in the feed are actually numerically lower than actually released versions. if you think this is implausible, see the last-most section.

why is lying in these cases bad? as developers we all know that lying to the user is often a good idea, especially when situations are too complicated to explain or do not concern the average user anyway. 
• if the user has a beta version that is newer than the latest official version, he'd surely understand a message like "You're more than up-to-date! You have version a.b.c while the newest official version of 'AppName' is still at x.y.z".
• if the feed is just plain broken, this is were the real problem is at. if any user, regardless of the installed version will always get the 'You’re up-to-date!' message, and not notice any problem that they could communicate to the developer. due to this message the developer themselves will also never notice any problem. i have the distinct impression many developers test their update, check for updates, and if they get the "You’re up-to-date" message, they think their feed is fine and move on. but due to the behaviour mentioned above they didn't actually test anything.

[...]

PR #1427, opened 2019-06-27
Fix issue #1411

core-code:
As discussed in bug #1411 this pull request changes two things:

• if the user is not running the latest, but a 'newer-than-latest' version, a slight hint of this fact is given instead of miscommunication that claims his version is identical to the newest version

• if Sparkle was unable to load the appcast, an error message is given instead of wrongly claiming that the copy is up-to-date. it is weird that Sparkle given error messages under some error conditions (no network) but just blindly claims that you're up-to-date under other conditions (unable to load appcast URL, empty appcast, etc). 

as mentioned in #1411 i believe both behaviours fixed in this PR have been causing much harm to date, both to developers as well as to users.

### DIFF
--- a/Sparkle/el.lproj/Sparkle.strings
+++ b/Sparkle/el.lproj/Sparkle.strings
@@ -4,6 +4,8 @@
 
 "%@ %@ is currently the newest version available." = "Το %1$@ %2$@ είναι η τελευταία διαθέσιμη έκδοση.";
 
+"%@ %@ is currently the newest version available.\n(You are currently running version %@.)" = "Το %1$@ %2$@ είναι η τελευταία διαθέσιμη έκδοση.\n(Die derzeit installierte Version ist %3$@.)";
+
 /* Description text for SUUpdateAlert when the update is downloadable. */
 "%@ %@ is now available--you have %@. Would you like to download it now?" = "Το %1$@ %2$@ είναι πλέον διαθέσιμο--έχετε το %3$@. Θέλετε να το κατεβάσετε τώρα;";
 
--- a/Sparkle/SUBasicUpdateDriver.h
+++ b/Sparkle/SUBasicUpdateDriver.h
@@ -17,6 +17,8 @@
 @interface SUBasicUpdateDriver : SUUpdateDriver <SPUDownloaderDelegate>
 
 @property (strong, readonly) SUAppcastItem *updateItem;
+@property (strong, readonly) SUAppcastItem *latestAppcastItem;
+@property (assign, readonly) NSComparisonResult latestAppcastItemComparisonResult;
 @property (strong, readonly) SPUDownloader *download;
 @property (copy, readonly) NSString *downloadPath;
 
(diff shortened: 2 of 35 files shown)


kornelski:
Thank you.

This introduces new message which isn't translated. Could you copy the old message under the new key for other languages, so at least they get the previous text instead of English?

core-code:
thanks for the review. we can't use /just/ the existing translation since it has one placeholder less, which could create trouble. i have added the new string to all translations now and have re-used the part that was already translated for each language, so that only the new part is in english.

for german i have fully translated the new string. FYI the arabic translation seems damaged to me - whats up with these placeholders - but i am not an expert on format strings so it may be fine.

Wevah:
FWIW, the Arabic string is correct as far as I can tell; the odd-looking format tokens are due to the RTL nature of the script.

core-code:
thanks for looking into that

kornelski:
That's good. Thank you,

core-code:
thanks for merging!

Sparkle/el.lproj/Sparkle.strings
```diff
@@ -4,6 +4,8 @@
 
 "%@ %@ is currently the newest version available." = "Το %1$@ %2$@ είναι η τελευταία διαθέσιμη έκδοση.";
 
+"%@ %@ is currently the newest version available.\n(You are currently running version %@.)" = "Το %1$@ %2$@ είναι η τελευταία διαθέσιμη έκδοση.\n(Die derzeit installierte Version ist %3$@.)";
```
DivineDominion: The German translation slipped into the parenthes

DivineDominion:
@core-code Could you also add this to the XPC branch's UI updaters?

Sparkle/el.lproj/Sparkle.strings
core-code: ouch. should i send a PR?

core-code:
> @core-code Could you also add this to the XPC branch's UI updaters?

aren't improvements to the Master branch automatically merged into other branches?

kornelski:
Nope, they're not :( The branches have diverged and require manual fixups.

core-code:
i've had a look at the XPC branch but the code is so convoluted that i was unable to add the full functionality of this patch in any satisfactory or clean way. what we could do is just add the part that makes sure we display an error if we couldn't actually fetch a feed, instead of claiming 'you are up to date'.

this would amount to just changing the call to 'didNotFindUpdate' in 
SUAppcastDriver.m' to
```

else if (item) {
        [self.delegate didNotFindUpdate];
    }
    else { // Do not talk about successfully checking things, if we actually didn't do it
        NSMutableDictionary *errorUserInfo = [NSMutableDictionary dictionaryWithDictionary:@{
                                                                                        NSLocalizedDescriptionKey: SULocalizedString(@"An error occurred in retrieving update information. Please try again later.", nil),
                                                                                        }];
        
        [self.delegate didFailToFetchAppcastWithError:[NSError errorWithDomain:SUSparkleErrorDomain code:SUAppcastError userInfo:errorUserInfo]];
    }
```

Issue #1432, opened 2019-06-30
Generate Appcast with Travis-CI


yakuhzi:
Is it possible to generate the appcast with travis-ci? I want to build my app with travis and then generate the appcast and put it to Github-Releases of my repo afterwards. My problem is, that the private key of Sparkle generated with `generate_keys` is stored on my local machine only. I could add the key to the travis keychain, but `generate_appcast` would require to allow access to the keychain via a user prompt, which is not available with travis.

This is the error I get, when I try to generate the app cast without a private key:
`Warning: Private key not found in the Keychain (-25300). Please run the generate_keys tool`
`Warning: could not sign APPNAME.dmg due to lack of private EdDSA key`

I tried to generate it without the key, which works on my local machine (with the above warnings):
`Written /Users/...../appcast.xml based on 1 updates`
But within travis, the /appcast.xml is absent and therefore no appcast is generated:
`Written /Users/...../APPNAME based on 1 updates`

kornelski:
The key could be stored as a file or a short string, but currently `generate_appcast` doesn't do that.

If you'd like, you could add export feature for the key generator tool, and ability to read the key from another place to the `generate_appcast` tool.

zorgiepoo:
Note we deprecated providing the private key as an argument because it is insecure and instead have an option to pass a file. CI systems can pass the key to standard input using "-" standard input file. Check updated `--help` for more info in Sparkle 2.2 or later.

PR #1435, opened 2019-07-05
Support providing private key as argument

Yakuhzi:
Added support for providing private key as argument in order to use it with travis-ci or other ci tools ( see #1432)


kornelski:
Thank you

Issue #1414, opened 2019-06-05
Sparkle does not work as expected in long-running apps & background apps


core-code:
over the past years there have been many claims that Sparkle does not work properly in long-running apps and background apps that are launched just once, but there has never been a definitive confirmation of any issue.

i believe i can confirm that there is a bug, and can also explain the cause and offer a solution here.

the way Sparkle works is this: if it is initialised (normally at app launch), it checks if the last check has been longer ago than the "checking interval". if thats the case, it checks for updates immediately, which works fine. this is the case that is important for normal apps that are launched on-demand. however, in the other case it schedules a timer, so that the update check is done at the right moment. this is the important case for all long-running apps as its unlikely to find an update at launch, and finding updates 'while running' should also work.

scheduling a timer is done like this in Sparkle/Sparkle/SUUpdater.m

`    self.checkTimer = [NSTimer scheduledTimerWithTimeInterval:delayUntilCheck target:self selector:@selector(checkForUpdatesInBackground) userInfo:nil repeats:NO]; // Timer is non-repeating, may have invalidated itself, so we had to retain it.
`

however, this only works perfectly if the Mac never goes to sleep! 
as is commonly known and has been documented often, NSTimer does only work properly for long running timers if the Mac never goes to sleep [1]. if the Mac sleeps, the fire date is delayed for the 'amount of sleep'. in concrete terms this means, if your app has been configured to check for updates every 5 days, but the Mac sleeps 66% of the time, the update check will actually only be performed every 15 days. if your updates are quite important this can lead to a big problem as updates aren't delivered to users in a timely manner, especially on Macs that are only used a few hours per day or week.

the normal way to fix NSTimer behaviour for long-running timers despite the 'sleep problem' is to rebuild timers after sleep, e.g. by observing 'NSWorkspaceDidWakeNotification' [2]. we have used a special timer class for this for years [3].

[1]
https://www.noodlesoft.com/blog/2010/07/01/playing-with-nstimer/
https://stackoverflow.com/a/16597866/1259710
https://forums.developer.apple.com/thread/106199
http://openradar.appspot.com/13760797

[2]
technically, NSTimers are /also/ broken if the system clock is changed, which can be observed with 'NSSystemClockDidChangeNotification'. however since this is far less common than a Mac sleeping, i won't go into this here

[3]
https://github.com/core-code/CoreLib/blob/master/Mac/JMCorrectTimer.m

Wevah:
[...]

PR #1437, opened 2019-07-08
Update SUUpdater.m

core-code:
try to fix issue #1414

### DIFF
--- a/Sparkle/SUUpdater.m
+++ b/Sparkle/SUUpdater.m
@@ -37,6 +37,7 @@ NSString *const SUUpdaterAppcastNotificationKey = @"SUUpdaterAppCastNotification
 
 @interface SUUpdater () <SUUpdaterPrivate>
 @property (strong) NSTimer *checkTimer;
+@property (assign) BOOL shouldRescheduleOnWake;
 @property (strong) NSBundle *sparkleBundle;
 
 - (instancetype)initForBundle:(NSBundle *)bundle;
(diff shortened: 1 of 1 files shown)


Sparkle/SUUpdater.m
```diff
@@ -294,6 +299,24 @@ - (void)scheduleNextUpdateCheck
     self.checkTimer = [NSTimer scheduledTimerWithTimeInterval:delayUntilCheck target:self selector:@selector(checkForUpdatesInBackground) userInfo:nil repeats:NO]; // Timer is non-repeating, may have invalidated itself, so we had to retain it.
 }
 
+- (void)receiveSleepNote
+{
+    if (self.checkTimer)
+    {
+        [self.checkTimer invalidate];
```
kornelski: Is it possible that OS sends sleep, but not wake? Maybe it'd be safer to keep the timer during sleep, and reinit it in the wake event?

core-code: thats a valid suspicion given the (lack) of reliability of many macOS APIs. however, we've used similar code for 6 years without problems. also for some months we are running similar code in a fashion that would send some feedback to our server if it were ever the case that those two notifications are not triggered in a pair. not a single hiccup to date. however, i can try to re-work the code to rely solely on the wake-note if you wish.

kornelski: If you say it's a well-tested method, let's roll with it.

PR #1443, opened 2019-08-02
In 'newer than latest' alert, show displayVersionString instead of versionString

gotow:


kornelski:
Thank you

gotow:
Geez - thank _you_ for contributing to and maintaining Sparkle! I really appreciate it!

 - Jon

Issue #1442, opened 2019-08-01
What is the recommended way to change an app name, including the path of `.app`, when updating?


jfenton-tsc:
I'm running into trouble figuring out how to set our app up for a name change--mostly because I'm misunderstanding things, I think. In short, we want to push an update where the app is renamed from `Old App Name.app` to `Name App Name.app`, in the Applications folder.

I've looked at the following issues to try and make this behavior work: 

#400
#469
#782

 I'll try to run down how we have things set up, the behavior we want, and the behavior we're seeing instead:

### Description of the problem

**The current version is configured like this:**
filename: `Old App Name.app`
bundle name: `Old App Name`
bundle identifier: `com.company.oldappname`

in the `Info.plist` for the app, we have:
```
<key>SUBundleName</key>
<string>New App Name</string>
```
**NOTE:** prior to adding this `SUBundleName` to the plist, Sparkle emitted an error where, during extraction of the update, it couldn't find a `.app` or `.pkg` of the expected name; (that being `Old App Name.app`). It seems `SUBundleName` has fixed this.

in the `xcconfig` for Sparkle, we have:

`SPARKLE_NORMALIZE_INSTALLED_APPLICATION_NAME = 1`

**The 'update' is configured like this:**
filename: `New App Name.app`
bundle name: `New App Name`
bundle identifier: `com.company.oldappname`

**NOTE:** we found changing the bundle _identifier_ broke Sparkle's codesign verification, so we're just... not going to change it.

**The behavior we expect:**
1. currently installed `Old App Name.app` detects an update. It either prompts the user to install or runs automatically--however it's configured.
1. it downloads the DMG (or zip) specified in its appcast, then extracts files from it.
1. it looks for a `.app` or `.pkg` that _matches_ `SUBundleName`, in this case `New App Name.app`
1. it copies that .app to a temporary folder, then copies it over the currently-installed app and restarts. 

This is where what I expect and what happens differ:

Because `SPARKLE_NORMALIZE_INSTALLED_APPLICATION_NAME=1` _and_ `SUBundleName` are both set, I would _expect_ the installed `Old App Name.app` to be renamed to `New App Name.app`--instead, it's left as `Old App Name.app`, or set _back_ to `Old App Name.app` if it had been changed previously.

### Do you use Sandboxing in your app?

No.

### Version of `Sparkle.framework` in the latest version of your app

1.20.0

### Version of `Sparkle.framework` in the old version of app that your users have (or N/A)

1.20.0

[...]

PR #1444, opened 2019-08-02
check if SUBundleName is set before normalizing

jfenton-tsc:
Inspired by https://github.com/sparkle-project/Sparkle/issues/1442

When changing the name of an app, there's a couple things to do in order
to ensure auto-updates still work:

1. set SUBundle Name. This basically tells Sparkle "ey, expect an app
of {SUBundleName}.(app|pkg) instead of the normal name."
2. set `SPARKLE_NORMALIZE_INSTALLED_APPLICATION_NAME` in
commonConfig.xccconfig to reset the name of the app.

in (2), we always used the host's bundle name, instead of checking to
see if SUBundleName was set, which means normalization doesn't work as
expected during app name changes.

A problem that arises is that we will ALWAYS normalize the installed `.app` path to `SUBundleName`--even if the thing you're updating to has a _different_ bundle name from `SUBundleName`. 

I think that's expected behavior, if you specify `SUBundleName` and `Normalize`, but it does seem odd.

## Testing I've done

made `Some App Name.app` update to `Different App Name.app`
- set `SUBundleName` to `Different App Name`
- compiled with `SPARKLE_NORMALIZE_INSTALLED_APPLICATION_NAME = 1`
- saw `Some App Name.app` become `Different App Name.app` when installing the update

made `Some App Name.app` update to a later build of `Some App Name.app`
- set `SUBundleName` to `Different App Name`
- compiled with `SPARKLE_NORMALIZE_INSTALLED_APPLICATION_NAME = 1`
- saw `Some App Name.app` become `Different App Name.app` when installing the update, even though it's really installing `Some App Name`

## Other To-Dos?
- Do docs need to be updated, since this is a behavior change?
  - I think `SPARKLE_NORMALIZE_INSTALLED_APPLICATION_NAME` should be mentioned a bit more prominently--I had to do some digging to find out it was an option for the behavior I want
- Soliciting other users' feedback
  - it's possible (probably) that this behavior change will have an effect on others that are using it.

### DIFF
--- a/Sparkle/SUInstaller.m
+++ b/Sparkle/SUInstaller.m
@@ -152,8 +152,14 @@
 {
     NSBundle *bundle = host.bundle;
     assert(bundle != nil);
-    
-    NSString *normalizedAppPath = [[[bundle bundlePath] stringByDeletingLastPathComponent] stringByAppendingPathComponent:[NSString stringWithFormat:@"%@.%@", [host objectForInfoDictionaryKey:(__bridge NSString *)kCFBundleNameKey], [[bundle bundlePath] pathExtension]]];
+   
+   NSString * baseBundleName = [host objectForInfoDictionaryKey:@"SUBundleName"];
+   
+   if (baseBundleName == nil) {
+      baseBundleName = [host objectForInfoDictionaryKey:(__bridge NSString *)kCFBundleNameKey];
+   }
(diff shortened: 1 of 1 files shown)


kornelski:
This looks good. Thank you.

Issue #1449, opened 2019-08-07
[ui-separation-and-xpc] Cancelled downloads continue on their own


mangerlahn:
I am implementing my own front end to the Sparkle Updater using the new `SPUUserDriver`. 

During development I realised that the updating process does not stop when one of the cancellation handlers is called. 

I am not sure if this is intended or not: 
The SPUDownloader lives on after it has been set to nil by the download driver. The url session still has a reference to the downloader. Since the download driver is still set as the delegate of the downloader, it will continue to receive download updates even if its own reference to the downloader is nil. 

I created a new branch that fixes this behaviour by basically exposing the downloaders cleanup method so that it can be called by the download driver. Before I open a PR about this I just wanted to make sure that this is a bug and not intended. I could imagine that the download continues to get some sort of prefetching behaviour or something. Thats why I am asking :)

kornelski:
Yes, it would be nice to cancel.

mangerlahn:
Closing since the associated PR was merged.

PR #1451, opened 2019-08-08
#1449 Fix ongoing download after the update has been cancelled

mangerlahn:
This is a PR to the download cancellation bug explained in #1449 .


kornelski:
Thank you

PR #1454, opened 2019-08-13
SUFeedUrl generate appcast file extension

Neoster:
When the SUFeedURL extension is "php", replace filename.php with appcast.xml in generate_appcast


kornelski:
Thanks for the PR. Please use deleteLastPathComponent and appendPathComponent instead of find'n'replace.

Neoster:
@kornelski I did process your requested changes.

kornelski:
The '*ing` methods don't modify the variable, but make a new value, which you're not using. So this code doesn't do anything now.

Neoster:
Ok, I am sorry about that. I am not familiar with the Sparkle ecosystem yet, so I will figure out a way to test this. Will keep you posted.

kornelski:
Thank you

PR #1477, opened 2019-09-29
ui-separation-and-xpc branch fixes

ksuther:
- bsdiff wasn't an explicit dependency for the Autoupdater target
- Added explicit casts to silence Xcode 11 warnings


kornelski:
Thanks

PR #1479, opened 2019-10-02
Added bsdiff as an explicit dependency to targets

ksuther:
Added bsdiff as an explicit dependency to BinaryDelta and Sparkle Unit Tests targets
Fixes build failures on Xcode 11
Same fix as in #1477 for other targets


kornelski:
Thanks

PR #1480, opened 2019-10-04
Hide "Skip..." and "Remind..." buttons.

kennethjohnsen:
Added possibility of setting SUHideSkipButton and SUHideRemindButton.
This is useful when automatic updates are disabled, since the user will not be reminded anyway until he/she manually starts an update-check, so the buttons may confuse the user.


kornelski:
We've discussed this for critical updates, which is a specific context.

I see this solution solves the problem for you, but for Sparkle in general it's less beneficial, because:

* Optional settings means existing Sparkle users don't get this functionality unless they implement it themselves,

* Implementing it means more things in documentation, more work to integrate with Sparkle.


So while I agree with the problems you're trying to solve, I'm not convinced that such low-level specific config is the right solution.

I'd prefer the skip button to be hidden automatically for critical updates.

It's also a great catch that "remind me later" doesn't make sense for manual update checks. But in this case it also makes sense to just hide it automatically, so that all Sparkle users benefit from it out of the box, and we don't have to add documentation or extra integration work for them.

kennethjohnsen:
I changed the code accordingly.
Let me know if it meets the discussed requirements.

kornelski:
Great! Thanks.

mickaelhero:
Hi how to get this feature ? The feature exist when writing pod 'Sparkle' in Podfile? And how to set critical update?

kornelski:
You can't, and probably shouldn't, because it's a property of the update, not a decision of the old app.

nicolasbosi95:
In the fix, I see the value for "SUEnableAutomaticChecks" is being obtained by calling 

`NSNumber *automaticChecksEnabled = [self.host objectForInfoDictionaryKey:SUEnableAutomaticChecksKey]`

And with this, only the value in Info.plist for this key will be taken into account. Visibility of the button cannot be changed in runtime, since calls to `-[SUUpdater setAutomaticChecksEnabled:(BOOL)]` are saved to UserDefaults.

Which makes me wonder, is that the intended behavior?
If it's not, then why not do:

`NSNumber *automaticChecksEnabled = [self.host objectForKey:SUEnableAutomaticChecksKey];`

This way, the priority will be for the value in UserDefaults first. If it does not exist there, check InfoDictionary.
I find this alternative compliant with the [documentation](https://sparkle-project.org/documentation/customization/):

> [...] Setting to NO disables update checks, but can be overridden by a call to SUUpdater’s setAutomaticallyChecksForUpdates:

kornelski:
Yeah, it should have used `host objectForKey`

PR #1483, opened 2019-10-08
fix public key doesn't match problem

qwertyyb:
fix public key doesn't match problem when use generate_appcast -s <key_str> <folder>


qwertyyb:
there's a bug when exec generate_appcast, It always get error that the public key doesn't match and this will fix it

kornelski:
Thanks

PR #1484, opened 2019-10-08
Terminology Update for BR

vitu:
In Catalina, Apple is introducing some terminology changes for BR.

Sparkle is only affected by “Download”, which is now either “Download” — when used as a noun — or “Baixar” (plus variants) — when used as a verb.

Also, the word “app”, though not present in source, is being used more in target to avoid potential issues with application gender and incorrect grammar.


kornelski:
Thanks

PR #1486, opened 2019-10-14
SPM File Structure

liamrosenfeld:
This organizes the project into the file structure required for Swift Package Manager. It is a first step towards full SPM support.


liamrosenfeld:
I'm making progress on a full SPM package, just it's rather slow because of a general lack of resources on the topic (especially for objc packages) .

1024jp:
SwiftPM is great. It can even mix obj-c with Swift. However,  I suppose it is not suitable for Sparkle.framework because, as I know so far, SwiftPM doesn't support including resource files. This specification can be changed in the future ( cf. [\[SwiftPM\] Need a way to include resources with targets](https://bugs.swift.org/browse/SR-2866)), but you cannot include resource files such as xib or other graphical files at the moment at least.

kornelski:
If we can't produce a Swift package currently, I'd rather not reorganize the repository for it. Internal reorg has already been problematic for merging Sparkle v1 and v2, and I'd rather not make it worse.

liamrosenfeld:
Yeah, that makes sense. The [Swift Forum Thread](https://forums.swift.org/t/swift-pm-bundles-and-resources/13981/28) for bundle resources is slowly moving along, and it would make sense to hold off until that comes to a conclusion.

WowbaggersLiquidLunch:
The Swift Evolution [proposal](https://github.com/apple/swift-evolution/blob/master/proposals/0271-package-manager-resources.md) that allows Swift Package Manager to include resources with targets has been [accepted](https://forums.swift.org/t/se-0271-package-manager-resources/30730/63), and will be [available in Swift 5.3](https://forums.swift.org/t/se-0271-package-manager-resources/30730/74). This should pave way for full SPM support.

PR #1494, opened 2019-10-25
Let SPUUpdaterDelegate handle errors currently suppressed

martinjankoehler:
When using a custom ```SPUUpdaterDelegate```, for example to display updater status information in the preference window, we've noticed that errors are lost which occur during scheduled update checks.

Steps to reproduce:
1.  Use ```Sparkle.framework``` of  ```ui-separation-and-xpc``` 67819be18a4ef48e85ea30dbbf8de582f5ef405c
2. Set ```SUFeedURL``` to ```wrongscheme://...```
3. *ASSERT:* ```-[SPUUpdaterDelegate updater:didAbortWithError:]``` is called

Step 3. currently fails, because ```-[SPUScheduledUpdateDriver basicDriverIsRequestingAbortUpdateWithError:]``` and ```-[SPUScheduledUpdateDriver coreDriverIsRequestingAbortUpdateWithError:]``` are swallowing the ```error``` by passing ```nil``` instead

### DIFF
--- a/Sparkle/SPUScheduledUpdateDriver.m
+++ b/Sparkle/SPUScheduledUpdateDriver.m
@@ -83,8 +83,8 @@
 - (void)notifyDelegateAboutError:(nullable NSError *)error
 {
     if (error == nil) { return; }
-    if ([self.updaterDelegate respondsToSelector:@selector(updater:didAbortWithError:)]) {
-        [self.updaterDelegate updater:self.updater didAbortWithError:(NSError * _Nonnull)error];
+    if ([self.updaterDelegate respondsToSelector:@selector(updater:scheduledUpdateCheckDidAbortWithError:)]) {
+        [self.updaterDelegate updater:self.updater scheduledUpdateCheckDidAbortWithError:(NSError * _Nonnull)error];
     }
 }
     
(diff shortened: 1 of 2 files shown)


kornelski:
Thanks for the PR.

The intention is to avoid surprising users with error popups for actions they didn't explicitly initiate.

Your UI may handle this gracefully, but if someone else implemented their delegate with an alert box, this change would be very disruptive.

Perhaps a better fix would be to make it an explicit flag, and change:

```objc
[self abortUpdateWithError:nil];
```

to something like:

```objc
[self abortUpdateWithError:err showUI:NO];
```

and pass on that flag to the delegate.

martinjankoehler:
@kornelski Thanks for the feedback! 
Not sure what's better, but another option would perhaps be to add a separate delegate method which makes the context explicit, like 
```[self.updaterDelegate updater:self.updater scheduledCheckDidAbortWithError:error];```
That way, no existing delegates would have to be changed to handle the ```showUI``` argument.

martinjankoehler:
@kornelski I've fixed this in 1c89c7d

kornelski:
Thanks

PR #1499, opened 2019-11-12
Support generating appcast with localizations

### LINKED ISSUE #1205: generate_appcast should support localized Release Notes
For an app archive called **myApp.1.10.2.zip generate_appcast** will automatically add release notes if there is a file named **myApp.1.10.2.html**. This is great but doesn’t really help if you use localized release notes.

It would be great if **generate_appcast** would also add localized release notes (German in my example) sticking to the naming convention **myApp.1.10.2.de.html** (or maybe **myApp.1.10.2.html.de**?).

al1k0s:
Previously you would need to use you own solution to add the localized release notes into the appcast but not today.
Fixes #1205


kornelski:
That's cool. Thanks!

PR #1504, opened 2019-11-22
feat: properly compare versions with commit SHA

aleclarson:
When two versions have the same major/minor/patch numbers (eg: `1.5.5-335d3e2` and `1.5.5-a655360`), the newer version is used (determined by its `date` property).

### DIFF
--- a/Sparkle/SUBasicUpdateDriver.m
+++ b/Sparkle/SUBasicUpdateDriver.m
@@ -119,7 +119,12 @@
     SUAppcastItem *item = nil;
     for(SUAppcastItem *candidate in appcastItems) {
         if ([self hostSupportsItem:candidate]) {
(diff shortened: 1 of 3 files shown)


kode54:
Yeah, this commit breaks version comparison in my project, because my versions are based on a Git tag:

tagname-commitcount-ghash

kornelski:
I'm surprised that your scheme has worked before. I don't think Sparkle should be adding support for more complicated versions. Your case of having the same version for all releases is an edge case. You should be increasing the version, not just commit count.

kode54:
How about I just increment a decimal point by one, per every published release, forever more? Is that sufficient?

kornelski:
Yes, if you keep incrementing the "patch" part of the version, it'll be fine. These are 3 numbers, not a decimal fraction.

kode54:
I changed my versioning scheme, and it works out fine for Sparkle 1.23.0. The only issues I've had so far now are with my own maintenance scripts.

PR #1512, opened 2019-12-11
Update SUPlainInstaller.m

core-code:
Try to preserve Finder tags while updating apps


kornelski:
Thanks!

core-code:
sorry to followup here but this code requires 10.9 but i just see now that you still support 10.7 (wow).

should i send another PR that wraps this in an @available block?

kornelski:
Oh yes, wrapping it in a check would be good.

PR #1516, opened 2019-12-12
Fix second bold sentence in read-only update alert dialog. (#15…

quinntaylor:
• Updated error dialog code in SUBasicUpdateDriver, and created a local variable to reuse `[aHost name]`.
• Updated content of Sparkle.strings files, splitting existing affected translations into 2 chunks each.
• Changed "Use Finder to copy" to back to "Copy", since most translations still use that terminology.
• Fixed the Catalan translation, which was using an old string key.
• Removed a few unnecessary comments in the Polish translation.

### DIFF
--- a/Sparkle/SUBasicUpdateDriver.m
+++ b/Sparkle/SUBasicUpdateDriver.m
@@ -65,13 +65,14 @@
     [super checkForUpdatesAtURL:URL host:aHost];
 	if ([aHost isRunningOnReadOnlyVolume])
 	{
+        NSString *hostName = [aHost name];
         if ([aHost isRunningTranslocated])
         {
-            [self abortUpdateWithError:[NSError errorWithDomain:SUSparkleErrorDomain code:SURunningTranslocated userInfo:@{ NSLocalizedDescriptionKey: [NSString stringWithFormat:SULocalizedString(@"Quit %1$@, move it into your Applications folder, relaunch it from there and try again. %2$@ can’t be updated if it’s running from the location it was downloaded to.", nil), [aHost name], [aHost name]] }]];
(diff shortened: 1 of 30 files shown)


kornelski:
Thank you for the PR. It's a good idea to improve formatting of this error.

Sparkle/SUBasicUpdateDriver.m
```diff
@@ -3,7 +3,7 @@
 //  Sparkle
 //
 //  Created by Andy Matuschak on 4/23/08.
-//  Copyright 2008 Andy Matuschak. All rights reserved.
```
kornelski: Please don't remove copyright attribution. That's not legal.

Sparkle/SUBasicUpdateDriver.m
```diff
@@ -65,13 +65,14 @@ - (void)checkForUpdatesAtURL:(NSURL *)URL host:(SUHost *)aHost
...
+            [self abortUpdateWithError:[NSError errorWithDomain:SUSparkleErrorDomain code:SURunningTranslocated userInfo:@{ NSLocalizedRecoverySuggestionErrorKey: [NSString stringWithFormat:SULocalizedString(@"Quit %1$@, move it into your Applications folder, relaunch it from there and try again.", nil), hostName], NSLocalizedDescriptionKey: [NSString stringWithFormat:SULocalizedString(@"%1$@ can’t be updated if it’s running from the location it was downloaded to.", nil), hostName], }]];
         }
         else
         {
-            [self abortUpdateWithError:[NSError errorWithDomain:SUSparkleErrorDomain code:SURunningFromDiskImageError userInfo:@{ NSLocalizedDescriptionKey: [NSString stringWithFormat:SULocalizedString(@"%1$@ can't be updated, because it was opened from a read-only or a temporary location. Use Finder to copy %1$@ to the Applications folder, relaunch it from there, and try again.", nil), [aHost name]] }]];
```
kornelski: This change is backwards. "Use Finder to copy" is the new, intended version. Finder very specifically is required here, because e.g. copying from command line or a 3rd party tool like PathFinder doesn't achieve the same effect, and we've been getting bug reports about this.

Sparkle/SUBasicUpdateDriver.m
quinntaylor: Thanks, I'm used to updating copyrights in internal-only code. I'll revert this change and leave it untouched for now.

Sparkle/SUBasicUpdateDriver.m
quinntaylor: Ah, it makes sense that the translations would lag behind the localizable string in English. I've changed the key back to the newer value.

kornelski:
Thank you

PR #1519, opened 2019-12-14
Separate issue templates for Sparkle 1.x and 2.x

jakob:
Create separate issue templates for Sparkle 1.x and 2.x.

The intention is to let users select the version of Sparkle they are using, and to have the `1.x` and `2.x` labels  assigned automatically.


kornelski:
Thank you

Issue #1505, opened 2019-11-23
SUUpdater shim aborts when no SUFeedURL is present when instantiated


johnwbyrd:
At some expense, I have discovered that SUUpdater shim in ui-separation-and-xpc silently fails in release mode on versions of MacOS 10.13 and earlier (but not 10.14), if the feed URL is not set when the SUUpdater object is created.

Previous SUUpdater behavior was to defer setting and testing the first update until requested via checkForUpdatesInBackup.  For situations where the SUFeedURL is set dynamically, the current implementation of requiring it to go in Info.plist is insufficient.

In release mode, Sparkle should NEVER call abort() due to run-time conditions, which it does all over the place.  Sparkle is merely update software.  The worst it should ever do at a client site is fail silently.  In release mode, please log errors instead of just dying on the main thread.

jakob:
I've tried to fix the issue by delaying the updater startup just like in Sparkle 1.x. I can no longer reproduce the crash, even when the feed URL is set in code.

Can you verify if the issue is fixed now?

I did not remove the abort() call, that's a separate issue.

zorgiepoo:
Resolved at some point.

PR #1524, opened 2019-12-14
Sparkle 2.x: Delay startup of updater

jakob:
This PR tries to address the problem reported in #1505

It delays the startup of the updater when using the SUUpdater compatibility class, so that the host has a chance to configure the updater before it is started (eg. in willFinishLaunching:)

I've not made any attempts to fix the `abort()` calls, those are a separate issue.


kornelski:
Thanks. 

That's a subtle timing issue :(

PR #1538, opened 2020-01-23
Remove .h files from Build Rules for bsdiff and ed25519

pteeson:
On branch VendorBuildPhasesFix
Changes to be committed:
	modified:   Sparkle.xcodeproj/project.pbxproj

Building the Sparkle framework produced 10 Warnings caused by .h files
being added to the Build Rules Compile Sources. Thanks to tip from
Stackoverflow https://stackoverflow.com/questions/6509600/

"Click on your project, and check that this file is not present in the
tab Build Phases. Normally no header files should stay here.
Clean and build it again, it should work!"

It does work.


kornelski:
Thanks. If I'm reading the project file correctly, it removes .h files from the project completely, not just from build phases. Is that so?

pteeson:
So far as I can see no they are not removed. If they were I thinkJenkins would have not completed the checks.

What I did in Xcode was to 
(1) go to the bsdiff Target and 
(2) select the Build Phases Tab
(3) then the Compile Sources and click on the grey triangle to reveal the items
(4) then Command click select all the header files
(5) and click on the minus sign.

This removes them from the compilation step but leaves them in the Vendor/bsdiff folder
Ditto for the ed25519 Target.

Then did a project clean and build and those 10 warnings were no longer given. 
Also the project built OK. In particular the Sparkle Framework.

Finally I did the git stuff: commit, push to origin (a.k.a. my Sparkle fork), and created a pull in upstream (a.k.a. the sparkle-project-Sparkle).

Using Terminal I looked at my origin  and the headers are still there.
And Xcode also shows them there.

So as far as I can see things worked as desired.

P.S. There are some other Warnings that I will look into but that’s for a different thread….

kornelski:
Thanks

PR #1545, opened 2020-02-06
Remove most production-reachable instances of abort().

bdkjones:
A first stab at resolving issue #1510. 

I'm sure there will be things to tweak and if this approach seems sensible, the user-facing strings in the `NSAlert` windows will need to be localized. 

I used `NSException` for the programmatic errors that should immediately crash with extreme prejudice. In some spots I just removed the `abort()` and in others I relaxed the crash to a no-op with suitable console logging.

### DIFF
--- a/Sparkle/SPUAutomaticUpdateDriver.m
+++ b/Sparkle/SPUAutomaticUpdateDriver.m
@@ -69,14 +69,12 @@
 {
     // Nothing really to do here.. this shouldn't be called.
     SULog(SULogLevelError, @"Error: resumeInstallingUpdateWithCompletion: called on SPUAutomaticUpdateDriver");
-    abort();
 }
 
 - (void)resumeUpdate:(id<SPUResumableUpdate>)__unused resumableUpdate completion:(SPUUpdateDriverCompletion)__unused completionBlock __attribute__((noreturn))
 {
     // Nothing really to do here.. this shouldn't be called.
     SULog(SULogLevelError, @"Error: resumeDownloadedUpdate:completion: called on SPUAutomaticUpdateDriver");
-    abort();
 }
 
 - (void)basicDriverDidFindUpdateWithAppcastItem:(SUAppcastItem *)updateItem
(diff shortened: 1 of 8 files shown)


bdkjones:
Also, as I mentioned in #1510, I don't really understand the codebase well enough to know what's going on with the `abort()` calls when the user clicks "Skip This Update" and so forth. I'm not sure if those are there because those switch cases should be unreachable, etc. 

I'd need to spend more time tracing what's going on before I could competently mess with those.

bdkjones:
I also glanced at the build settings and it looks like ARC is used everywhere (there are a lot of targets; I didn't go through them all). If certain files are still manually memory-managed, please advise.

kornelski:
If there's manual memory management, it'd be good to move it to ARC. Sparkle is older than ARC, so there's probably no big reason why they're not ARC, they may just be older parts of the codebase.

kornelski:
I'd prefer something noisier than just log and nils. How about

```
#if DEBUG
abort()
#endif
```


or replace abort with a `make_some_noise(msg)` function that throws alerts or something else depending on debug/release?

bdkjones:
Sure, I’m open to anything. I just didn’t see many other uses of compiler macros throughout the codebase, so I wasn’t sure they were welcome. Also, I didn’t think they’d have any effect for most Sparkle users because the framework is pre-built for release before it’s incorporated by app authors. They’d have to build a debug release of Sparkle and use that, no?

kornelski:
Oh, prebuilt Sparkle will be a problem indeed. In that case I don't know how to distinguish between "runs on developer's machine" vs "runs on users' machines". I think the distinction is important, because on users' machines it should try to update whenever it can and stay quiet about problems users can't fix. But such behavior on dev's machine may give them impression that incorrectly integrated Sparkle is working.

bdkjones:
It’s *possible* to see if a debugger is attached by looking at ptrace flags and mach ports for exception-handling, but I don’t think it’s worth it. If a debugger is attached, you could assume it’s running on the developer’s machine and throw exceptions accordingly. But there’s a small runtime penalty for the checks and it’s complex and somewhat fragile. I also don’t think it’s needed because developers will extensively test Sparkle when implementing it in their apps, so they’ll catch stuff like calling the wrong method, etc. I highly doubt any developers ever use a debug build of Sparkle—they simply drop the release framework in their project and start integrating it.

PR #1566, opened 2020-03-26
Better Appcast Errors

bdkjones:
While implementing a custom UI for my updater, I found that Sparkle reported a very generic error message when the Internet connection was offline:

    An error occurred while checking for updates. Please try again later.

In the debugger, I could see Sparkle log an `NSURL` error with a more specific localized description: `The Internet connection appears to be offline`. However, that error was not available to me once Sparkle got to notifying my `SPUUpdateDriver` about it.

This pull request removes an extra layer of `UserInfo` Dictionary while reporting AppCast errors and exposes the underlying `NSError` from an appcast failure so that implementors can get at it and provide more descriptive text to users, if they choose to do so.

This file is managed by ARC, so there are no memory leaks from a lack of `release`.

### DIFF
--- a/Sparkle/SUAppcast.m
+++ b/Sparkle/SUAppcast.m
@@ -105,9 +105,14 @@
         } else {
             SULog(SULogLevelError, @"Encountered download feed error: %@", error);
             
-            NSDictionary *userInfo = [NSDictionary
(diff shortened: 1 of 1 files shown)


kornelski:
Thanks

PR #1582, opened 2020-05-02
Use the hash of the archive for extraction when generating an appcast

### LINKED ISSUE #1559: generate_appcast should invalidate cache better
Currently generate_appcast extracts bundles to a folder per version, but that breaks it when developers try to re-release the same version. 

Instead, it should use hash of the archive file as the unarchiving folder name, so that any change to the archive updates the appcast.

Wevah:
Fixes #1559.

Notes: I'm not sure if 64 K is a good read buffer size or if larger would be better.


kornelski:
Very helpful. Thanks!

PR #1584, opened 2020-05-05
Fix warnings from Xcode 11.4 and recent commits

ksuther:
Fixes format warning, noreturn warning, and framework @import warning.


kornelski:
Thank you

PR #1597, opened 2020-05-26
Fix CI tests

rajivshah3:
The CI tests currently seem to fail silently (or not run at all). The new CI script runs most of the unit tests successfully, but a few are still failing

- Testing on Xcode 10.3 and 11.4
- Removed unnecessary `brew update`
- Removed old code coverage configs
- Set Swift Version to 4.2 in top level config
- Fix `SUCodeSigningVerifierTest.m` on macOS 10.15 (see df718b0)


kornelski:
Thanks!

There's still a failure. I presume it's just a timeout? Can you bump the time limit on it?

rajivshah3:
Yep, will do

rajivshah3:
@kornelski I think the `testDeprecatedDownloader` test might be failing because it uses `NSURLDownload initWithRequest:delegate:` which is deprecated. The [docs](https://developer.apple.com/documentation/foundation/nsurldownload/1416302-initwithrequest?language=objc) say macOS 10.3–10.11, so does that mean it's not available on 10.15?

PR #1607, opened 2020-06-09
Do not check for installation type (to match with 1.x implementation).

nagasunilt:


kornelski:
@zorgiepoo Does that have security implications?

zorgiepoo:
This won't work reliably. 2.x needs to know the installation type (app vs guided pkg vs interactive pkg) ahead of time to determine whether or not the installer needs to be launched as root or not. See SUInstallerLauncher.m. The installer just verifies it wasn't lied to. 1.x does not have such an isolated helper tool that might be privileged.

zorgiepoo:
I think one proper way of making the default nicer is allowing pkg's to be delivered without archiving/zipping them and having a flat pkg "unarchiver" type. In this case, Sparkle would assume a .pkg file installation type is guided. Regarding 'trusting' the 'archive' filename, passing it to /usr/sbin/installer should probably be fine(?) (interactive pkgs probably shouldn't be implicitly trusted by 'archive' filename; they should still need to be archived and installation type specified).

The above likely conflicts with ed25519 PR, which should really be a higher priority getting merged first.

Barring a proper alternative solution that makes Sparkle aware of the installation type, you need to explicitly provide the installation type if you aren't using the regular (and most recommended) app one. So 2.x doesn't do the wrong thing and cause installation failures or oddities in some user cases.

nagasunilt:
@zorgiepoo Thanks for the response. I'm using Microsoft AppCenter to store/distribute our pkg installer. AppCenter generates the appcast.xml file and unfortunately `sparkle:installationType` is not part of the appcast file it generates and there is no way for me explicitly tell AppCenter to add the installation type.

I've created a ticket with AppCenter support team to resolve above and meanwhile, want to see if there are any other alternatives to make Sparkle updates work for my pkg installer with 2.x branch?

zorgiepoo:
My opinions are one of:

1. Don't use pkg =p
2. Fork like you were considering for your own purposes but modify the correct calling code as noted above, not the safeguard you tried to remove.
3. Help contribute to have sparkle to support downloading non-archived pkg's that must be guided (not interactive). There's an open PR that may need to be cleaned up and may need to address security concerns.

This check removed here is a safeguard, not a security a check. As indicated in the comments and documentation.

nagasunilt:
Thanks @zorgiepoo for the prompt response. My notes below:

> Don't use pkg =p (especially if you don't have a good reason to)
Our macOS app has daemon included with it so have to use pkg (would love to hear if there are alternatives other than using pkg in this case).

> Fork like you were considering for your own purposes...
I will most likely go this route (forking off 2.x branch) for time-being but can you please elaborate on "modify the correct calling code as noted above" part as I don't really see where it was suggested in earlier comments. Which calling code needs to be modified and to what?

> Help contribute to have sparkle to support downloading non-archived pkg's...
Agreed and will definitely contribute on this support.

kornelski:
You can put your daemon's executable inside your main app bundle's resources. Use Sparkle delegate to stop the daemon before update. Check on app (re)start whether it needs to (re)install the daemon.

nagasunilt:
We want the daemon to run even when app is not running/launched. Also, daemon needs root access for its normal operation so the user will be prompted for root password every time app tries to start the daemon.

We currently make our daemon as launch daemon and create a plist entry during pkg installation so daemon is always running irrespective of app state.

kornelski:
* Installer prompts for root password anyway.
* You can create the same plist entries during app's first startup.

zorgiepoo:
Yeah as Kornel says, it doesn't look like you have a good reason to use a pkg. There's an "official" API for installing and updating a daemon securely called SMJobBless you may want to check out. I believe [this](https://developer.apple.com/library/archive/samplecode/EvenBetterAuthorizationSample/Introduction/Intro.html#//apple_ref/doc/uid/DTS40013768-Intro-DontLinkElementID_2) is the latest sample project for it but may want to double check.

I would likely not tie Sparkle into your launch daemon as users can update your app manually.

nagasunilt:
From what I read ([here](https://stackoverflow.com/questions/32980194/sandboxed-app-with-helper-app) and [here](https://developer.apple.com/library/archive/documentation/Security/Conceptual/AppSandboxDesignGuide/DesigningYourSandbox/DesigningYourSandbox.html#//apple_ref/doc/uid/TP40011183-CH4-SW6)), SMJobBless would not work with sandboxed apps so don't think I can go that route for our sandboxed app.

Also for Kornel's suggestion, I believe this is how it would look like (if I understood it correctly :p). I'm not bringing in Sparkle update process into the workflow just yet as users could update manually as well. 

1. Distribute the app as .app or .dmg file.
2. Have users move .app to /Applications folder (requires root password).
3. Upon first start, app creates plist entries in /Library/LaunchDaemon and starts the daemon (requires root password).
4. For future updates, user would do (2). App has to reload the new daemon binary so needs root password.

If above workflow is how it would look like, user would need to enter root password twice during fresh install and updates. Moreover, I might be missing something with using pkg installer but what is the downside of using it in our case?

Also, could you guys please suggest minimalistic changes that are needed for my private fork to make pkg installer updates work without the presence of `installationType` tag?

zorgiepoo:
> From what I read (here and here), SMJobBless would not work with sandboxed apps so don't think I can go that route for our sandboxed app.

It can. You use an XPC service which is not sandboxed, to call SMJobBless. I believe the SMJobBless sample app explains this. 2.x branch works this way when you used it in a sandboxed app.

> Have users move .app to /Applications folder (requires root password).
This doesn't require root for the majority of users.

> Upon first start, app creates plist entries in /Library/LaunchDaemon and starts the daemon (requires root password).
For a sandboxed app, you'd need an unsandboxed XPC service, same as above.

> Also, could you guys please suggest minimalistic changes that are needed for my private fork to make pkg installer updates work without the presence of installationType tag?

I don't have the source in front of me at this right moment but the concept is you need to convince the SUAppcastItem the installationType is really a guided pkg install type. Grep for where/how the installationType is passed/used before the installer is launched. You can do what you initially suggested for yourself if you don't mind for some edge cases like the user copying/moving the app in such a way it doesn't need root privileges to write to and the installer failing in that case.

nagasunilt:
> I don't have the source in front of me at this right moment but the concept is you need to convince the SUAppcastItem the installationType is really a guided pkg install type. Grep for where/how the installationType is passed/used before the installer is launched. You can do what you initially suggested for yourself if you don't mind for some edge cases like the user copying/moving the app in such a way it doesn't need root privileges to write to and the installer failing in that case.

Thanks for the suggestions @zorgiepoo. SUAppcastItem object while initializing looks for "sparkle:installationType" in enclosure and defaults to SPUInstallationTypeApplication type if installationType tag does not exist. I changed the default type to be SPUInstallationTypeGuidedPackage if the tag does not exist in my private repo and it works well with updates without any errors.

Do you think of any functional or security related issues using this approach? I plan to use this private repo until AppCenter allows me to set "sparkle:installationType" in appcast xml.

Changes here: https://github.com/sparkle-project/Sparkle/compare/2.x...nagasunilt:default-pkg-type 

Thanks for all your help so far.

nagasunilt:
Hello @zorgiepoo, can you please take a look at the changes from above comment and let me know if it is safe to set default installationType to guided package until I have the ability to modify the appcast xml file.

zorgiepoo:
Changing the default makes sense for your case. Just one note is if you do want to ship an app instead in the future you'll have to either specify installation type in the appcast or revert that relying on app default.

nagasunilt:
Got it. Thanks for all your help. Very much appreciate it.

Issue #1583, opened 2020-05-03
On hold: Update swift helpers to use Swift Argument Parser


Wevah:
The fairly recent [Swift Argument Parser](https://github.com/apple/swift-argument-parser) makes parsing arguments a lot more simple and straightforward.

(I'll take this issue.)

Wevah:
I just realized this might require Swift 5.0 so maybe it should be put off for now.

Wevah:
Is there a reason to stay with Swift 4.2? Making sure the tools compile on older Xcodes?

kornelski:
Personally I don't care about old Swift versions. I'm not sure if some users aren't stuck with older versions of Xcode if they require older SDKs or something.

Wevah:
I’ll start a branch for this so it’s in reserve.

Wevah:
The distributed binaries are currently distributed without the Swift runtime embedded, so it's likely ok to bump the Swift version.

Wevah:
The concern I have is that the argument parser package compiles to something like 2 MB; I'm not sure if keeping the binary size of the CLI tools down is a priority or not.

(I also need to rebase.)

kornelski:
Size of the CLI tools isn't a very big concern. They're already embarassingly large due to Swift runtime :)

Wevah:
Oh yeah, I forgot they still need to target OSs < 10.14. ¬_¬

PR #1608, opened 2020-06-11
Argument Parser and Swift 5

Wevah:
Updates various Swift tools to use Swift 5 and the [Argument Parser](https://github.com/apple/swift-argument-parser) package.

Since the tools are currently distributed without an embedded Swift runtime, users on OSs before 10.14.4 would have already needed to install the shared Swift libs.

Care is being taken to preserve the old semantics and help text as much as possible (with a few stylistic changes for consistency with Argument Parser, etc.).

(#1583)


Wevah:
This unfortunately probably won't compile under Xcode 10, but the work will be done if/when the tools no longer need to build under Xcode 10.

kornelski:
Thank you

Wevah:
Re: Travis: It's complaining about SSH fingerprints, which is weird, since `swift-argument-parser` is on GitHub. 🤔

Wevah:
Also, SPM-in-Xcode needs Xcode 11 (whoops).

A workaround for all of this could be to check out `swift-argument-parser` as a regular git submodule.

Wevah:
If building on Xcode 10.3 should still be supported, I may need to revert to Swift 4.2 (though it looks like SAP should still build), but I won't know until I can fix these CI issues. 😖

Wevah:
Switching to HTTPS for checkout of the package fixed the error there, but now some tests are failing that are completely unrelated to anything I've touched. 🤔

kornelski:
I've restarted CI, but it consistently fails on:

> Missing package product 'ArgumentParser', please fix package resolution errors before building (in target 'generate_keys')

Wevah:
> I've restarted CI, but it consistently fails on:
> 
> > Missing package product 'ArgumentParser', please fix package resolution errors before building (in target 'generate_keys')

Is this in the 10.3 build or 11.5?

Wevah:
The issue might be that Xcode 10 doesn't have built-in support for SPM. I'll look at other options (old-fashioned git submodule might work).

zorgiepoo:
We don't use travis now and this PR should be rebased before being further reviewed. We also bumped the Xcode version to at least 11 in CI (the repo ReadMe was updated regarding the policy change for what versions should be supported further onwards). Another recent policy change we made is that changes such as this one will also need to land in 2.x branch first.

zorgiepoo:
I used the generate_keys changes here as a reference when implementing #1730 so this was quite useful to look at. I didn't write a way, like here, to use another/newly created keychain however because I wanted to still use the login keychain and didn't want to have to unlock other keychains. My rationale is if we use the login keychain to store the data, we should also be able to import to there. Maybe having another option to use a custom keychain still has its own use case too. (Also generate_keys is unfortunately pretty out of date in the master branch which important enough for me to address).

Wevah:
I haven't touched this in a while, but IIRC the export option was just to move it from the login keychain to a separate keychain (maybe that's what you said; my brain is a bit fried).

Agreed about being able to import back to the login keychain.

zorgiepoo:
I just noticed 2.x has been using Swift 5 for some while now (not sure when). Adopting Argument Parser (this is welcome) and using a separate keychain (if people still want that since you can import to login keychain now) need to be in separate PR's. Closing due to staleness.

PR #1612, opened 2020-06-16
Replaced deprecated method

philippec:
-initWithBase64Encoding has been deprecated since OS X 10.9, and that
is our deployment target.


kornelski:
Thanks

PR #1621, opened 2020-06-28
Fix casting Boolean to BOOL on Apple Silicon

kainjow:
This fixes the following error when compiling for Apple Silicon on Xcode 12:

> SUHost.m:231:23: Cast from function call of type 'Boolean' (aka 'unsigned char') to non-matching type 'BOOL' (aka 'bool')


Wevah:
Does `CFPreferencesCopyAppValue()` access a different database than going through `-[NSUserDefaults initWithSuiteName:]`?

PR #1623, opened 2020-07-07
feat: read SUFeedURL from env if not found in the plist

dannyow:
Small extension to generate_appcast command.

Users can set feed url as env variable SUFeedURL, it will be used when the default SUFeedURL key cannot be found in the app Info.plist.

Motivation - more apps configure feed straight from code without using to the plist entry. On appcast generation 
when feed url cannot be found links in appcast xml are incomplete/broken (contain only a file name).
For such cases, user can specify feed url as env variable and get correct appcast xml on output.


kornelski:
Sorry, this is something I intentionally left out. 

I don't want to support feeds that are only set programmatically. This API is prone to race conditions in Sparkle, and also breaks 3rd party applications that manage updates.

Please always put a working URL in the `Info.plist`.

Issue #1506, opened 2019-11-23
Support for `package.swift`?


CharlesJS:
Are there any plans to add support for `package.swift` to the Sparkle repo, so that it can be used via the Swift Package Manager? The CocoaPods support is nice, but the Swift PM is much more convenient to use from Xcode.

kornelski:
As far as I can tell, Swift Package Manager completely forbids havin any ObjC code, so Sparkle is not allowed to use it.

CharlesJS:
This is not true:

https://developer.apple.com/documentation/swift_packages

> Swift packages are reusable components of Swift, Objective-C, Objective-C++, C, or C++ code that developers can use in their projects.

CharlesJS:
What does appear to be true (for now) is that Swift and Objective-C can't be combined within a single Swift package, but you can work around that by breaking the Objective-C parts out and using dependencies:

https://medium.com/@joesusnick/swift-package-manager-with-a-mixed-swift-and-objective-c-project-part-1-2-19fbb43d0460

sindresorhus:
You *can* mix Swift and Objective-C in the same package, but not in the same target. So you can have a Swift target that depends on the Objective-C target. 

Example: https://github.com/sindresorhus/ExceptionCatcher/blob/b5eb74b6c9e01cf51c3619928abdb13af4a73505/Package.swift#L20-L29 (The `Internal` target there is the Objective-C part)

sindresorhus:
@kornelski Would you accept a PR that adds support for the Swift Package Manager?

kornelski:
Yes, I'd love to.

Issue #1474, opened 2019-09-22
[Feature request] Swift Package manager support.


xjbeta:
It's easier to add a library with SPM in Xcode 11.

kornelski:
That would be nice, but last time I looked, it wasn't clear to me if SPM supports ObjC frameworks too.

If you know how to make SPM packages, can you make a pull request?

xjbeta:
`.....  contains mixed language source files; feature not supported`

liamrosenfeld:
I'm looking into creating an SPM package and it appears the `contains mixed language source files; feature not supported` error is just caused by `CheckLocalizations.swift` being stored inside of `Sparkle/` instead of anywhere else

PR #1631, opened 2020-07-27
Adding support for Swift Package Manager

thebluepotato:
TL;DR: Following the steps of the closed PR #1486, I've restructured the project and added the required `Package.swift` file. All tests succeed, but it still needs extensive testing in integrating into projects. This should address #1474 and #1506.

### Restructuring the project
As initiated by #1486 which was never merged, the [Swift Package Manager](https://github.com/apple/swift-package-manager) ("**SPM**") prefers when projects are structured with two main subfolders, `Sources` and `Tests`. This is not required _per se_ but it makes the Package manifest (the `Package.swift` file) a lot less verbose. This process has been done via Xcode with the original project file open so as to preserve references. Some paths in `.xcconfig` files were updated as well.
AFAIK, the original project file builds, tests and runs as before (but this should be tested).
### Adding the Package manifest
The Package manifest declares one main product, the Sparkle framework. `bsdiff` and `ed25519` are also products, but this is not actually required. Those products were added to mimic the targets of the Sparkle xcodeproj. Please note that the ability to bundle resources and localization in SPM was added with Swift 5.3, currently in beta with Xcode 12. Using the 5.3 SPM tools requires that certain files be explicitly excluded, hence the verbosity of certain targets (e.g. excluding most .plist files).
#### Targets
The manifest then declares the required targets. In detail:
- `fileop`: `fileop.m` is excluded since it's used only for the executable `fileop`. When public headers are not in an `include` folder, their containing folder needs to be explicitly declared (here `.`). Moreover, `AppKitPrevention.h` is given to the C compiler as well.
- `bsdiff`: also declares the public headers path
- `ed25519`: the file `precomp_data.h` isn't really a header. The only way for SPM to be selective about which headers are public headers and which aren't is to move the public headers in a separate folder, preferably named `include`. Otherwise, building other targets which depend on `ed25519` failed because `ge_precomp` was not defined (I think it was when building the tests). The source files were copied from the submodule, which means that compiling via SPM does not use the same files from the submodule directly. This could be changed either via symlinks or by removing the submodule.
- `Sparkle`: the main target. The manifest is mostly self-explanatory here, but some comments are given:
  - `CheckLocalizations.swift` is excluded as it is not part of the target.
  - `DarkAqua.css` is marked as a resource to be bundled. The code calling that resource is also adapted for that situation (see below).
  - A set of `defines` is added here based on the preprocessor flags defined in `ConfigCommon.xcconfig`.
  - Finally, the `xar` and `bz2` libraries are linked.
#### Tests
For the test targets, the Sparkle Unit Tests had to be split between Swift and ObjC since SPM does not support mixed language targets. Both targets bundle their resources (see below for more info).They also define `SWIFT_PACKAGE_TESTING` which is currently a dirty hack to get them to compile despite `AppKitPrevention.h`. Indeed, for some yet unknown reason, the tests seem to import AppKit which generates a build error. The define allows the compiler to output only a warning instead of an error. This flag is used _only_ for the tests. In all other contexts, `AppKitPrevention` _still_ generates an error as expected.
The Swift tests target additionally uses an `unsafeFlag` to link the custom bridging header. It should be noted that certain tests in that target needed additional imports (such as CommonCrypto for example) which are wrapped in preprocessor directives so this is only imported when using SPM. This has also been done for Swift files in certain executables, even though the targets are not yet implemented.
Finally, the UITests target is defined, builds but cannot run because no target application is set.
### Accessing SPM Bundle resources
In Swift 5.3, Packages can include resources and have them localized. Accessing the resources however requires a slight change at the call site. In Swift, the bundle is accessed via `Bundle.module` and in Objective-C via the `SWIFTPM_MODULE_BUNDLE` macro. This was implemented throughout the project as shown by the two following examples:
#### Swift
``` swift
#if SWIFT_PACKAGE
let archiveResourceURL = Bundle.module.url(forResource: appName, withExtension: archiveExtension)!
#else
let archiveResourceURL = Bundle(for: type(of: self)).url(forResource: appName, withExtension: archiveExtension)!
#endif
```
#### Objective-C
``` objc
#if SWIFT_PACKAGE
    self.testFile = [SWIFTPM_MODULE_BUNDLE pathForResource:@"signed-test-file" ofType:@"txt"];
    self.pubKeyFile = [SWIFTPM_MODULE_BUNDLE pathForResource:@"test-pubkey" ofType:@"pem"];
#else
    self.testFile = [[NSBundle bundleForClass:[self class]] pathForResource:@"signed-test-file" ofType:@"txt"];
    self.pubKeyFile = [[NSBundle bundleForClass:[self class]] pathForResource:@"test-pubkey" ofType:@"pem"];
#endif
```
AFAIK there aren't many calls to bundled resources outside of the tests except `DarkAqua.css`. I'll have to test whether the `xib` files actually work.
Please note that there currently seems to be a bug with the Objective-C accessor with the target name. Indeed, in the process of generating the Obj-C code, the Bundle is assumed to be named with underscores instead of spaces or hyphens (e.g. `Sparkle_Sparkle_Unit_Tests_ObjC`), however, when generating the bundle, Xcode preserves the spaces/hyphens and merely generates a warning. Accessing the bundle results in a runtime error because it can't be found. Using underscores in the target name instead of spaces or hyphens works in terms of making the bundle accessible, but there still is a warning because the bundle name uses underscores. Accordingly, the solution is to not use spaces/hyphens/underscores at all in the target name.
### Testing
Testing with Cmd+U surprisingly yields a failure for `testUnarchivingDmg()` (and sometimes other tests, weird). All "failing" tests, however, succeed when done individually.

## Future steps
Steps to be done include, but are not limited to, adding SPM tests to the current CI pipeline and checking how well this integrates with future development of Sparkle. Indeed, although the changes to the sources were minimal, this does change the project structure somewhat. Integration of the various tools should also be considered.
This is still a draft because there has not been any testing with using the Sparkle with SPM in another project (yet). It should be noted that I have never worked with Sparkle and stumbled on it only because I was trying to add SPM support to another project 😅. In my experience, when testing succeeds, it bodes well for integrating into other projects via SPM. This PR is open for discussion, testing and changes. I hope this will help further advance the adoption of SPM throughout the ecosystem!


thebluepotato:
I ended up creating an `include` directory for the `Sparkle` target as well in order to expose only the public headers. This fixes the issue with `AppKitPrevention.h` and allows us not to rely on a hacky define anymore. The test targets now need to be pointed to the Sparkle Source directory to find the private headers.
I've tested integrating Sparkle into a project via SPM and currently linking fails for this project because of a duplicate `_main` symbol. One is in `AppDelegate.o` (of course) and the other is in `Sparkle.o`. Surprinsingly, after looking through the second file, it seems that SUBinaryDeltaTool defines a `main` as well and this is exposed in the compiled library. Will continue investigating.

michelf:
That's rather nice. I wish I could use it, but I assume SwiftPM won't work for the 2.x branch that has XPC targets.

thebluepotato:
> That's rather nice. I wish I could use it, but I assume SwiftPM won't work for the 2.x branch that has XPC targets.

I'm currently investigating how we could build Sparkle 2.0 from source via SPM but building the Updater.app and wrapping it as a resource might be a bit too much to ask (for now). However, with Swift 5.3, SPM is entirely capable of distributing binary targets as an `XCFramework`. Initial testing shows that it works well (and eases the way for Intel/Apple Silicon support). I'll try to figure out the best way to make it work as a Package but I'll keep investigating the "from source" path.

thebluepotato:
Little update: after lots of cheats, I've been able to bundle Autoupdate.app within the Package and it works nearly flawlessly. However, SPM does not currently support building a target and then bundling the product within another target. It also does not currently support building `.app` targets, only Swift executables. Moreover, there is a lot of customisation happening when Sparkle is built through Xcode with multiple build scripts which cannot be easily emulated with SPM. As such, before SPM gains more [extensible build tools](https://forums.swift.org/t/package-manager-extensible-build-tools/10900), this PR will remain a draft (in hopes also that it helps serve as an example in bundling resources).
**However**, distributing Sparkle as a binary Package (as it does with Cocoapods) is nearly trivial and easily extensible to 1.x and 2.x. I'll draft a PR very soon, but it would merely involve adding a Package manifest to the repository (with a binary target) and the following steps to Sparkle's release process (roughly) :
```
xcodebuild archive -project 'Sparkle.xcodeproj' -scheme 'Sparkle' -configuration Distribution -destination 'platform=macOS,arch=x86_64' -archivePath './Archives/Sparkle.framework-macos.xcarchive' BUILD_LIBRARY_FOR_DISTRIBUTION=YES SKIP_INSTALL=NO
xcodebuild -create-xcframework -framework './Archives/Sparkle.framework-macos.xcarchive/Products/Library/Frameworks/Sparkle.framework' -output 'Sparkle.xcframework'
zip -r Sparkle-$(VERSION-NUMBER).zip Sparkle.xcframework
swift package compute-checksum Sparkle-$(VERSION-NUMBER).zip
# Then the url to the zip file and checksum should be put into the Package manifest and you're done
```

Initial testing has shown that this is equivalent to the Cocoapods solution, but works natively in Xcode (version 12).

kornelski:
So I guess now it only needs documentation>?

thebluepotato:
This PR should be closed since #1634 and #1635 have been successfully merged some time ago. This PR tried to implement Spark as a source code Swift Package but this is not currently feasible whereas binary targets are very close to what was being done with Cocoapods and actually works even better.

Issue #1620, opened 2020-06-23
Support Apple Silicon (ARM64)


bdkjones:
There’s about six months until ARM Macs are in the wild. To be ready for them, apps that integrate Sparkle must be rebuilt as Universal binaries. To rebuild, they’ll first need a Universal build of Sparkle.

I’m opening this issue to track that progress and discuss any changes that must be made to support ARM.

I suggest this be made a high priority because it blocks app developers who use Sparkle from rebuilding their own apps.

kornelski:
I've verified that Sparkle works on Big Sur x86-64. I've got a tip that Sparkle is compatible with ARM already.

bdkjones:
![well-that-was-easy](https://user-images.githubusercontent.com/926566/85472800-385b7c00-b567-11ea-9b4b-a339648f38e8.jpg)

XericDesign:
In the WWDC session on migrating to Apple Silicon they showed an arm64 build of Sparkle but not how they did it. Can the next update posted here please be a Universal 2 binary?

paulz:
we were able to build successfully Distribution package on Xcode Version 12.0 beta (12A8158a) with no code changes. 
```
lipo -archs Sparkle.framework/Versions/A/Sparkle
x86_64 arm64
```
hoping to make an update when Xcode 12 release is available.

kainjow:
I got a warning when compiling though: #1621

cazoobi:
Sparkle throws an error when you build your project on a DTK in Xcode. It states something about the .framework being compiled for intel and how it will be ignored.

The app builds and runs but when you try anything to do with the update mechanism, it doesn't work.

Im guessing like @bdkjones initially said, we might need a universal variant of the framework.

kornelski:
Did you build an ARM version of Sparkle yourself? Downloads from GitHub releases are compiled for Intel only for now.

cazoobi:
> Did you build an ARM version of Sparkle yourself? Downloads from GitHub releases are compiled for Intel only for now.

I did not, I actually didn't realise I could compile my own version of the framework for security reasons.

rknall:
Will this be a hard requirement for only support universal binaries? We are not planning to release one, but rather individual ones, and therefore it would require us some major reengineering effort.

[...]

PR #1634, opened 2020-07-29
Enable SPM support via binary target

thebluepotato:
## Summary
Following off off #1631, I've added support for the Swift Package Manager ("**SPM**") via a binary target. This should address #1474 and #1506 and can trivially be implemented for the 2.x branch.

### Why a normal `target` doesn't work
As detailed in #1631, building Sparkle from Source via SPM is feasible but is limited by certain limitations SPM currently has. For example:
- It was not possible to build the Autoupdate.app with SPM because SPM does not yet support building an app (only libraries and executables but [this might change in the future](https://forums.swift.org/t/my-swiftpm-wishlist-aka-proposal-proposals/35292/4)).
- Moreover, it is not possible to use the build product of one target as a resource to be bundled within another target. This means that the Autoupdater.app would have to be built and added to the repository itself.
- Finally, the extensive use of build scripts within the various Xcode targets cannot be emulated with SPM. Those scripts are used, amongst others, to complete certain localizations.

These limitations led me to drop the idea of building Sparkle from Source via SPM. Maybe future [extensible build tools](https://forums.swift.org/t/package-manager-extensible-build-tools/10900) will address all this.

### Using a `binaryTarget`
As of Swift 5.3, Packages can depend on binary dependencies and such dependencies can be distributed via Swift Packages. I strongly recommend this [document](https://developer.apple.com/documentation/swift_packages/distributing_binary_frameworks_as_swift_packages) and the corresponding [WWDC20 video](https://developer.apple.com/wwdc20/10147) for more info.
As can be seen in `Package.swift`, distributing a binary target is straighforward. The Package manifest defines a library with one target which is a `binaryTarget`. The latter requires a HTTPS URL pointing to a ZIP-compressed XCFramework and a checksum (generated with `swift package compute-checksum Framework.zip`.

### Implementation
This PR adds `release-spm` to the project's `Makefile`, which is mostly based on `release`. It first archives the `Sparkle` target, wraps it into `Sparkle.xcframework` and finally builds the `Distribution` scheme. The `make-release-package.sh` post-build script is updated to compress the XCFramework, compute its checksum and update the Package manifest in the repo. Once the build has completed, the updated Package manifest should be committed before tagging/releasing. The `Sparkle-SPM-x.x.x.zip` file should be uploaded as a release asset along with the usual compressed tarballs. All in all, Sparkle distributed via SPM as a binary target is very similar to what's currently done via Cocoapods.

### Integration
Projects depending on Sparkle can now drop Cocoapods and use the Swift Package Manager included in Xcode. Adding a dependency is trivial. I have tested this in one project depending on Sparkle and it was able to build, run and download the latest update without any hitches.
**Note**: binary targets are a feature of Swift 5.3 and are only available starting with Xcode 12 (currently in beta).
**Note**: compiling with Xcode 12 (beta 3) means we get both Intel and Apple Silicon support in the XCFramework for free :partying_face:.
**Note**: importing Sparkle as Swift Package needs the zip archive to be released first. A convoluted temporary workaround to test the integration:

1. clone this PR
2. run `make release`
3. Create a local `SparkleTest` repo
4. From the result of `make release`, copy `Package.swift` and `Sparkle.xcframework` to the `SparkleTest` repo
5. In `Package.swift`, change the `.binaryTarget` to the following: `.binaryTarget(name: "Sparkle", path: "./Sparkle.xcframework")`
6. Commit
7. In your project, after deintegrating Cocoapods and adding all other dependencies, add a new Swift Package dependency. As URL, indicate `file:///path/to/repo/SparkleTest` and click Next.
8. Select the `master`/`main` branch. You should be able to add Sparkle now.

### Further improvements/bugs
One should note the following improvements to be made or bugs to be addressed:

- [x] ~There should be only one `release` and no separate `release-spm`. This has been done for demonstration purposes but the two should be "merged" (i.e. release-spm should also `cat` the Podspec, etc.). Currently, `release` is broken because it uses the same script after building and it cannot find the XCFramework.~ There is now again only one `release` in the Makefile.
- [x] ~The release script uses a template Package manifest to generate the "real" manifest by replacing placeholders with the correct version and checksum. This was hacked together and can certainly be improved. Some regex magic with `sed` should allow for updating the "real" manifest directly.~ Implemented regex magic, Package manifest is automatically updated after each build of the `Distribution` scheme.
- [x] ~The checksum of the ZIP file changes after each `make release-spm`. The XCFramework is of course identical if there have been no code changes but `zip` includes metadata which changes the archive ever so slightly.~ See [here](https://stackoverflow.com/a/9714323/4552438). This is unavoidable and affects also other archives. The script was updated to remove some extra attributes.
- [x] The checksum is calculated with `swift package compute-checksum` but apart from a few sanity checks, it merely outputs the SHA-256 checksum. ~`shasum -a 256` could be used instead.~ `shasum`'s output is not ideal and the only "drawback" of the `swift` command is that it generates a `.build` directory but this one is cleaned up by the script. The script now chekcs the Xcode version first.
- [x] ~The binary tools that come with Cocoapods via the compressed tarball are not available via SPM. Indeed, the Package may only declare a binary target that is a ZIP-compressed XCFramework. However, the SPM checks out the Git repository as well so we could included some separate downloader (or just tell the user in README that they should download the tarball to use the tools)~ It seems after all that you can put pretty much anything inside the XCFramework before zipping it. You can actually add othe files alongside the XCFramework inside the archive. Accordingly, once Xcode has downloaded the Package, you should be able to find all the binaries (and debug symbols) with `Show in Finder`. I've asked [here](https://forums.swift.org/t/what-are-the-limitations-to-what-can-be-bundled-in-a-binary-target/38877) for more info and hopefully Xcode will soon display all the archive's file in the Project navigator (and not just the XCFramework).
- [ ] The `Distribution` scheme had to have `Parallelize Build` deactivated. This is not limited to this branch, it seems that when builds are parallelized, `BinaryDelta` is built _after_ `make-release-package.sh` is run. See [here](https://github.com/sparkle-project/Sparkle/issues/1620#issuecomment-664505404). When the option is off, builds can be done via the command line but it seems to break building via Xcode.
- [x] Update README and Travis-CI

I'm open to suggestions, feedback, etc.


thebluepotato:
There are still some improvements that I'd like to address:

- [ ] Beyond the parallelized build issue mentioned above, the `Distribution` scheme does not currently build the archive and XCFramework. I do not know if those steps can be integrated in the scheme. Currently, the archive and XCFramework are built _before_ the `Distribution` scheme in `make release` or in Travis CI. I think those steps should be built-in
- [ ] The documentation for the 1.x releases is mostly on the website so I guess installing via SPM should be addressed there

PR #1635, opened 2020-07-29
Enable SPM support via binary target [2.x]

thebluepotato:
This PR is the 2.x extension of #1634. Commits done to the 1.x version will be mirrored here. Please read #1634 for details.


thebluepotato:
This PR is now ready for review. The only difference from the 1.x branch (other than the additional SparkleCore framework) is that the `XCFrameworks` scheme isn't shared. This was more or less random and I don't know what advantages or disadvantages there are between shared or non-shared schemes. I'd like your input on whether it should be shared or not.

michelf:
Am I correct saying the 2.x SPM package does not make available the XPCs but only the framework? I was looking forward to trying it, but without the XPCs it won't work from my sandboxed app. I assume this is a SPM limitation.

thebluepotato:
@michelf As far as I understand how the XPCs work, they are all bundled within the package binary (just like if you were to download the compressed tar archive form the releases page). When added via Xcode however, which is porbably the most common way now to use the SPM, it will only show the XCFrameworks. The other binaries and XPCs are avalaible though, by clicking "Show in Finder" on said XCFramework. The Xcode team is currently examining the possibility to show all files bundled in the SPM zip file within the editor, but I don't think there's an ETA yet. See here: https://forums.swift.org/t/what-are-the-limitations-to-what-can-be-bundled-in-a-binary-target/38877/

PR #1636, opened 2020-07-30
Many cleanups and forward-ported fixes and features

gwynne:
The following changes, forward-ports from the main branch, and cleanups have been made:

- The release package generated by the Distribution target includes the `generate_keys` and `sign_update` tools, and generates an `.xz` archive instead of a `.tar.bz2` (forward-ported).
- `build-docs.sh` no longer incorrectly executes commands it echoes in its output (forward-ported).
- The Distribution scheme defaults to building Release rather than Debug.
- `generate_keys` has commented-out support for marking the generated key syncable over iCloud (currently disabled since it requires adding entitlements to the tool).
- `generate_keys` now print considerably more useful explanatory text and provides more correct output.
- Forward-port of #1499 (release notes localization support).
- `generate_appcast` now allows overriding the downloadable URL prefix, uses hashes for the names of output directories when extracting archives, supports a `-h` flag for printing usage information, supports specifying an Ed25519 public key inline on the command line, and overriding the keychain used for looking up stored key information (all forward-ported).
- Corrections to Swift build configuration, in particular the specification of Whole Module Optimization and the requirement for Swift 5.
- As the 2.x beta requires 10.9, update a number of old 10.7 references to 10.9 instead.
- Fix or silence dozens of warnings, mostly in old Objective-C code.
- Disable `COPY_PHASE_STRIP` for all Release configurations as it generates a large number of warnings about copying codesigned binaries and offers little or no benefit.


kornelski:
Thank you

PR #1640, opened 2020-08-29
fix: read SUAutomaticallyUpdate from Info.plist

trevordevore:
`SUAutomaticallyUpdate` will now be read from `Info.plist` if it has not been set in `NSUserDefaults`. This brings the behavior into line with the documentation.


kornelski:
Thank you

Issue #1642, opened 2020-09-01
Provide option to specify release notes url for generate_appcast


andalman:
We have an app where the `SUFeedURL` points to a location on our own server. However, we prefer to host our updates and release notes on S3.

The `--download-prefix-url` argument ( #1569 ) to `generate_appcast` solves the first problem – where to find the updates – nicely. However, there currently isn't a way to specify a URL prefix for release notes.

It would be great if there were an argument, say `--release-notes-prefix-url`, that behaves like the "download prefix" argument, but works for release notes.

If you agree, I can submit the pull request.

kornelski:
Yes, please make a PR

yeahdongcn:
When can we release this? It looks very useful for us.
Thanks @andalman for providing the new feature.

PR #1644, opened 2020-09-01
Add —release-notes-url-prefix to generate_appcast

andalman:
This is a PR for #1642 

Added a `—release-notes-url-prefix` argument to `generate_appcast` so that users can provide a prefix that will be used to construct the URL for the location of release notes.

(This commit is modeled after the fix for #1571, which introduced `--download-url-prefix`.)


kornelski:
Thank you

PR #1645, opened 2020-09-01
Use angle-bracketed includes for all public framework headers

milend:
Xcode 12  now suggests turning on [CLANG_WARN_QUOTED_INCLUDE_IN_FRAMEWORK_HEADER](https://xcodebuildsettings.com/#clang_warn_quoted_include_in_framework_header) which ends up producing warnings in projects which import the Sparkle headers. The Xcode setting corresponds to `-Wquoted-include-in-framework-header`.

This PR fixes all the includes in the public framework headers, so that no such warnings are emitted anymore. Sparkle continues to build without any issues after the change.

### DIFF
--- a/Sparkle/SPUDownloadData.h
+++ b/Sparkle/SPUDownloadData.h
@@ -12,7 +12,7 @@
 #import <Foundation/Foundation.h>
 #endif
 
-#import "SUExport.h"
+#import <Sparkle/SUExport.h>
 
 NS_ASSUME_NONNULL_BEGIN
 
--- a/Sparkle/SPUDownloader.h
+++ b/Sparkle/SPUDownloader.h
@@ -11,7 +11,7 @@
 #else
 #import <Foundation/Foundation.h>
 #endif
-#import "SPUDownloaderProtocol.h"
+#import <Sparkle/SPUDownloaderProtocol.h>
 
 @protocol SPUDownloaderDelegate;
 
--- a/Sparkle/SPUDownloaderSession.h
+++ b/Sparkle/SPUDownloaderSession.h
@@ -11,8 +11,8 @@
 #else
 #import <Foundation/Foundation.h>
 #endif
-#import "SPUDownloader.h"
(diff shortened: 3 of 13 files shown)


kornelski:
The CI doesn't seem to like it.

milend:
@kornelski: Looks like it was caused by exposing definitions twice. Once through the Sparkle framework (Clang module) and then again via a quoted include. The [CI log](https://travis-ci.org/github/sparkle-project/Sparkle/jobs/723230782) did not actually print the full diagnostic which I'm including here for future reference:

```text
/repos/Sparkle/Sparkle/SUFileManager.m:99:73: error: reference to 'SUAuthenticationFailure' is ambiguous
            *error = [NSError errorWithDomain:SUSparkleErrorDomain code:SUAuthenticationFailure userInfo:@{ NSLocalizedDescriptionKey: @"Unable to grant authorization to perform action because it is explicitly turned off" }];
                                                                        ^
In file included from /repos/Sparkle/Sparkle/SUFileManager.m:10:
/repos/Sparkle/Sparkle/SUErrors.h:44:5: note: candidate found by name lookup is 'SUAuthenticationFailure'
    SUAuthenticationFailure = 4001,
    ^
In module 'Sparkle' imported from /repos/Sparkle/Sparkle/SUErrors.h:17:
/repos/Sparkle/Sparkle/SUErrors.h:44:5: note: candidate found by name lookup is 'SUAuthenticationFailure'
    SUAuthenticationFailure = 4001,
    ^
```

`SUFileManager.m` first imports `SUErrors.h` via a quoted include and then `SUErrors.h` in turn imports `<Sparkle/SUExport.h>` which ends including the `Sparkle` Clang module and thus redefines the `SUError` type.

The issue should be addressed by 321bab1.

kornelski:
Thank you

zorgiepoo:
This breaks SparkleCore in the 2.x branch. I can't think of a better solution than using "" includes and inserting a pragma ignore warning.

Issue #1646, opened 2020-09-01
Provide output filename for generate_appcast


andalman:
Our application has an `SUFeedURL` that points to the following location: `https://get.180g.co/updates/vellum/`.

`generate_appcast`, however, makes the assumption that this URL ends in `<filename>.xml`. And when it creates an appcast, it determines the output filename by using the `lastPathComponent`. In our case, that is just `vellum`, which we are forced to rename later.

I thought a useful feature might be to add a `-o filename` option to `generate_appcast`. However, in looking at the code, I realized that `generate_appcast` can create multiple appcasts, depending on the `SUFeedURL` specified in each archive in the folder. So, clearly, a single output option will not work.

I still think that providing a way to specify the filename(s) for generate_appcast's output would be useful. In the hopes of starting a discussion – and coming up with a good solution – here are some possible options:

1. Allow `-o filename`, but it only works when `generate_appcast` would create a single appcast file. Otherwise, it throws an error and doesn't write anything. Or, it throws an error and writes out as before, ignoring the `-o` option.
1. Add an option that provides a mapping from appcast filename to output filename, e.g. `-o vellum:update-vellum.xml`.
1. Add a application plist key that specifies the filename. For instance, our application could specify a `SUGenerateAppcastFilename` of `update-vellum.xml`. `generate_appcast` would look for that when it creates the appcast; if it doesn't find it, then it falls back to its current behavior. (This doesn't seem great since it requires changing an application to affect the output of a `generate_appcast`.)

I did wonder whether this need was too specific, but I looked at all of the Sparkle-based applications I have on my computer, and it doesn't look like we are the only ones who would benefit from this feature. For instance, Coda2, Fantastical, Kaleidoscope, Things, Transmit, and duet all have an `SUFeedURL` that doesn't end with `<filename.xml>`.

Thoughts?

kornelski:
Thanks for the suggestion. I think option 1 is sensible.

andalman:
I also think that's the simplest and clearest.

Do you think `generate_appcast` should not write anything if `-o` is specified and it would write multiple appcast files? Or just ignore the `-o` option in that case?

kornelski:
I think returning an error is OK. If someone needs to write multiple appcasts, and use custom names, they can split apps into multiple directories and run generate_appcast on each individually.

andalman:
Great. I have a fix. Shall I submit a PR?

kornelski:
Yes please

PR #1648, opened 2020-09-03
Feature 1646 generate appcast output filename

andalman:
This is a PR for #1646 

Added a `-o /path/to/file.xml` command line option that directs `generate_appcast` to save a single appcast to the given location.

If this option isn't specified, then `generate_appcast` falls back to its current behavior: the appcast(s) are written to the archive folder. Furthermore, as discussed in the issue, `generate_appcast` will error out if `-o` is specified and multiple appcasts would have been generated.

In addition, this PR contains a commit that adds a CommandLineArgs struct. This should make it slightly cleaner to add new args in the future – at least until `generate_appcast` gets a new arg parser.


kornelski:
Looks great, thanks!

Issue #1647, opened 2020-09-03
generate_appcast doesn't always update release notes when --release-notes-url-prefix is used


andalman:
With the addition of `--release-notes-url-prefix` (#1642), it is now possible for multiple runs of `generate_appcast` to create differing release notes URLs for the same archive update. (Before this fix, the release notes URL was always determined by the archive's `SUFeedURL` and, therefore, was unlikely to change across `generate_appcast` runs.)

However, if the appcast file isn't deleted between runs, `generate_appcast` will *not* update the release notes URLs in  the appcast. Here's a (contrived) example:
- Run `generate_appcast --release-notes-url-prefix http://oops.com/ folder-of-archives/`
- Open the `appcast.xml` file and confirm that it contains items with release notes URLs that begin with `http://oops.com`
- Notice that you didn't mean to type `http://oops.com`
- Immediately re-run `generate_appcast` using the correct prefix
- Open the `appcast.xml` file and notice that the release notes URL prefixes still begin with `http://oops.com`

I have a fix, and can submit a PR if desired.

kornelski:
Yes, please submit a fix

PR #1649, opened 2020-09-03
generate_appcast updates the release notes element’s URL when required

andalman:
This is a PR for #1647 

Previously, generate_appcast could only add or remove the release notes element for an update. This was probably fine, because it was unlikely to change: the release notes URL was determined by the SUFeedURL.

However, now that —release-notes-url-prefix was added, the release notes element CAN change. So, with this commit, if the release notes element is present for an update, the URL will be set correctly.


andalman:
Not sure why the Travis CI build failed - I built with Xcode12 and ran the "failing" test (according to Travis), but it passed for me.

kornelski:
It failed due to a timeout. Could be a fluke.

PR #1651, opened 2020-09-10
generate_appcast follow symbolic links

shadone:
When archives are stored as symlinks e.g. due to storing binary files in a
git-annex repository, everything works file except file size in the generate
feed xml file. This patch makes generate_appcast correctly determine file size
by following the symlink


kornelski:
Thanks! What an unexpected gotcha.

PR #1652, opened 2020-09-10
Fix issue #1568 by creating a lockfile

core-code:
Fixes issue https://github.com/sparkle-project/Sparkle/issues/1568

Issue #1568, opened 2020-04-01
improved interoperability with app-updaters


core-code:
this is not a bug report per-se but rather a request for an enhancement that would be in the best interest of users of apps using Sparkle.

Sparkle allows apps to implement auto-update functionality, which is great. however, some users prefer to have software that can keep all their apps up-to-date at once as this has various benefits. we are offering one of those products [1], but there are other solutions on the market, both open-source as well as commercial.

all in all, our product works great [2] and can update more than 5000 apps with the click of a button. however, there is a rare problem when our product tries to update an app, while the app also tries to update itself at the very same point-in-time using Sparkle. this can lead to failures on either end, and i guess in the absolute worst case it could even leave the user with a destroyed app. this is uncool.

this report is about preventing this problem in the interest of all users of apps with Sparkle that also use an app-updater (ours or any other).

note that we try to be /very/ careful when updating an app and we have a long array of protection-measures [e.g. we have a good look if any org.sparkle-project.* process is running inside the app-bundle] and sanity-checking but our logfiles still indicate that some updates fail because a concurrent update has been performed.

this *mainly* affects those apps that have configured Sparkle to perform updates in a transparent fashion. in this case our app will ask the app-to-be-updated to quit, in order  update it - but this is exactly the time when your transparent update-at-quit code kicks in: perfect collision course. but it *also* affects those apps using Sparkle in their stock settings if the user has asked the app to update itself while asking our app to update it too, at the very same time. yes, users will do nonsense like that and there is no way to prevent it, so we should deal with it gracefully.

long story short we'd like to prose a solution to avoid *any* app-updater trying to update an app at the very same time as it also tries to update itself. i imagine a solution has two parts:
1.) a way for Sparkle to tell the outside world that it is going to update the host app. if this 'flag' is set then no outside app-updater should try to update this app
2.) a way for app-updaters to tell Sparkle not to update its host app because an update is already being performed by the app-updater
so, whoever comes first "wins" and should continue doing the update without having to fear that the other party will do it at the same time. basically we need a mutex/lock for app-updates.

implementing a completely safe/atomic IPC lock is probably hard and/or overkill but even a simple implementation could greatly help to prevent the discussed problem.

i haven't put too much thought into it since i want to hear your preferences here first but i suppose we could use either a lockfile or NSDistributedNotification's for this. a lockfile is probably better, unless the 'sandbox' comes into play but i am not sure how many of apps using Sparkle are sandboxed.

[...]

PR #1655, opened 2020-09-16
Create lockfile for improved compatibility with 3rd party updaters

vitorgalvao:
Rebase of https://github.com/sparkle-project/Sparkle/pull/1652, done and submitted with permission from the author (@core-code).

Closes https://github.com/sparkle-project/Sparkle/pull/1652.
Closes https://github.com/sparkle-project/Sparkle/issues/1568.

### DIFF
--- a/Sparkle/SUGlobalUpdateLock.m
+++ b/Sparkle/SUGlobalUpdateLock.m
@@ -59,7 +59,7 @@
 
 - (NSString *)fileLockPath
 {
-    return [NSString stringWithFormat:@"/private/var/tmp/%@.Sparkle.pid",  [self identifier]];
+    return [NSString stringWithFormat:@"/private/tmp/%@.Sparkle.pid",  [self identifier]];
 }
 
 @end
--- a/Sparkle/SUGlobalUpdateLock.m
+++ b/Sparkle/SUGlobalUpdateLock.m
@@ -1,0 +1,65 @@
+//
+//  SUGlobalUpdateFileLock.m
+//  Sparkle
+//
+//  Created by Bibhas Acharya on 7/12/20.
+//  Copyright © 2020 Sparkle Project. All rights reserved.
+//
+
+#import "SUGlobalUpdateLock.h"
+#import "SULog.h"
+
+
+@implementation SUGlobalUpdateLock
+
++ (SUGlobalUpdateLock *)sharedLock
+{
+    static dispatch_once_t once;
+    static SUGlobalUpdateLock *sharedInstance = nil;
+    dispatch_once(&once, ^{
+        sharedInstance = [self new];
+    });
+    return sharedInstance;
+}
+
+- (void)lock
+{
+    NSFileManager *fileManager = [NSFileManager defaultManager];
+    NSString *fileLockPath = [self fileLockPath];
+    
+    BOOL success = [fileManager createFileAtPath:fileLockPath contents:nil attributes:nil];
+    if (!success) {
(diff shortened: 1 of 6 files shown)


kornelski:
What about recovery? What if Sparkle makes the lock file, and the host app crashes before unlocking?

Sparkle/SUGlobalUpdateLock.m
```diff
@@ -0,0 +1,65 @@
...
+    if (resp == nil) {
+        // If there's no bundle identifier, use the executable path
+        resp = [[[NSBundle mainBundle] executablePath] stringByReplacingOccurrencesOfString:@"/" withString:@"."];
+    }
+    return resp;
+}
+
+- (NSString *)fileLockPath
+{
+    return [NSString stringWithFormat:@"/private/var/tmp/%@.Sparkle.pid",  [self identifier]];
```
kornelski, line 62: I think it should be `/tmp` (or `/private/tmp` if you want to make it Apple-y), because `var/tmp` is not automatically cleared on reboots. It'd be counter-productive to persist the lock file across reboots.

vitorgalvao:
@core-code I’ve given you direct access to my fork. Feel free to clone that one and push changes at will, which should auto-update this PR.

core-code:
first of all, we've worked on multiple solutions here. this is the simplest one, because i we kind of agreed doing it that way in the bug-report and i thought this is easiest to get merged as a starting point. as mentioned here (https://github.com/sparkle-project/Sparkle/issues/1568#issuecomment-690587519) we also have a more complex solution available too.


> What about recovery? What if Sparkle makes the lock file, and the host app crashes before unlocking?

well, we thought about this scenario. there is no perfect solution, because we obviously can not execute code during or after crashing (signal handlers notwithstanding).

so, this is addressed in 2 ways:
• this is a simple solution that only created the lockfile as information to outside third party updaters. so, if there is a stale lockfile because of a crash, 3rd party updaters (that have support for this lockfile) will refuse to update the app. but the lockfile will not prevent Sparkle itself from updating the app. just as is currently the case, if the app crashes during update, the user can just go ahead and try the update again. no changes so behaviour here except that after trying an update in the app itself, and if it crashes, 3rd party updaters 'can not' update the app until a reboot. this is pretty hypothetical since crashing during update are rare to begin with and a user trying an update in-app first and then switching to a 3rd party updater is even more unlikely

• also, to address this exact point, the lockfile is always deleted when the Sparkle-enabled app launches. this is this line:

`    [[SUGlobalUpdateLock sharedLock] unlock];  // For safety, remove active lock if present in any case and not just when relaunching from an app update
`

Sparkle/SUGlobalUpdateLock.m
core-code: good catch - changed as per your recommendation.

kornelski:
Thanks for the update!

core-code:
awesome thanks!

core-code:
@zorgiepoo it seems this functionality is not present anymore in recent Sparkle 2.x releases. i guess it was never in Sparkle 2.x to begin with? will a PR to re-add this again be accepted?

core-code:
thanks for the explanation! the idea about the lock-files was also that those could also be written by other updating frameworks, so that there was a single "API" for external stand-alone updaters to query before proceeding with updating an app. 

however in practice the only other common updater framework is Squirrel and they are not willing to accept patches to improve anything regarding compatibility and safety, so that idea never played out.

so, we'll just query the mach-lookup status service as you've suggested.

PR #1657, opened 2020-09-24
Use spaces

schriftgestalt:


kornelski:
Thanks for the cleamup. This also accidentally adds `self.window.movableByWindowBackground = YES;`. Can you remove it?

schriftgestalt:
done.

PR #1662, opened 2020-10-03
Fix warnings when using modules and importing headers (2.x)

zorgiepoo:
This brings over this PR to silence warnings about #import being used:
https://github.com/sparkle-project/Sparkle/pull/823

And also silences `-Watimport-in-framework-header` when using @import in framework headers as noted in:
https://developer.apple.com/documentation/xcode-release-notes/xcode-11_4-release-notes

This is less priority than the double quote include PR.


kornelski:
Thanks!

PR #1664, opened 2020-10-12
Cleanup generate_keys and fix URLs for localized release notes

gwynne:
This PR includes the following changes:

- `generate_keys/main.swift` reorganized somewhat and cleaned up significantly. The code has been heavily commented, failure messages are displayed consistently (and where the context allows, using ANSI color), error messages and informational text has been rewritten for grammar and clarity, and top-level code has been encapsulated in function context for consistency. The remaining vestiges of the abandoned "syncable keychain item" implementation have been removed.

- `generate_keys` now has a "lookup mode", enabled by passing the `-p` flag on the command line. In this mode, `generate_keys` looks up whether a key already exists - if so, it outputs the public key as a single line of Base64 and immediately exits with success; if not, an error is printed and it exits with failure. This mode is intended to assist other tools and scripts in retrieving the public signing key automatically, without having to hardcode Sparkle's keychain matching parameters or any assumptions about how the data in the keychain item is laid out. An obvious use of this would be in a shell script Xcode build phase which automatically updates the value of `SUPublicEDKey` in an app's `Info.plist`.

- `generate_appcast` now generates full URLs for links to localized release notes in an appcast, utilizing the same logic used to generate the link to "base" release notes (use a release notes URL prefix if specified, otherwise use the feed URL if available). Previously, localized release notes links would be specified only as filenames. Before:

```xml
<sparkle:releaseNotesLink>https://bucket.s3.amazonaws.com/release%20notes-123.html</sparkle:releaseNotesLink>
<sparkle:releaseNotesLink xml:lang="de">release notes-123.de.html</sparkle:releaseNotesLink>
<sparkle:releaseNotesLink xml:lang="zh">release notes-123.zh.html</sparkle:releaseNotesLink>
```

After:

```xml
<sparkle:releaseNotesLink>https://bucket.s3.amazonaws.com/release%20notes-123.html</sparkle:releaseNotesLink>
<sparkle:releaseNotesLink xml:lang="de">https://bucket.s3.amazonaws.com/release%20notes-123.de.html</sparkle:releaseNotesLink>
<sparkle:releaseNotesLink xml:lang="zh">https://bucket.s3.amazonaws.com/release%20notes-123.zh.html</sparkle:releaseNotesLink>
```

This change guarantees that localized release notes always load correctly when applicable even when their filenames contain otherwise problematic characters such as spaces - previously, in many cases they would simply fail to load.

### DIFF
--- a/generate_appcast/ArchiveItem.swift
+++ b/generate_appcast/ArchiveItem.swift
@@ -187,21 +187,24 @@ class ArchiveItem: CustomStringConvertible {
         if self.getReleaseNotesAsHTMLFragment(path) != nil {
             return nil
         }
-        guard let escapedFilename = path.lastPathComponent.addingPercentEncoding(withAllowedCharacters: .urlPathAllowed) else {
+        return self.releaseNoteURL(for: path.lastPathComponent)
+    }
+    
+    func releaseNoteURL(for unescapedFilename: String) -> URL? {
+        guard let escapedFilename = unescapedFilename.addingPercentEncoding(withAllowedCharacters: .urlPathAllowed) else {
             return nil
         }
-        
         if let releaseNotesURLPrefix = self.releaseNotesURLPrefix {
-            // If a URL prefix for release notes was passed on the command-line, use it
+            // If a URL prefix for release notes was passed on the commandline, use it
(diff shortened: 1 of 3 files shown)


kornelski:
That's great. Thank you

PR #1676, opened 2020-11-16
Remove max size from update alert.

gnachman:
Some users would prefer to resize the update alert window to make release notes easier to read.

It's not clear to me what the benefit of limiting the max window size is, especially in height. Forcing the width to 700 pts is useful because it makes the presentation more predictable.

This PR relaxes only the max-height restriction.

### Risks

I have tested this change on macOS 10.15. Unfortunately I don't have installs of older OS versions to test with. The only risk I see is that interface builder won't let you add a width constraint to the root view. Considering that is commonly done programmatically, I believe this is a bug/limitation of IB and not a real restriction imposed by AppKit.

### DIFF
--- a/Sparkle/ar.lproj/SUUpdateAlert.xib
+++ b/Sparkle/ar.lproj/SUUpdateAlert.xib
@@ -206,7 +206,6 @@ DQ
                     </button>
                 </subviews>
                 <constraints>
-                    <constraint firstAttribute="width" constant="700" id="VNp-am-qKn"/>
                     <constraint firstItem="10" firstAttribute="leading" secondItem="7" secondAttribute="trailing" constant="22" id="2ef-4I-Il3"/>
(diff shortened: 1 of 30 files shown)


kornelski:
I have mixed feelings about fixed width. Windows that don't change the width at all feel unnaturally rigid to me.

OTOH I have no problem with relaxing max height. Maybe it should just be a large number instead of 700?

gnachman:
I'm fine with removing the width restriction. That also removes the only part of this PR that makes me nervous.

Windows that are resizable but have some large arbitrary bound are rare. I can't think of an example in macOS. What purpose would such a limit serve?

kornelski:
It'd effectively be unrestricted, but easier to implement.

gnachman:
I've removed the width limitation. I don't see quite how adding a large limit would make this easier.

kornelski:
Thanks

PR #1683, opened 2020-11-27
Update SUSystemProfiler.m

pierswalter:
Masked sysctl hw.cputype value when evaluating the CPU type.
Used macros from mach/machine.h instead of magic numbers.
Added support for ARM in evaluation of hw.cpusubtype.
Used "Other" instead of "Unknown" as visibleCPUType for values not covered by the explicit cases.


kornelski:
Thanks!

PR #1684, opened 2020-11-27
Made buttons wide enough to display the full text.

pierswalter:
Made buttons wide enough to display the full text.
Also made window wide enough to display the widened buttons (el, fi, ru, uk).


kornelski:
Thanks. I suppose we should switch to Base locale now to avoid having so many nearly-duplicate files with small differences.

pierswalter:
Makes sense.
Eventually, I might look at that if nobody else will have done this by then. But this will not happen this year…

Instead, I’ll create another xib file related pull request soon that will once again touch all languages.
All of this is stuff that I’ve done on my private Sparkle fork over the years…

Piers

Issue #1685, opened 2020-11-27
Switch to Base localization


kornelski:
Because of compatibility with 10.7, Sparkle had separate UI xib files for each language. Now that Apple doesn't support anything below 10.9, we can drop 10.7 compat too and use Base localization, i.e. a single xib file for layout and .strings files for text.

Command-line interface builder has an option to extract all text from a xib file and write it to a .strings file.

PR #1686, opened 2020-11-27
Selectable titles

pierswalter:
Made titleText and descriptionText selectable. This allows users to copy the text of all GUI elements when reporting a bug.
(only two flags have been set in each file - the high number of changes in some of the xib file is a result from opening the files in a newer IB version)


kornelski:
Some elements have ` misplaced="YES" `. It's difficult to review these changes. Maybe it'd be better to do #1685 first?

Also maybe we shouldn't have bugs that need reporting ;)

pierswalter:
Yes, taking care of #1685 first makes sense.

I originally made this change many years ago when our software was still using Sparkle 1.5b6 because we needed to support older macOS versions than the then current Sparkle did.
Back then, our users did run into the occasional bug. But this has not happened any more since we updated to a more current Sparkle version.

I will wait until #1685 has been resolved (or until I will have done this myself) and will reapply this change and provide a new pull request at that time.

Issue #61, opened 2011-10-23
Add an "Older Release Notes" popup menu


andymatuschak:
***Andy Matuschak (andymatuschak)** [reported (on Launchpad)](https://bugs.launchpad.net/sparkle/+bug/228445) on' 2008-05-09:*

This is a really cool idea (thanks, Matt Gough!): add a popup menu to
the update alert for older release notes. This solves the tricky problem
of not being able to merge HTML files well without some server-side
support.

This ticket was migrated from the old trac: re #5
Originally reported by: Matt Gough

PR #1688, opened 2020-11-27
Allow an appcast to prevent the new version from being installed automatically

pierswalter:
Allow an individual appcast to override the SUAutomaticallyUpdate user defaults setting and prevent the new version from being installed automatically. This is helpful if the version referenced by the appcast is a paid upgrade that should not be applied without an explicit user decision. To use this feature, the new appcast property "sparkle:doNotAutomaticallyUpdate" shall be set to 1 or YES.

Please note that there is a corresponding pull request in the sparkle-project.github.io project.

### DIFF
--- a/Sparkle/SUAppcastItem.h
+++ b/Sparkle/SUAppcastItem.h
@@ -34,6 +34,7 @@ SU_EXPORT @interface SUAppcastItem : NSObject
 @property (copy, readonly) NSDictionary *deltaUpdates;
 @property (strong, readonly) NSURL *infoURL;
 @property (copy, readonly) NSNumber* phasedRolloutInterval;
+@property (nonatomic, readonly) BOOL doNotAutomaticallyUpdate;
 
 // Initializes with data from a dictionary provided by the RSS class.
 - (instancetype)initWithDictionary:(NSDictionary *)dict;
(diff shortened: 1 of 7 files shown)


pierswalter:
I've now rewritten the configuration logic as discussed in the sparkle-project.github.io pull request #61.
Instead of a flag that enables the new feature, the appcast can now contain an optional version number (`<sparkle:minimumAutoupdateVersion>`).

If this version has been configured and if the current app's `CFBundleVersion` compares to be smaller, the automatic update will be disabled and the _update available_ GUI is displayed instead.

kornelski:
Looks good. Have you tested it?

pierswalter:
Yes.
I've tested the following four cases, all of which passed.
1. `sparkle:minimumAutoupdateVersion` not used in appcast -> automatic update is performed (if enabled via user defaults)
2. `sparkle:minimumAutoupdateVersion` (126.199) < `CFBundleVersion` (126.200) -> automatic update is performed (if enabled via user defaults)
3. `sparkle:minimumAutoupdateVersion` (126.200) == `CFBundleVersion` (126.200) -> automatic update is performed (if enabled via user defaults)

4. `sparkle:minimumAutoupdateVersion` (126.201) > `CFBundleVersion` (126.200) -> automatic update is **not** performed, instead the _update available_ GUI is shown.

kornelski:
Great, thanks!

Issue #1616, opened 2020-06-19
Accessing SUSystemProfiler output to present to users


pilky:
Hey, I'm wanting to give users a way to see what data is set to my server during an update. This is possible during the initial UI, but if someone clicks through that UI or if they want to double check what data is sent (which may have changed since first enabling updates) there is no way to do that.

To fix this I want to present the data from SUSystemProfiler to the user. Unfortunately this isn't public (presumably because it requires SUHost which also isn't public). Would anyone have any objections to adding a `systemProfileArray` property to SPUUpdater that basically returns `[SUSystemProfiler systemProfileArrayForHost:self.host]`? 

I'm probably going to do that myself regardless, this is more a question of whether you would be open to me creating a PR afterwards to get it into the 2.x branch?

kornelski:
Adding that method seems fine.

pilky:
I've added a PR for this change (and a few others) at https://github.com/sparkle-project/Sparkle/pull/1690. This is my first Sparkle PR so hopefully I've got everything ok but please let me know if there are any questions or any changes needed

pierswalter:
I believe that this issue may be resolved with my PR #1687 which had been merged on Nob 27, 2020.

pilky:
@pierswalter Just as a heads up, this will conflict with my similar changes on the 2.x branch. I've named my property `systemProfileArray`, though it works a bit differently. Rather than storing the profile it generates it whenever asked. It also hooks into the additional filtering of profile keys I added to the delegate. I also switched all other locations over to calling the property rather than SUSystemProfiler directly to ensure this filtering is correctly applied everywhere

PR #1690, opened 2020-11-29
System profiler privacy and transparency

pilky:
This PR adds some fixes and enhancements to system profiling. Specifically it adds 2 things:
- Add `SPUUpdater.systemProfileArray` property. This allows developers to access this information to display to users
- Add filtering of system profile array (via `-[SPUUpdaterDelegate allowedSystemProfileKeysForUpdater:]`). This helps developers improve privacy by choosing which profile items to collect and send with update requests. This filtering happens before any additional feed parameters are added.

It also fixes 2 issues with the system profile code:
- Fix an issue with the CPU type property that caused the incorrect value to be returned for ARM/Apple Silicon
- Update a few types in code that incorrectly assumed the profile array had the type `NSArray<NSString *, NSString *>` (some values are NSNumbers)

This is partly related to https://github.com/sparkle-project/Sparkle/issues/1616 and has been implemented on the 2.x branch

This image provides an example of what these changes enable https://coppiceapp.com/images/blog/designing_for_privacy/privacy-sheet.png

### DIFF
--- a/Sparkle/SUSystemProfiler.m
+++ b/Sparkle/SUSystemProfiler.m
@@ -36,7 +36,7 @@ static NSString *const SUSystemProfilerPreferredLanguageKey = @"lang";
     return [[NSDictionary alloc] initWithContentsOfFile:path];
 }
 
-+ (NSArray<NSDictionary<NSString *, NSString *> *> *)systemProfileArrayForHost:(SUHost *)host
++ (NSArray<NSDictionary<NSString *, id> *> *)systemProfileArrayForHost:(SUHost *)host
 {
     NSDictionary<NSString *, NSString *> *modelTranslation = [self modelTranslationTable];
 
--- a/Sparkle/SUSystemProfiler.m
+++ b/Sparkle/SUSystemProfiler.m
@@ -37,7 +37,7 @@ NSString *const SUSystemProfilerPreferredLanguageKey = @"lang";
     return [[NSDictionary alloc] initWithContentsOfFile:path];
 }
 
-+ (NSArray<NSDictionary<NSString *, id> *> *)systemProfileArrayForHost:(SUHost *)host
++ (NSArray<NSDictionary<NSString *, NSString *> *> *)systemProfileArrayForHost:(SUHost *)host
 {
     NSDictionary<NSString *, NSString *> *modelTranslation = [self modelTranslationTable];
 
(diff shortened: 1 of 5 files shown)


Sparkle/SUSystemProfiler.m
```diff
@@ -57,6 +58,8 @@ @implementation SUSystemProfiler
     error = sysctlbyname("hw.cputype", &value, &length, NULL, 0);
     int cpuType = -1;
     if (error == 0) {
+        //hw.cputype is a bitmask that can contain multiple bits of info. We only want the last 2 bytes to get the architecture
```
kornelski, line 61: The comment says last 2 bytes, but it masks 1 byte.

Sparkle/SUSystemProfiler.m
```diff
@@ -36,7 +37,7 @@ @implementation SUSystemProfiler
     return [[NSDictionary alloc] initWithContentsOfFile:path];
 }
 
-+ (NSArray<NSDictionary<NSString *, NSString *> *> *)systemProfileArrayForHost:(SUHost *)host
++ (NSArray<NSDictionary<NSString *, id> *> *)systemProfileArrayForHost:(SUHost *)host
```
kornelski, line 40: If properties aren't guaranteed to be strings, you have to document what types they have. But I'd prefer to keep them all as strings for simplicity. They're going to be sent in a URL anyway.

pilky: I was a bit hesitant to change it people were assuming certain values might be NSNumbers, but I suppose this is the first time this has technically be exposed publicly so that should be fine. 

I'll get all the values converted to strings tomorrow and revert all the type declarations.

pilky: I've updated everything to strings but I've had to make some slight tweaks (they shouldn't really affect much):

1. For some reason we were dividing the CPU frequency by 1000 for the display frequency so that it would show in GHz. Given the CPU frequency and the readable label explicitly state MHz I removed that change
2. The RAM was being added as a float. In reality it seems it was always been treated as an int so it shouldn't affect anyone (unless someone desperately needs to know if someone has someone installed a fraction of a MB of RAM)
3. I've removed calls to `-description` when parameterising the feed URL as these are liable to break and are now unnecessary as everything is a string

kornelski:
Thank you

PR #1694, opened 2020-12-06
Support getting app icon from asset catalog

CharlesJS:
Currently, Sparkle 2.x displays a generic icon if the app uses an asset catalog for its app icon instead of a distinct image file in the file system. This should allow asset catalog app icons to be recognized.


kornelski:
Where is `SPARKLE_ICON_NAME` coming from?

CharlesJS:
> Thanks for the PR. It seems like a good idea.
> 
> However, I don't understand where is `SPARKLE_ICON_NAME` coming from? Does this need to be documented?

Comes from the project's common config: https://github.com/sparkle-project/Sparkle/blob/2.x/Configurations/ConfigCommon.xcconfig#L16

PR #1696, opened 2020-12-09
Hide automatic update button and skip button, depending on `SUAutomaticallyUpdate` key

lorenrh:
This PR adds 3 new keys to be able to let the developer choose what buttons should be shown in the update window. When a key doesn't exist, it will keep the same behavior as of today, showing all the buttons. The keys are the following: 

`SUHideAutomaticallyInstallUpdates` - If set to true it will hide the _Automatically check for updates_ checkbox
`SUHideSkipButton` - If set to true it will hide the _Skip this version_ button
`SUHideLaterButton` - If set to true it will hide the _Remind me later_ button

Signed-off-by: Lorena Rangel <lorena.rangel@docker.com>

### DIFF
--- a/Sparkle/SUUpdateAlert.m
+++ b/Sparkle/SUUpdateAlert.m
@@ -326,11 +326,28 @@ static NSString *const SUUpdateAlertTouchBarIndentifier = @"" SPARKLE_BUNDLE_IDE
         self.laterButton.enabled = NO;
         self.laterButton.hidden = YES;
     }
+
+    if([self automaticallyUpdateEnabled]) {
+        self.skipButton.hidden = YES;
+        self.automaticallyInstallUpdatesButton.hidden = YES;
+        [self.laterButton setTitle:SULocalizedString(@"Install later", @"Alternate title for 'Remind me later' button when automatic updates are enabled")];
+    }
+
     [self.window center];
 }
 
 - (BOOL)automaticChecksEnabled {
     NSNumber *automaticChecksEnabled = [self.host objectForKey:SUEnableAutomaticChecksKey];
+    if (automaticChecksEnabled == nil)
+    {
+        return false;
+    }
+
+    return [automaticChecksEnabled boolValue];
+}
+
+- (BOOL)automaticallyUpdateEnabled {
+    NSNumber *automaticChecksEnabled = [self.host objectForKey:SUAutomaticallyUpdateKey];
     if (automaticChecksEnabled == nil)
     {
         return false;


kornelski:
We already have functionality for this. There's SUAutomaticallyUpdate plist key to hide autoupdate options. The skip/later button is hidden if you mark an update as critical.

lorenrh:
All these options are tied to a functionality behind the scenes and specific use cases. This PR is mostly an esthetic change. When our users arrive to Sparkle's update window is because they have checked for updates manually, so the buttons don't make much sense.

In a normal workflow, we don't want to prevent any automatic updates, neither marking them as critical as the update popup will have a very different text and it will appear anytime, disrupting the user in their own workflow.

lorenrh:
Hello,

I just re-read the message and think it might have not been clearly explained our use case. We have delta updates in our application and have automatic updates enabled so the users are always up-to-date. We also rollout the updates through a canary group, so not all users get the updates at the same time.

We already have an in-app system for letting the user know that there's a new update available. But when a user is not part of the canary group, he can still get the update by clicking on a  Check for Updates button. This will add him into the canary group and the Sparkle Update dialog will appear, as an informational pop-up. That's the only time in our workflow, that the Sparkle update pop up appears.

We've tried with marking the updates as critical, but it changes the text as well making the pop-up appear anytime, which is not the desired workflow for us. We don't want to prevent individual updates just to hide the button and we would like to have our users always updating, so we wouldn't like to change `SUAutomaticallyUpdate`.

I'm happy to hear your thoughts.

kornelski:
I'm torn about this. Sadly, Sparkle doesn't have good UI customization story, so such a small change is not just a matter whether it fits your vision for your app, but whether that's a desirable for all Sparkle users, and something that Sparkle can commit to maintaining.

To me it doesn't fit other Sparkle settings (which are more semantic than visual) and your update workflow is different than what Sparkle usually does (where user is in control).

It's normal for Sparkle to display Later and Skip buttons even if user asked for updates manually. User may check release notes decide whether that's an update they wanted right now, or they can wait (e.g. it's a fix for a bug they're experiencing vs a functionality change they would rather not deal with right now). 

I assume that you've made your Sparkle always automatically update, so you're not interested in offering a choice, and think of that screen more like a start of the installation process.

But this window is meant to offer a choice (apart from exceptional situations like critical update). Having only "Install Update" makes this window weird — there's no real choice there. User may close the window, but that is less obvious and unclear what happens then.

To fit better your workflow, I propose a compromise:

* If `SUAutomaticallyUpdate` is set to `YES` in `Info.plist` (not just user defaults), then hide the "automatically install updates" checkbox and ignore user preferences. This is what documentation currently implies, but actually this has never been implemented fully, so it's kind of a bug.

* `SUAutomaticallyUpdate` is set to `YES` in `Info.plist`, then hide the Skip button.

* If automatic updates are enabled in any way (either via `Info.plist` or just user preference), then rename "Remind me Later" to "Install Later" (which is a more accurate description of what actually happens).

lorenrh:
Thanks you so much for your feedback, it's actually a perfect solution. I'll update the PR!

lorenrh:
Hello,

I've updated the PR. Whenever `SUAutomaticallyUpdate` is set to YES, it will hide the check for updates button, the skip button and will change the button title to "Install later", instead of "Remind me later". 

Let me know what you think.

kornelski:
Thank you

kornelski:
@zorgiepoo Another one to port to 2.x :)

lorenrh:
Thanks for taking a look and merging so quickly! :)

zorgiepoo:
@kornelski @lorenrh 

Sadly there looks like a couple of regressions that needs to be addressed in `master` first :( (I can handle making changes in 2.x later, so don't worry about that, and since things are different there I have to re-evaulate it).

At minimum, currently the user defaults is queried, but what we really want is to only use query the info.plist setting as we discussed. Otherwise a user that opts into automatic downloading cannot opt out of it later. Instead of creating a new method in SUUpdateAlert you should also just use `[self.host boolForInfoDictionaryKey:SUAutomaticallyUpdateKey]`

> If automatic updates are enabled in any way (either via Info.plist or just user preference), then rename "Remind me Later" to "Install Later" (which is a more accurate description of what actually happens).

The other (harder to reason about) issue is this does not account for when users don't have automatic privileges to install an update, then it really is remind me later. More so, even when an update is downloaded automatically, I believe Sparkle can remind the user about the update. I would leave this as "Remind me Later" and it saves us making a localization update.

zorgiepoo:
Also, I know this is a slightly different discussion, but the remind me later button doesn't do anything different from closing the window, so having that button in the first place is an interesting thing to wonder about.

PR #1702, opened 2020-12-28
Base internationalization

zorgiepoo:
Base internationalization for issue #1685. I'd have to look at changes needed for 2.x branch after.

There were key-mismatch issues I fixed up with Hebrew, Catalan, and some cleanup I did with Finnish but otherwise I extracted the string files from the xibs before replacing them. I added localizations to the TestApplication MainMenu.xib so you can test different languages. I also fixed issues that were preventing me from running the test app (added a shared scheme for it, disabled hardened runtime, re-signed altered app with ad-hoc signature). And changed the check localizations script to check for Base Sparkle.strings file (moved it from English -> Base).


zorgiepoo:
Added auto layout constraints to automatic update alert nib too as there were text overflowing issues there (note, this nib is obsoleted in 2.x).

kornelski:
Awesome

Issue #1482, opened 2019-10-07
Catalyst Release Notes Support


Vortec4800:
Can I use Sparkle to auto-update apps (distributed outside of the Mac App Store of course) that are built using Catalyst? Is there anything I need to do special to make it all work?

kornelski:
I haven't tried Catalyst. If it's compatible with Cocoa and can use frameworks, it should be able to use Sparkle.

samdeane:
Sparkle builds against the normal macOS sdk, so I'm pretty sure you can't use it directly from a Catalyst app.

You can probably use it by [bridging to macOS via a plugin](https://www.highcaffeinecontent.com/blog/20190607-Beyond-the-Checkbox-with-Catalyst-and-AppKit), and having the plugin link against Sparkle and create the SUUpdater object.

I'm not sure how well this will work, but I'm planning to try it. Will let you know... :)

samdeane:
Update: I think the approach I described above should work, but I've currently hit an issue.

My bridging plugin links to Sparkle, embeds the framework, and owns an `SPUStandardUpdaterController` object. It sets itself to be the updaterDelegate.

The various SU keys are in the main application's `Info.plist`, as usual.

Checking for updating appears to be working fine.

However, currently, when it finds an update, I'm getting a crash in `[SPUStandardUserDriver setUpFocusForActiveUpdateAlert]`, calling `[self.activeUpdateAlert window]`.

It seems that the system is actually crashing inside WebView:

![Stack](https://user-images.githubusercontent.com/206306/75354371-c61ea780-58a4-11ea-9519-50b23f36e990.png)

As yet, I haven't figured out what's going on here.

samdeane:
I did wonder if this was something to do with `WebView` being deprecated on 10.15, and the system log is showing

`2020-02-26 14:29:30.695271+0000 Action Status[14298:327670] -[WebView initWithCoder:]: unrecognized selector sent to instance 0x7b180006e1c0`

which looks suspiciously like WebView itself isn't there.

However, I've tried swapping out WebView for WKWebView, and I'm still getting a crash in a similar place, so I don't think it's that.

samdeane:
This crash seems to be related to using WebKit from the bridging plugin, and not specifically anything to do with Sparkle. I suspect that it may be caused by WebKit getting confused about whether or not it's running under AppKit or Catalyst.

[...]

PR #1704, opened 2020-12-31
Adopt WKWebView

zorgiepoo:
Adopt WKWebView (WebKit2) where possible.

We fallback to legacy WebView in two cases:

* On systems prior to macOS 10.11 (it may be possible to support WKWebView in 10.10, but due to an 10.11+ only API I didn't bother putting in the effort)
* For apps using the downloader XPC service which are most likely sandboxed apps without an outgoing network entitlement. WKWebView doesn't function in that case (FB6993802 - https://github.com/feedback-assistant/reports/issues/1)

Additional notes:

* Added an abstraction for using WebView or WKWebView
* Web view is created and inserted programmatically rather than from the nib which has previously caused issues (and is not allowed for WKWebView prior to 10.12 due to some NSCoding bug).
* We need to rely on WKWebView private API for drawing transparent background (FB7539179 - https://github.com/feedback-assistant/reports/issues/81). Additional notes in the code.
* We always inject a custom stylesheet for dark mode colors and default font family/size (for WKWebView). The JS user script code may be worth reviewing.
* This should also improve the Catalyst support story (#1482)

I'll file dupes to the above feedback issues. This is worthwhile back porting to 1.x after a review. I also want to personally test WKWebView on macOS 10.11 when I get the chance later next week.


kornelski:
Sweet

PR #1705, opened 2020-12-31
Eliminate assumption that user driver calls are invoked in any thread

zorgiepoo:
This reduces friction in creating a SPUUserDriver. This assumption that the calls could be made from any thread is a relic of the past of separating the user driver into a different process from the updater/scheduler, which is no longer sane or supported.

This should speed up things very slightly too.


kornelski:
Would it be possible to add an assert there that ensures the right thread calls it?

zorgiepoo:
Sprinkled checks in the standard user driver for decent coverage. Xcode's main thread UI checker should also help a bit.

PR #1706, opened 2021-01-02
Migrate to GitHub Actions CI (2.x)

zorgiepoo:
Experimental test to see if GitHub Actions CI works and if travis can be done away with.

Fixes issues and simplifies the UI test too.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
       matrix:
         xcode: ['/Applications/Xcode_12.2.app/Contents/Developer', '/Applications/Xcode_11.7.app/Contents/Developer']
     name: Build and Test Sparkle
-    runs-on: macos-latest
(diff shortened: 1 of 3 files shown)


zorgiepoo:
Got this working finally.

Notes:
* macos-latest is an alias for github's "latest" macOS that is stable / isn't a preview (it is currently on a version of macOS 10.15). This should be used.
* Added a matrix to test an Xcode 12 and Xcode 11 versions like how travis is set up currently (downside is we maintain which versions to use instead of relying on a default, but that seems like current situation anyway).
* Logs are uploaded / you can download them if one of the build / execution steps fails. You can eg examine xcresult bundles and look at where things failed.

zorgiepoo:
actually, if Xcode versions are being specified, probably macOS version should too (eg Big Sur may not properly support Xcode 11.x)

kornelski:
Thank you

PR #1708, opened 2021-01-03
Update INSTALL to include submodule init step

boyvanamstel:
The install description currently fails to notify the user that they should install the git submodule(s) present in the repo.

This causes `make release` to fail.


kornelski:
Yeah, thanks.

PR #1709, opened 2021-01-05
Adopt WKWebView (1.x)

zorgiepoo:
Some small differences between 1.x and 2.x caused pain but aside this is more or less the changes from the 2.x PR (#1704)


kornelski:
Awesome

PR #1710, opened 2021-01-05
Fix Package.swift being updated after tag

### LINKED ISSUE #1674: SwiftPM binary target checksum catch-22
I am trying to move the Sparkle dependency in the IINA project from CocoaPods- to SwiftPM-managed (iina/iina#2877). Every time when I ask Xcode to add Sparkle, it says it can not download the artifact (which I assume is the Zip archive?). Here is the output from Xcode, edited to include only Sparkle-related entries: 

> Fetching https://github.com/sparkle-project/Sparkle
> Cloning https://github.com/sparkle-project/Sparkle
> Checking out https://github.com/sparkle-project/Sparkle at 1.24.0
> Removing https://github.com/sparkle-project/Sparkle
> artifact of binary target 'Sparkle' failed download: invalid status code 404

The `404` code seems like Xcode can not find the artifact.

I tried on both Xcode 12.2 and 12.3 Beta (12C5020f).

thebluepotato:
TL/DR; This PR should fix #1674.

### What it does

Adds a simple script called `release-move-tag.sh` under `Configurations`. It runs as a separate step in `make release`.

I chose to have it as a separate script to improve readability and to allow Xcode to build the Distribution target without invoking anything tag-related.

The script checks that there have been no commits since latest tag to prevent accidentally moving an old tag. Indeed, the script expects to be run after tagging the release just before `make release`, so there should be no commits in-between.

### Tests

I've tested the following successfully:

- Bump version in ConfigCommon.xcconfig and Sparkle.podspec
- Commit and tag
- `make release` and approve moving the tag

It all seemed ok to me. All that would be left is pushing to remote and uploading the corresponding artifacts (including the ZIP for SPM).

### Up for review

I thought about adding some version checks but it seems a little overkill as it seems already safe enough. The script actually asks for user approval before moving the tag. Accordingly, `make release` requires one interaction if we are really releasing (and asks so **before** opening the directory with the assets to be uploaded). We could also remove the interaction entirely and just commit Package.swift and move the tag whenever there have been no commits since latest tag.

I've added comments that are up for review. Naturally, once the changes are approved, I'll do the same on the 2.x branch (**EDIT**: now available #1711).

### DIFF
--- a/Configurations/release-move-tag.sh
+++ b/Configurations/release-move-tag.sh
@@ -1,16 +1,23 @@
 #!/bin/bash
 set -e
 
-# Finishing up - force re-tag
+# Convenience script to automatically commit Package.swift after updating the checksum and move the latest tag
 latest_git_tag=$(git describe --abbrev=0) # gets the latest tag name
-read -p "Do you want to commit changes to Package.swift and force move tag '$latest_git_tag'? (required for Swift Package Manager release) " -n 1 -r
-echo    # (optional) move to a new line
-if [[ $REPLY =~ ^[Yy]$ ]]
-then
-    long_message=$(git tag -n99 -l $latest_git_tag) # gets corresponding message
-    long_message=${long_message/$latest_git_tag} # trims tag name
-    long_message="$(echo -e "${long_message}" | sed -e 's/^[[:space:]]*//')" # trim leading whitespace
-    git add Package.swift
-    git commit -m "Update Package.swift"
-    git tag -fa $latest_git_tag -m "${long_message}"
+commits_since_tag=$(git rev-list ${latest_git_tag}.. --count)
+if [ "$commits_since_tag" -ge 1 ]; then
+    # If there has been more than one commits since the latest tag, it's highly likely that we did not intend to do a full release
+    echo "WARNING: More than one commit ($commits_since_tag) since tag '$latest_git_tag'. Did you tag a new version?"
+    echo "Package.swift has not been commited and tag has not been moved."
+else
+    # TODO: add sanity check to see if version is actually being updated or not?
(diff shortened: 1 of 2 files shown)


Configurations/release-move-tag.sh
```diff
@@ -0,0 +1,26 @@
...
+
+# Convenience script to automatically commit Package.swift after updating the checksum and move the latest tag
+latest_git_tag=$(git describe --abbrev=0) # gets the latest tag name
+commits_since_tag=$(git rev-list ${latest_git_tag}.. --count)
+if [ "$commits_since_tag" -gt 0 ]; then
+    # If there have been commits since the latest tag, it's highly likely that we did not intend to do a full release
+    echo "WARNING: $commits_since_tag commit(s) since tag '$latest_git_tag'. Did you tag a new version?"
+    echo "Package.swift has not been committed and tag has not been moved."
+else
+    # TODO: add sanity check to see if version is actually being updated or not?
```
thebluepotato, line 12: Since the script does not even proceed if there have been commits since the latest tag (i.e., we have not tagged the latest changes), I don't think it's actually necessary to check that the version number has changed. That check would actually be a little tricky.

Configurations/release-move-tag.sh
```diff
@@ -0,0 +1,26 @@
...
+    read -p "Do you want to commit changes to Package.swift and force move tag '$latest_git_tag'? (required for Swift Package Manager release) " -n 1 -r
```
thebluepotato, line 13: The script asks for user approval here. It would be entirely feasible to move the tag automatically when there have been no commits. Indeed, not much damage can be done as if there are actually no changes to Package.swift (e.g., we're running the script on its own for some reason), `git add Package.swift` fails with an error and nothing else is done.
Should I remove interaction in favour of automaticity? What do you think?

kornelski: Let's keep a check for now

PR #1713, opened 2021-01-06
Don't initialize the webview if we don't show release notes

vincode-io:
This change enables Catalyst apps to embed Sparkle in a Plugin as long as they have `SUShowReleaseNotes` set to `false`.


zorgiepoo:
This is a good idea, I was just thinking this.

Can you make sure we don't addSubview: in windowDidLoad when self.webView.view == nil? Better yet, maybe this code instantiating the web view should just move to windowDidLoad next to the code that inserts the subview and sets the frame. No need to set the .hidden property anymore too.

vincode-io:
Thanks for the advice.  I've made the suggested changes.

PR #1715, opened 2021-01-06
Don't initialize the webview if we don't show release notes

tfonfara:
This change enables Catalyst apps to embed Sparkle in a Plugin as long as they have SUShowReleaseNotes set to false.

Adoption of #1713 for version 1.x (original issue #1482)


kornelski:
Looks good. Thank you.

PR #1720, opened 2021-01-10
Readme sync

zorgiepoo:
Read me sync from 1.x and some internal doc changes.


zorgiepoo:
Hm I don't know if the CI runs were canceled but I guess it doesn't matter for this.

PR #1727, opened 2021-01-18
Add pull request template and update issue templates

zorgiepoo:
These changes require development changes to be made on 2.x branch first. This is separate from changing the default branch, which can happen and be done independently.

On the repo level, I also want to protect master and 2.x from direct pushes because inevitably it's tempting to push changes directly that should be reviewed -- and changes that get integrated that way are less pleasant to track.

And I want to require a minimum of two project members to review PR changes (with a reasonable bypass for trivial bug-fixes, localization or doc updates or critical things that should rarely be the case). With that, I will put a commitment in reviewing changes here, and this would result in less work for project members and better quality review.

(Note PR / issue templates must be pushed to default branch, currently master).


zorgiepoo:
Thanks. As discussed @kornelski I enabled two changes:

1. Two approvers (usually us) are required for future and on-going PRs
2. Locked master and 2.x from direct pushes (so create PRs)

In theory we can probably bypass this for special cases, but in spirt let's try not to normally and see how this works out. (One exception may be if a PR comes from either one of us, then only 1 approval is sufficient)

I'm going to clean up some of the other stale PRs.

PR #1734, opened 2021-01-24
Alert adjustments 1x

zorgiepoo:
Fix fallout issues from #1696 for `master`: 
* Constraints not being updated when automatic checkbox is hidden
* Setting being read from user defaults rather than Info.plist which broke the user toggling the automatic checkbox setting in other cases
* Showing Install Later for cases where app cast prevents automatic installing of updates
* Capitalization title fixes

Some simplification / sync with 2.x too (#1733)

I'm not a huge fan with the growing complexity here of additional features and the "Install Later" change which I may not preserve for 2.x.

## Checklist:

- [x] My change is being backported to master branch (Sparkle 1.x), if deemed necessary. Please create a separate pull request for 1.x, should it be backported.

## Type of change

## Testing

- [x] Sparkle Test App (with testing lots of different boolean flags in Info.plist)
- [ ] Unit Tests
- [ ] My own app
- [ ] Other (please specify)

macOS version tested: macOS 11.1 (20C69)

### DIFF
--- a/Sparkle/Base.lproj/SUUpdateAlert.xib
+++ b/Sparkle/Base.lproj/SUUpdateAlert.xib
@@ -174,11 +174,6 @@ DQ
                             <font key="font" metaFont="smallSystem"/>
                         </buttonCell>
                         <connections>
-                            <binding destination="-2" name="hidden" keyPath="allowsAutomaticUpdates" id="141">
-                                <dictionary key="options">
-                                    <string key="NSValueTransformerName">NSNegateBoolean</string>
-                                </dictionary>
-                            </binding>
                             <binding destination="93" name="value" keyPath="values.SUAutomaticallyUpdate" id="135"/>
                         </connections>
                     </button>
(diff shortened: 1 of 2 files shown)


zorgiepoo:
The more I'm thinking about it, I really do *not* want to use `Install Later`. I want to use `Install on Quit` (where guarantees are strong enough and it's a string we already use) or something dismissive like providing an option to hide the button, or just not showing the button, or reverting back to "Remind Me Later" which is still accurate. Supporting "Install Later" as to imply "this update will be installed at some point without user's manual action at some point" is being very problematic.

zorgiepoo:
I believe the right solution here really seems to be to offer an option to hide the `later` button instead, which in all contexts (install on quit / remind me later) is always functionally equivalent to closing the window and a "no-op" UI decoration.

zorgiepoo:
Hiding the skip button for automatic updates is actually dangerous too when you encounter an appcast with a paid version that prevents automatic downloads.

kornelski (approved):

PR #1384, opened 2019-03-28
Allow updating to previous versions if newest appcast item is blocked due to phased rollout

fjaeger:
The current version of phased rollouts would not find other valid items if the latest one is not ready yet based on the phased rollout group. This PR changes this behavior by using the system's rollout group when finding valid appcast items.

PR #1736, opened 2021-01-25
Port phased group rollout to 2.x

zorgiepoo:
Ports phased group updates to Sparkle 2 (from #1381, #1384, #1385), cc @fjaeger 

There is a couple significant issues with the implementation in 1.x (`master`) that this also addresses:

* `SUUpdateGroupIdentifierKey` was being read from Info.plist instead of user defaults, which basically as far as I can see results in a new random update group being used every time a schedule check occurs (this regressed in #1384)
* Like [preventing automatic updates](https://sparkle-project.org/documentation/publishing/) this needs to be an app cast element, not an enclosure attribute. #1385 was wrong. Otherwise delta update items won't inherit the rollout phase and it doesn't make sense for the delta items to have different rollout.

There's another bug with the updater delegate handling of app cast items, but it's not specific to phased updates and I'll handle it separately at some point.

## Checklist:

- [x] My change is being backported to master branch (Sparkle 1.x), if deemed necessary. Please create a separate pull request for 1.x, should it be backported. **I can back port the diverging changes from 1.x when I get the time, but not for a few days most likely.**

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected) - **It uses an app cast element instead of an enclosure attribute from 1.x**
- [x] My change is or requires a documentation or localization update - **It'd be nice to document this feature on the [website](https://github.com/sparkle-project/sparkle-project.github.io) next to the preventing auto downloading of updates section, although only recommending a bug-free version of it**

## Testing

- [x] Sparkle Test App
- [x] Unit Tests - **ported the app cast unit test**
- [ ] My own app
- [ ] Other (please specify)

I tested making sure rollout is created, rollout period persists until a download is finished and a new one is created, and tested multiple runs with a date to make sure sometimes I was in the group and other times I was not.

macOS version tested: macOS 11.1 (20C69)


fjaeger:
Thanks for including me in this PR review!

I think the idea of **not** reading the group identifier from the user defaults was to kind of randomize the rollout group for each individual user for each update. This means that user X is not always receiving an update only after the same number of hours. I have no preference one the behavior here.

zorgiepoo:
Great to have you here :). The refactor you made in #1384 did not make it obvious that you actually intended to do this.

So in master we have:

```
#define NUM_UPDATE_GROUPS 7
+ (NSUInteger)updateGroupForHost:(SUHost*)host {
    NSNumber* updateGroupIdentifier = [self updateGroupIdentifierForHost:host];
    return ([updateGroupIdentifier unsignedIntValue] % NUM_UPDATE_GROUPS);
}

+ (NSNumber*)updateGroupIdentifierForHost:(SUHost*)host {
    NSNumber* updateGroupIdentifier = [host objectForInfoDictionaryKey:SUUpdateGroupIdentifierKey];
    if(updateGroupIdentifier == nil) {
        updateGroupIdentifier = [self setNewUpdateGroupIdentifierForHost:host];
    }

    return updateGroupIdentifier;
}

+ (NSNumber*)setNewUpdateGroupIdentifierForHost:(SUHost*)host {
    unsigned int r = arc4random_uniform(UINT_MAX);
    NSNumber* updateGroupIdentifier = @(r);

    [host setObject:updateGroupIdentifier forUserDefaultsKey:SUUpdateGroupIdentifierKey];

    return updateGroupIdentifier;
}
```

But nobody should be setting `SUUpdateGroupIdentifierKey` in Info.plist, I think this was an incident. when a new group identifier is set, it is later ignored because of this difference. The current code sets a new group identifier after a download finishes (which I respect here too).

> There's another bug with the updater delegate handling of app cast items, but it's not specific to phased updates and I'll handle it separately at some point.

The unit tests are failing and fixing this bug is the easiest way for me to fix this unfortunately, but might be a few days before I get to that. Essentially, the app cast items should be filtered for max/min OS and phased group testing before asking the delegate what the best app cast item is.

zorgiepoo:
Updated with filtering app cast items before doing comparisons, should be all good here.

Sparkle/SUAppcastItem.m
```diff
@@ -169,6 +175,19 @@ - (BOOL)isMacOsUpdate
...
 
+- (NSDate *)date
+{
+    if (self.dateString == nil) {
+        return nil;
+    }
+    
+    NSDateFormatter* dateFormatter = [[NSDateFormatter alloc] init];
+    dateFormatter.locale = [[NSLocale alloc] initWithLocaleIdentifier:@"en_US"];
+    dateFormatter.dateFormat = @"E, dd MMM yyyy HH:mm:ss Z";
```
zorgiepoo, line 186: generate_appcast uses..
`EEE, dd MMM yyyy HH:mm:ss ZZ`

In my testing, I have been using EEE (eg `Mon`) for pubDate in the test app cast, so in practice it might not matter.

zorgiepoo:
For reference, added a minor fix that computes current date only once for all app cast items.

PR #1743, opened 2021-01-30
Deprecate installUpdatesIfAvailable

zorgiepoo:
Deprecate installUpdatesIfAvailable


zorgiepoo:
Rationale in previous issues #864 , header documentation of this method that people aren't likely to read, and I could have sworn there were other issues about you saying we shouldn't have added it but I can't find it. This method is not being added to 2.x.

kornelski (approved):

PR #1746, opened 2021-01-31
Backport flat non-archived package support from #1745 (for 1.x)

zorgiepoo:
Add flat package support from #1745

Differences:
* Packages don't use prevalidation in 1.x
* Check for `sparkle_interactive` rather than installation type attribute which doesn't exist in 1.x

Only guided package installs are supported for flat packages.

## Checklist:

- [x] My change is being backported to master branch (Sparkle 1.x), if deemed necessary. Please create a separate pull request for 1.x, should it be backported.

## Type of change

## Testing

- [x] Sparkle Test App
- [x] Unit Tests
- [x] My own app
- [ ] Other (please specify)

Tested same cases in #1745

macOS version tested: 11.1 (20C69)


Sparkle/SUFlatPackageUnarchiver.m
```diff
@@ -0,0 +1,64 @@
...
+
++ (BOOL)canUnarchivePath:(NSString *)path
+{
+    return [path.pathExtension isEqualToString:@"pkg"];
+}
+
+// Note in 2.x this method returns YES, but 1.x does not support pre-validation of pkgs
++ (BOOL)mustValidateBeforeExtraction
+{
+    return NO;
```
kornelski, line 34: Isn't that a security vulnerability? For other formats, DSA is checked before extraction. If it's not extracted, what does check DSA?

Can you add a test that a bare installer isn't accepted without a (Ed)DSA signature or with an invalid signature?

zorgiepoo: No (I wrote the validation logic long ago by the way), this is indicating if validation should be done before extraction, which in 1.x is only done for binary delta files which are considered "unsafe". In 2.x, it is also done for pkg because if a pkg is archived 2.x mandates developer to specify installation type in app cast item. This is not a security concern for pkg's, it's an efficiency test; if validation fails there's no need to extract the archive. In this specific case of SUFlatPackageUnarchiver it doesn't matter because there's nothing to really extract. It can't return YES here anyway because code later down in the installer expects pre-validation not to be supported for pkg (for 1.x).

There are unarchiver tests I added and I did manual tests checking invalidate signature does indeed fail.

zorgiepoo: I mean I could have it return YES but then I'd have to store an additional ivar to recall encountering the flat package and change `-validateWithUpdateDirectory:` logic to allow revalidation for (only non-archived) packages, but this seems complicated for 1.x, and 2.x has the proper approach anyway.

tldr; no security implication other than for binary delta files in 1.x, validated bad signature fails.

zorgiepoo: It's never done for normal app installs by the way due to code signing checks, things need to be extracted anyway.

zorgiepoo:
I forgot about mpkg, will probably have to add that if that's a flat-package too.

zorgiepoo:
Added mpkg support performing a similar test in #1758

kornelski (approved):

PR #1747, opened 2021-02-01
Fix minor unimplemented preventsAutoupdate synthesize warning

zorgiepoo:
Minor synthesize compiler warning bug I introduced.


ks1707 (commented):
-

PR #1778, opened 2021-02-21
Revert logic that prevented users from disabling automatic downloads

zorgiepoo:
In 2.x, updates can be downloaded automatically and presented later even if they can't be installed automatically without user intervention. Users need to in control and be able to disable automatic downloads regardless of developer's Info.plist intention.

## Checklist:

- [ ] My change is being backported to master branch (Sparkle 1.x), if deemed necessary. Please create a separate pull request for 1.x, should it be backported.

## Type of change

## Testing

- [x] Sparkle Test App
- [ ] Unit Tests
- [ ] My own app
- [ ] Other (please specify)

Tested turning automatic downloads on for Test App.

macOS version tested: 11.2 (20D64)

### DIFF
--- a/Sparkle/SUUpdateAlert.m
+++ b/Sparkle/SUUpdateAlert.m
@@ -392,10 +392,7 @@ static NSString *const SUUpdateAlertTouchBarIndentifier = @"" SPARKLE_BUNDLE_IDE
     
     // When we show release notes, it looks ugly if the install buttons are not closer to the release notes view
     // However when we don't show release notes, it looks ugly if the install buttons are too close to the description field. Shrugs.
-    // Automatic downloads is enabled by developer if they set SUAutomaticallyUpdateKey in Info.plist,
-    // rather than the user toggling the setting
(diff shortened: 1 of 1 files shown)


kornelski:
What's the logic here in the end? 

Last time I checked it required user to opt-in before any automatic downloads happen, which authors don't like.

I think we should take into account a "no-preference" case where downloads can default to enabled if author set to, and then disabled with user's opt-out.

zorgiepoo:
> I think we should take into account a "no-preference" case where downloads can default to enabled if author set to, and then disabled with user's opt-out.

This is exactly what this PR does. If the developer opts to enabling automatic downloads, the user can opt out.

Previously we made a change where users could not opt out if developers opted in with Info.plist. This is reverting that.

In 2.x, especially, updates can be automatically downloaded and presented to the user later if authorization privileges are needed so it's more problematic there. 1.x won't try to enable automatic downloads in that case.

kornelski (approved):

PR #1796, opened 2021-03-11
Spelling correction

homebysix:
Minor spelling correction in markdown file. No functional changes to code.


kornelski (approved):

zorgiepoo:
This is actually a dependency which seems to be unused and was intended to be ran in CI. The typo in any case should be fixed [upstream](https://github.com/nst/objc_dep) before here. Then we should either integrate in CI again and possibly add it as a git submodule or remove the dependency. Btw, 2.x is the main development branch now instead of master.

homebysix:
Sounds good, I'll submit upstream.

PR #1801, opened 2021-03-22
Update and replace bundle directories even more atomically on apfs

### LINKED ISSUE #1416: Atomic updates behavior
What logic does Sparkle follow to actually update an app? Does it install the new app, then uninstall the previous one? Does it uninstall the previous one then install the new one? Or overwrite existing files?

If the power cord is pulled from the computer while Sparkle is updating, what's the expected outcome?

Thanks!

zorgiepoo:
With these changes, it should not be possible for neither an old or new bundle to be at the destination. That is, the old bundle directory is replaced by the new one in one step (instead of us moving the old bundle away, and putting the new bundle in its place).

We also perform temporary directory cleanup after final installation stage and relaunching an app is performed because we can save some time (150 ms+ in one of my cases with VLC).

Fixes #1416

## Checklist:

## Type of change

## Testing

- [x] Sparkle Test App
- [ ] Unit Tests
- [x] My own app
- [x] Other (please specify): injecting framework into VLC

Tested updating on apfs volume, non-apfs USB drive, apfs USB drive, SMB mount, tested with `SPARKLE_NORMALIZE_INSTALLED_APPLICATION_NAME` on. Note in normalization, non-apfs volume, or old OS cases, we fall back to legacy path of moving old bundle aside which is what we do today.

macOS version tested:
11.2.3 (20D91)
10.14 latest (VM)
10.9 latest (VM)

### DIFF
--- a/Autoupdate/AppInstaller.m
+++ b/Autoupdate/AppInstaller.m
@@ -552,6 +552,7 @@ static const NSTimeInterval SUDisplayProgressTimeDelay = 0.7;
             
             NSError *thirdStageError = nil;
             if (![self.installer performFinalInstallationProgressBlock:nil error:&thirdStageError]) {
+                [self.installer performCleanup];
                 self.installer = nil;
                 
                 dispatch_async(dispatch_get_main_queue(), ^{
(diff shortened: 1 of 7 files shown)


kornelski (approved):

kornelski:
It's a bit risky that we don't have pre-10.13 machine in the CI

zorgiepoo:
Our CI is coupled with the development environment we support rather than deployment environment (i.e, test running without building anything), which is roughly N - 1 (GitHub actions doesn't support 10.14 as far as I know). Sadly, other than some static analysis, anyone that supports old deployment targets may need to manually test on old deployments every once in a while, as CI infrastructure and decoupling development environment isn't going to happen.

PR #1802, opened 2021-03-22
Add quicker & more reliable updating to features

zorgiepoo:
Add quicker & more reliable updating to features.

Sparkle 2 is perceivably faster at relaunching an application (it saves an extra process spawn / app launch cycle because the installer is already running at the point the user wants to relaunch). 1.x branch seems to not validate the update async as well, delaying the start of installation. 2.x also doesn't show status window if the install is instant.

Sample attached using VLC:
https://user-images.githubusercontent.com/857267/111960741-5bdb3500-8aad-11eb-84ef-36a1a74494e9.mov

Two points: look at how fast old app disappears and new app appears animated in Dock, and look at how fast app terminates when Install and Relaunch is clicked / activated.

The auth execution model in 1.x using fileop is just built off of sticks and stones to me regarding reliability, and was replaced by a more robust architecture of deciding to run the installer (Autoupdate) as root or not (and not shelling out with `AuthorizationExecuteWithPrivileges`)

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,15 +4,15 @@
 * Support for writing custom user interfaces
 * Command line utility to update Sparkle-based applications and bundles
 * More modern architecture that moves extraction, validation, and installation into a submitted launchd agent/daemon.
-* Faster updates that are relaunched quicker
-* More robust updates that require authorization
+* Faster installs with shorter update/relaunch times.
+* More robust updates that require user authorization.
 * Adoption of atomic-safe bundle replacement updates leveraging APFS
 * Decoupled AppKit and UI logic in the framework from core functionality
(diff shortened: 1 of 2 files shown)


zorgiepoo:
To clarify and after thinking about it more on efficiency, 1.x architecture actually can't start validation of update sooner due to potential concern of the download/update being hampered, whereas 2.x moves the archive to a safer place before it even starts extraction. 1.x also can't do any operations that might require authorization before starting Autoupdate whereas 2.x already has authorization by this point. Further optimizations in 2.x architecture in the future are possible for making update-relaunch time shorter like releasing quarantine ahead of time as one example.

zorgiepoo:
Updated / prettified changelog, updated status / helping on readme a bit.

kornelski (approved):

PR #1809, opened 2021-03-29
Universally start update cycle one runloop cycle later

zorgiepoo:
Universally start update cycle one runloop later. Before it was only done for SUUpdater stub and SPUStandardUpdaterController, but now it's done for SPUUpdater which both of those classes use. This is to avoid UI being shown too early which is too easy to do, and gives a chance for developer to check for updates immediately instead of update cycle starting if using SPUUpdater directly. Basically mirrors behavior in 1.x (initially I wanted to avoid deferral behavior in the core updater, but now I'm realizing this was a bad idea).

Improved code that signals when updates can be checked too.

## Checklist:

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected) - Logs are outputted in the case anyone was trying to check for updates before starting the updater. Now you need to check for updates after starting the updater which makes more sense, if you want to bypass the scheduler.

## Testing

- [x] Sparkle Test App
- [ ] Unit Tests
- [x] My own app
- [x] Other (please specify) - sparkle-cli

Tested sparkle test app still works, tested permission prompt still works, tested sparkle-cli still works with and without `--check--immediately`, tested prefpane plugin.

macOS version tested: 11.2 (20D64)

### DIFF
--- a/Sparkle/SPUUpdater.m
+++ b/Sparkle/SPUUpdater.m
@@ -147,6 +147,7 @@ NSString *const SUUpdaterAppcastNotificationKey = @"SUUpdaterAppCastNotification
     }
     
     self.startedUpdater = YES;
+    [self setCanCheckForUpdates:YES];
     
     // Start updater on next update cycle so we make sure the application invoking the updater is ready
     // This also gives the developer a cycle to check for updates before Sparkle's update cycle scheduler kicks in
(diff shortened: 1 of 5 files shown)


Sparkle/SPUUpdater.m
```diff
@@ -261,6 +266,10 @@ - (NSString *)description { return [NSString stringWithFormat:@"%@ <%@>", [self
 
 - (void)startUpdateCycle
 {
+    if (![self canCheckForUpdates]) {
+        return;
```
kornelski, line 270: when would that happen? maybe log a warning?

kornelski (approved):

Sparkle/SPUUpdater.m
zorgiepoo: It's intentional behavior. Caller can startUpdater:, and then eg checkForUpdates before startUpdateCycle is called (on next runloop cycle). Once checkForUpdates is invoked, you can't check for updates until that operation is completed. Some comments are in startUpdater: which calls this method.

PR #1818, opened 2021-04-11
Defer requiring feed URL until checking for updates

zorgiepoo:
Defer requiring/validating feed URL until checking for updates. Also don't automatically reset update cycle if updater hasn't been started yet.

This avoids issues if developer doesn't supply feed URL before starting the updater but provides one before the automatic update cycle starts. Which I didn't consider when making #1809.

## Checklist:

## Type of change

## Testing

- [x] Sparkle Test App
- [ ] Unit Tests
- [x] My own app
- [ ] Other (please specify)

Tested code in sample app that sets properties like automaticallyDownloadsUpdates and invokes checkForUpdatesInBackground after implicitly starting updater but before automatic update cycle is started using SPUStandardUpdaterController or (deprecated) SUUpdater.

### DIFF
--- a/Sparkle/SPUUpdater.m
+++ b/Sparkle/SPUUpdater.m
@@ -142,7 +142,7 @@ NSString *const SUUpdaterAppcastNotificationKey = @"SUUpdaterAppCastNotification
 
 - (BOOL)startUpdater:(NSError * __autoreleasing *)error
 {
-    if (![self checkIfConfiguredProperly:error]) {
+    if (![self checkIfConfiguredProperlyAndRequireFeedURL:NO error:error]) {
         return NO;
     }
     
@@ -152,7 +152,9 @@ NSString *const SUUpdaterAppcastNotificationKey = @"SUUpdaterAppCastNotification
     // Start updater on next update cycle so we make sure the application invoking the updater is ready
     // This also gives the developer a cycle to check for updates before Sparkle's update cycle scheduler kicks in
     dispatch_async(dispatch_get_main_queue(), ^{
-        [self startUpdateCycle];
+        if ([self canCheckForUpdates]) {
+            [self startUpdateCycle];
+        }
     });
     
     return YES;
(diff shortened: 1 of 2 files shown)


Sparkle/SPUUpdater.m
```diff
@@ -201,30 +198,44 @@ - (BOOL)checkIfConfiguredProperly:(NSError * __autoreleasing *)error
...
+    }
+    
+    if (feedURL != nil) {
+        servingOverHttps = [[[feedURL scheme] lowercaseString] isEqualToString:@"https"];
+        if (!servingOverHttps) {
+            BOOL foundXPCPersistentDownloaderService = NO;
+            BOOL foundATSPersistentIssue = [self checkATSIssueForBundle:SPUXPCServiceBundle(@DOWNLOADER_BUNDLE_ID) getBundleExists:&foundXPCPersistentDownloaderService];
+            
+            NSBundle *mainBundle = [NSBundle mainBundle];
+            BOOL foundATSMainBundleIssue = NO;
```
kornelski, line 220: Maybe just drop support for ATS exceptions and require HTTPS all the time?

Sparkle/SPUUpdater.m
```diff
@@ -602,9 +606,8 @@ - (void)resetUpdateCycle
         return; // not even ready yet
     }
     
-    [self cancelNextUpdateCycle];
-    
     if (self.canCheckForUpdates) {
+        [self cancelNextUpdateCycle];
         [self scheduleNextUpdateCheck];
```
kornelski, line 611: Maybe scheduleNextUpdateCheck should cancel itself?

kornelski (approved):

Sparkle/SPUUpdater.m
zorgiepoo: We may be able to simplify this and remove the ATS checks and just say feed may need to be https. I don't think it makes sense to always require https as the system prods that (it makes more sense to require (Ed)DSA imo). The test app uses http too.

Sparkle/SPUUpdater.m
zorgiepoo: Maybe. They're not always called together currently so not refactoring that at the moment.

PR #1827, opened 2021-04-16
Show the present update in focus when checkForUpdates is invoked

zorgiepoo:
When the update is shown to the user, the menu item for checking for updates is no longer disabled. Instead it is enabled, and when invoked, will bring the update to utmost focus back to the user again.

This is consistent with other options (eg Preferences...) that bring up windows. Improved user experience here is that 1) updates are now less "lost" and 2) checking for updates now does something when update is present (especially nice if developer forgets to enforce menu item validation in a nibless approach).

We (re)define properties on SPUUpdater:

canCheckForUpdates - to mean that the user can checkForUpdates (start a new update check or show the already present update frontmost)

sessionInProgress - to mean if Sparkle's internal driver / scheduler is running (i.e, downloading appcast or update, showing update, starting installation).

I removed some setCanCheckForUpdates methods that I added/anticipated recently but in reality were never used. I wanted to make this change so that the updater & user driver has correct or reasonable API and user experience here going forward.

## Checklist:

## Type of change

This shouldn't be a breaking change. The recommendation / approach before was to use canCheckForUpdates property for menu item validation; that stays the same.

## Testing

- [x] Sparkle Test App
- [ ] Unit Tests
- [x] My own app
- [x] Other (please specify) - ImageOptim

Tested:
When update check is occurring, menu item is disabled.
When update is shown, menu item is enabled, and when invoked brings update alert up front. If the install button had no keyboard shortcut (from non-initiated check), it's re-activated again.
When update is extracting or installing, menu item is disabled.
Tested app using SUUpdater (ImageOptim)
Tested one of my apps that didn't do any menu item validation, and checkForUpdates was ignored in appropriate scenarios.

### DIFF
--- a/Sparkle/SPUAutomaticUpdateDriver.m
+++ b/Sparkle/SPUAutomaticUpdateDriver.m
@@ -89,6 +89,11 @@
     }
 }
 
+- (BOOL)showingUpdate
+{
+    return NO;
+}
+
 - (void)installerDidFinishPreparationAndWillInstallImmediately:(BOOL)willInstallImmediately silently:(BOOL)willInstallSilently
 {
     self.willInstallSilently = willInstallSilently;
--- a/Sparkle/SPUProbingUpdateDriver.m
+++ b/Sparkle/SPUProbingUpdateDriver.m
@@ -60,6 +60,11 @@
     [self abortUpdate];
 }
 
+- (BOOL)showingUpdate
+{
+    return NO;
+}
+
 - (void)basicDriverIsRequestingAbortUpdateWithError:(nullable NSError *)error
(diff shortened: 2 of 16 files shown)


zorgiepoo:
Argh, there's no reason check for updates should not bring the downloading / installing window up front too. So I guess I'm not satisfied with this approach yet.

zorgiepoo:
Alright, I made a little change to include other UI (such as permission prompt, download/extraction, install status) here and I think I'm happy with it. It's now really up to the user driver to show the active element in focus.

kornelski (approved):

PR #1836, opened 2021-04-23
Add private_key parameter in sign_update

jaroslavas:
Add second parameter in `sign_update ` tool to use key from pem file instead of the keychain. 

It helps when builds are made on CD machine where user can't enter a password to unlock keychain.

## Checklist:

## Type of change

## Testing

- [ ] Sparkle Test App
- [ ] Unit Tests
- [ ] My own app
- [x] Other (please specify)

Compared signature generated by using key from the keychain and from file - both are the same.
Tried updating app with a signature generated by using key file - app updates successfully.

macOS version tested: 10.15

### DIFF
--- a/sign_update/main.swift
+++ b/sign_update/main.swift
@@ -21,7 +21,7 @@ func findKeys() -> (Data, Data) {
     if res == errSecSuccess, let encoded = item as? Data, let keys = Data(base64Encoded: encoded) {
         return (keys[0..<64], keys[64..<(64+32)])
     } else if res == errSecItemNotFound {
-        print("ERROR! Signing key not found. Please run generate_keys tool first.")
+        print("ERROR! Signing key not found. Please run generate_keys tool first or provide path to pem file as a second parameter.")
     } else if res == errSecAuthFailed {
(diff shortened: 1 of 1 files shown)


kornelski:
This is fine, but for `generate_appcast` we take the key directly as a command-line argument. We don't use `pem` any more. That was for the old DSA. New EdDSA has such short keys that they can be just command-line args.

zorgiepoo:
It may also be good idea to scan for `-s` like `generate_appcast` uses (thinking for when we may adopt a proper argument parser later).

zorgiepoo:
I don't mind still having an option to take a file though, if you want to add two options..

jaroslavas:
OK, I've added both options with flags. 
I hadn't time for implementing proper arguments parsing like in `generate_appcast` though, so it's quite simple, but in my opinion that should be enough.

kornelski (approved):

zorgiepoo (approved):

zorgiepoo:
Thanks!

jaroslavas:
Thanks to you guys for this amazing library!

PR #1837, opened 2021-04-25
Clean up downloads better

zorgiepoo:
Clean up downloads more immediately when they aren't needed.

* Clear download if extraction fails (esp if delta update fails to apply)
* Clean up download in framework if already downloaded (but not installer-started) update to be resumed is skipped
* Fix download cancellation when using downloaded XPC Service + clean up download driver only at end of scheduler cycle
* Remove empty temporary directory that was containing agent / progress tool on clean up

We already have some fallback code that cleans up old temp cache directories when starting new downloads / installs, but we can do better job to make sure directories are cleaned up more promptly.

## Checklist:

## Type of change

## Testing

- [x] Sparkle Test App
- [ ] Unit Tests
- [x] My own app
- [ ] Other (please specify)

Tested download is cleared upon skipping download resumed update, download is cleaned on extraction failure, regular installation with test app, cancellation of downloading update, downloader XPC Service with these cases.

### DIFF
--- a/Downloader/SPUDownloader.m
+++ b/Downloader/SPUDownloader.m
@@ -118,6 +118,10 @@ static NSString *SUDownloadingReason = @"Downloading update related file";
 
 - (void)removeDownloadDirectory:(NSString *)directoryName bundleIdentifier:(NSString *)bundleIdentifier
 {
+    // Only take the directory name and compute most of the base path ourselves
+    // This way we do not have to send/trust an absolute path
+    // The downloader instance that creates this temp directory isn't necessarily the same as the one
+    // that clears it (eg upon skipping an already downloaded update), so we can't just preserve it here too
     dispatch_async(dispatch_get_main_queue(), ^{
         NSString *rootPersistentDownloadCachePath = [self rootPersistentDownloadCachePathForBundleIdentifier:bundleIdentifier];
         if (rootPersistentDownloadCachePath != nil) {
--- a/Downloader/SPUDownloader.m
+++ b/Downloader/SPUDownloader.m
@@ -111,12 +111,25 @@ static NSString *SUDownloadingReason = @"Downloading update related file";
     }
 }
 
-- (void)cancelDownload
+- (NSString *)rootPersistentDownloadCachePathForBundleIdentifier:(NSString *)bundleIdentifier
 {
-    [self cleanup];
+    return [[SPULocalCacheDirectory cachePathForBundleIdentifier:bundleIdentifier] stringByAppendingPathComponent:@"PersistentDownloads"];
 }
 
-- (void)cleanup
+- (void)removeDownloadDirectory:(NSString *)directoryName bundleIdentifier:(NSString *)bundleIdentifier
+{
+    dispatch_async(dispatch_get_main_queue(), ^{
+        NSString *rootPersistentDownloadCachePath = [self rootPersistentDownloadCachePathForBundleIdentifier:bundleIdentifier];
+        if (rootPersistentDownloadCachePath != nil) {
+            NSString *sanitizedDirectoryName = directoryName.lastPathComponent;
+            NSString *tempDir = [rootPersistentDownloadCachePath stringByAppendingPathComponent:sanitizedDirectoryName];
+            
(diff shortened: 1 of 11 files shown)


Sparkle/InstallerProgress/InstallerProgressAppController.m
```diff
@@ -193,8 +193,21 @@ - (void)cleanupAndExitWithStatus:(int)status error:(NSError * _Nullable)error __
...
+    if (![[NSFileManager defaultManager] removeItemAtPath:bundlePath error:&theError]) {
         SULog(SULogLevelError, @"Couldn't remove agent bundle: %@.", theError);
+    } else {
+        // There should be nothing else in the parent temporary directory given to us,
+        // so let us try to remove it. Note rmdir() will fail if there are unexpectably other
+        // items present
+        NSString *parentDirectory = bundlePath.stringByDeletingLastPathComponent;
+        const char *fileSystemRepresentation = parentDirectory.fileSystemRepresentation;
+        if (fileSystemRepresentation != NULL) {
+            if (rmdir(fileSystemRepresentation) != 0) {
```
kornelski, line 207: why isn't FileManager handling this properly?

Sparkle/SPUCoreBasedUpdateDriver.m
```diff
@@ -285,6 +306,12 @@ - (void)installerDidStartInstalling
     }
 }
 
+- (void)installerDidStartExtracting
+{
+    // The installer has moved the archive and no longer needs the download directory
```
kornelski, line 311: Is it guaranteed to move at the start? Shouldn't it be deleted after the end in case the installer implementation still uses the file?

Downloader/SPUDownloader.m
```diff
@@ -111,12 +111,25 @@ - (void)enableAutomaticTermination
...
 }
 
-- (void)cleanup
+- (void)removeDownloadDirectory:(NSString *)directoryName bundleIdentifier:(NSString *)bundleIdentifier
+{
+    dispatch_async(dispatch_get_main_queue(), ^{
+        NSString *rootPersistentDownloadCachePath = [self rootPersistentDownloadCachePathForBundleIdentifier:bundleIdentifier];
+        if (rootPersistentDownloadCachePath != nil) {
+            NSString *sanitizedDirectoryName = directoryName.lastPathComponent;
+            NSString *tempDir = [rootPersistentDownloadCachePath stringByAppendingPathComponent:sanitizedDirectoryName];
```
kornelski, line 125: This seems like duplicated code. If that dir is expected to exist, there must be another function somewhere that builds the path this way.

Sparkle/InstallerProgress/InstallerProgressAppController.m
zorgiepoo: Framework originally creates this temp directory and the agent app bundle inside it, but the progress agent / relaunched tool destroys it on cleanup. Since it is "transferred", the agent only removes the temp directory (after removing the agent app in it) if it's empty, which rmdir() does. Just a safety precaution. I could add a non-recursive dir removal method to SUFileManager but I don't consider that too important.

Sparkle/SPUCoreBasedUpdateDriver.m
zorgiepoo: The installer process moves the update from the download directory to its own cache directory before starting to extract it. So the framework no longer "owns" the download and is transferred. (This also prevents the framework from still accessing the download; think eg sandbox -> non-sandbox installer, user -> root installer) The download directory from the framework is safe to remove at this point.

Downloader/SPUDownloader.m
zorgiepoo: Yeah.. there's two reasons: 1. I don't want to send full path across XPC so I let this class compute most that it can. 2. The downloader instance and update session that creates the temp directory is not necessarily the same as the one used to clear the download later, so can't "save" it in the downloader or XPC Service.

Alternatively I could send/pass full path and have downloader validate sub-components but seems less safe.

kornelski (approved):

PR #1838, opened 2021-04-25
Fix delta updates edge cases

zorgiepoo:
This fixes two issues / edge cases with delta updates:

1. `-bestValidUpdateInAppcast:forUpdater:` did not ever properly support delta updates. All the top-level items of the appcast are non-delta items, so `-copyWithoutDeltaUpdates:` didn't make sense. Instead of asking delegate of delta updates, we just ask them for regular updates, and we retrieve the delta update ourself.
2. When resuming an update for later (instead of downloading/installing it immediately, eg for update that requires auth), we saved the appcast item, but did not save the secondary (non-delta) appcast item. So if the primary delta appcast item failed to extract, we couldn't fallback to regular update in that case. We now pass around and preserve the secondary update item too.

## Checklist:

## Type of change

## Testing

- [x] Sparkle Test App
- [ ] Unit Tests
- [x] My own app
- [ ] Other (please specify)

Tested success & failed delta update for regular installation and resumed installation (requiring auth). Tested test app works.

### DIFF
--- a/Sparkle/SPUAutomaticUpdateDriver.m
+++ b/Sparkle/SPUAutomaticUpdateDriver.m
@@ -77,15 +77,15 @@
     SULog(SULogLevelError, @"Error: resumeDownloadedUpdate:completion: called on SPUAutomaticUpdateDriver");
 }
 
-- (void)basicDriverDidFindUpdateWithAppcastItem:(SUAppcastItem *)updateItem preventsAutoupdate:(BOOL)preventsAutoupdate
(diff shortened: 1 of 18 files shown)


kornelski (approved):

kornelski:
Passing it explicitly is definitely better than smuggling through another property.

However, two args seem repetitive. Maybe change SUAppcastItem arg to SUSelectedAppcastItems object with primary and secondary fields?

zorgiepoo:
Agree but lots of refactoring and requires re-testing the changes. Will merge this for now and try to see if I can refactor it in another change.

PR #1853, opened 2021-05-10
Skip minor updates separately from major upgrades

### LINKED ISSUE #329: Set minimum application version for update
I have an application that originally had an update folder location for minor version 1 releases called "update". When I moved to version 2, I now have a folder location for minor releases of version 2 called "update2". My problem lies in the fact I accidentally released a version 2 minor version that pointed to the original "update" folder, not the "update2" folder. Is there any way to add something to the Appcast file that specifies version number of the application so I can place it in the original "update" folder allowing version 2 users to update to the newest version, but not allow version 1 users to receive this new update?

zorgiepoo:
Handle skipping minor updates separately from major upgrades (in the case of paid upgrades).

If the user skips a major upgrade, they skip future updates belonging to that upgrade (similar to if a developer decides to use a different appcast feed). But they also now do not skip updates to their current / older major versions, allowing developers to publish a fix to the last major version without it being automatically skipped.

Fixes #329 (along with #1850)

User driver update found API now encapsulates information in a state object (including its stage, if it is a user initiated, or if update is major one). I removed some legacy deprecated user drivers APIs that were becoming painful to support / will now result in compiler errors if not migrating to new APIs.

We also re-title install/skip buttons in the case of a major upgrade (i.e, "Skip Major Upgrade" instead of "Skip this Version", "Install Upgrade" instead of "Install Update")

## Checklist:

## Type of change

- [x] My change is or requires a documentation or localization update - Will update the paid section on our doc websites at a later point.

## Testing

- [x] Sparkle Test App
- [x] Unit Tests
- [ ] My own app
- [ ] Other (please specify)

Added skipped update tests with minimumAutoUpdate to test correct behavior.
Tested altering sparkle test app appcast and inspecting user defaults when skipping one or multiple updates, and making sure updates are correctly skipped.

I tested all of these:
- User is on 1.0. Updates are 1.5, and 2.0 (min autoupdate version to 1.5). User is presented 1.5, if you skip it, user will then automatically be presented 2.0.

- User is on 1.0. Updates are 2.0 (min autoupdate version to 1.5). User is presented 2.0, if you skip it, user will not automatically be shown new updates. If a new item shows up in the feed as 1.4 however, 1.4 will be automatically presented to user and 2.0 will stay skipped.

- User is on 1.0. Updates are 2.0 (min autoupdate version to 1.5). User is presented 2.0, if you skip it, user will not automatically be shown new updates. If a new item shows up in the feed as 2.1 (same min autoupdate version to 1.5), 2.1 will also be skipped automatically.

- User is on 1.0. Updates are 1.5. User is presented 1.5, if you skip it, user will not automatically be shown new updates. If a new item shows up in the feed as 1.6, 1.6 will be automatically presented to user.

- User is on 1.0. Updates are 1.4 and 1.5. User is presented 1.5, if you skip it, user will not automatically be shown new updates.

- User initiating for updates themselves clears out all skipped updates.

- Skipping a new update replaces currently stored skipped update for that same train / minimumAutoupdateVersion.

macOS version tested: 11.3 (20E232)


kornelski:
I'm not sure about exposing this in the UI. It adds new strings to translate, and I don't think users will notice subtle differences between Update and Upgrade.

zorgiepoo:
> I'm not sure about exposing this in the UI. It adds new strings to translate, and I don't think users will notice subtle differences between Update and Upgrade.

I think regardless of any skipping logic, having more clarity that an update is a major upgrade outside of the release notes is a decent addition. I do think you are right the subtle strings I added aren't that great way of doing this. Once there is the right way to do this (I'm not that interested in solving this right now), then translations can follow along.

zorgiepoo:
I've been thinking about this in the background a bit about other possibilities, but ultimately I still believe:

* Users should be able to download and install a major version from within the application using Sparkle
* Users should be able to opt-out of minor update notifications to a major upgrade, without disabling the updater system

Major update: an update significant in changes that needs more notice from user; often paid, but otherwise may present compatibility changes, significant overhauls, or workflow changes.

Disabling auto-updates is in progression to this. It's not a shop. It's presenting the user a choice of installing an update and deciding/respecting if they want to install it. Other application-specific details need to be handled separately, much like in the case if the user installs the update manually.

zorgiepoo:
I made changes to greatly simplify the implementation here by tracking a skipped major version and a skipped minor version, so up to two versions can be skipped, instead of us tracking an array of version pairs.

Instead of subtly renaming install/skip buttons, I added a confirmation alert for users skipping future updates to an upgrade.

This fills in the gap of what developers are currently doing for major upgrades and doesn't leave us in a currently half-supported state for them. There's also user driver API changes encapsulating the state and cleanup here I need in the future.

kornelski (approved):

Issue #1855, opened 2021-05-11
Critical and informational updates are not selectively chosen


zorgiepoo:
Critical and informational updates are not selectively chosen and are currently faulty in not persisting in future updates.

If a developer publishes a critical or informational update (say v1.5.1), and they post another update (say v1.5.2), chances are they want 1.5.2 to not be critical or informational if coming from v1.5.1, but may want them to be critical or informational if coming from say v1.5.0.

Relying on a certain amount of time passing until the next update is not a super good way of addressing this, and what some developers must be doing is spinning off new appcast feeds to work around this. I'm not sure if we can do better here. Selectively deciding based on version could be tricky, and these are a little bit edge-casey, at least less important than handling paid or beta updates imo.

PR #1862, opened 2021-05-31
Select critical/informational updates by version & store state in appcast items

zorgiepoo:
State such as if an appcast item passes min/max OS version, minimum autoupdate version, (and now) critical update are now stored as part of SUAppcastItem and are now resolved/computed at initialization time. These properties rely on external factors such as the application's current version or the OS version. This ends up simplifying ares in the codebase, not having to pass around additional flags, or having redundant properties. This also avoids re-computing these properties in some cases.

We also now add eg `<sparkle:criticalUpdate sparkle:version="2.0" />` so the (optional) version that was critical can be specified (this way versions >= 2.0 don't treat this update as critical, and you can leave this in for future for a desired amount of time). Eventually we may support something similar for informational updates. I deprecated the legacy path of putting this in a tag, eg `<sparkle:tags><sparkle:criticalUpdate  /></sparkle:tags>` in favor of making it root-level and adding a version attribute; this was the only tag we had built-in.

I also fixed sparkle-cli since I forgot to update its user driver usage there (since I've been updating the API around).

Related: #1855

## Checklist:

## Type of change

- [x] My change is or requires a documentation or localization update -- eventually I'll get to the website

## Testing

- [x] Sparkle Test App
- [x] Unit Tests
- [ ] My own app
- [x] Other - sparkle-cli

Added and adjusted unit tests. Used test app to test minimum autoupdate version / major upgrade, min/max OS version, critical updates (by varying the specified version constraint), informational updates (by varying the set of specified versions). Tested downloading/deferring update via sparkle-cli too.

macOS version tested: 11.4 (20F71)


zorgiepoo:
Need to test / re-evaluate this again after merge conflicts / supporting legacy appcast item initializers.

zorgiepoo:
Ready to review. @kornelski 

Summary:
* Add new & refine existing properties to SUAppcastItem: majorUpgrade, minimumOperatingSystemVersionIsOK, maximumOperatingSystemVersionIsOK, criticalUpdate, informationOnlyUpdate.
** These are evaluated on initialization
** These are cached instead of re-computed multiple times
** These reduced number of parameters we need to pass in several cases, and reduces number of objects/states we need to expose
* New designated initializer of SUAppcastItem (private but outside accessible)
* Critical updates are optionally selective; you can specify the last update version that was marked critical.
** Makes sense to only apply critical update path if host is below the version that contains a critical fix. Don't need to spin off new appcast or post critical updates unnecessarily in some cases.
** Took opportunity to migrate away from sparkle:tags to be consistent with other features
* Informational updates are optionally selective; you can specify set of host versions that should treat update as info-only.
** Makes sense for one-off "hosed" updates
** Makes sense for major upgrades
** Don't need to spin off new appcast in some cases.

kornelski (approved):

zorgiepoo:
cc @tonyarnold

PR #1867, opened 2021-06-09
Fix the build with Xcode 13 by explcitly importing ImageIO

danielpunkass:
Fix the build with Xcode 13 by explicitly importing ImageIO. Previously it was being implicitly imported along with Foundation.

## Checklist:

## Type of change

## Testing

- [x] Other: It's a build-only fix that doesn't impact the functionality of the framework.

### DIFF
--- a/InstallerLauncher/SUInstallerLauncher.m
+++ b/InstallerLauncher/SUInstallerLauncher.m
@@ -14,8 +14,8 @@
 #import "SPULocalCacheDirectory.h"
 #import "SPUInstallationType.h"
 #import "SUHost.h"
+#import <ImageIO/ImageIO.h>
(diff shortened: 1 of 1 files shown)


zorgiepoo (approved):

zorgiepoo:
Thanks, we'll probably bump our CI builder to Big Sur / Xcode 13 beta (and still 12) soon.

danielpunkass:
There are some new warnings too but I didn't think they were critical enough to submit a PR for, and it might take a little more thinking about how to address them (all due to the new "closure never gets called" warning).

zorgiepoo:
Those new warnings are for supporting Swift async/await. They're probably mostly false positives but I'll take a look at them.

PR #1878, opened 2021-06-27
Favor sparkle:version and sparkle:shortVersionString elements over enclosure attributes

### LINKED ISSUE #1207: sparkle unexpectedly requires "sparkle:shortVersionString" in delta update feed items
1.) download https://github.com/aonez/Keka/releases/download/v1.0.14/Keka-1.0.14.zip
2.) "check for updates"
3.) you'll see "Keka 1804 is now available - you have 1.0.14." this obviously confusing message (which compares a bundleversion with a shortversionstring) is presented because sparkle seems to ignore the sparkle:shortVersionString information present in the update feed here: http://www.kekaosx.com/update/keka.xml
4.) since the feed is valid i don't know why sparkle ignores the shortversionstring in this feed, but uses it just fine for most other feeds. same bug exists when replacing the sparkle framework in Keka with the newest version 'Sparkle-1.19.0rc2'

zorgiepoo:
We should recommend using sparkle:version and sparkle:shortVersionString elements inside the appcast item instead of their corresponding attributes inside the enclosure for consistency.

Delta updates do not need to specify sparkle:version or sparkle:shortVersionString, and can inherit the top level elements. This clears up some confusion and reduces duplication (and reduces error proneness / saves work when switching between specific features).

This is also consistent with informational updates. Regardless whether you want to make an update informational only or not (or omit the enclosure), the sparkle:version / sparkle:shortVersionString stays in the same top-level place.

This is a backwards compatible change. Very old versions of Sparkle (possibly up to a decade or more ago) have supported specifying sparkle:version / sparkle:shortVersionString as top level items. Winsparkle supports it as well because it's needed for info-only updates.

This is only changing what we recommend / standardize. Developers can still use the attribute variants in the enclosure if they want to.

External tools relying on Sparkle's feed format may have to adapt to this standardization (they have always needed to support info-only updates with a missing enclosure which apps do leverage, eg for major upgrades, anyway).

I updated generate_appcast to prefer the element variants and updated the tool so it understands them (this was a bug). I looked at the code that was looking for minimumSystemVersion and copied that.

Fixes #1207 as well.

## Checklist:

## Type of change

- [x] My change is or requires a documentation or localization update - requires website docs update which I can get to later (as well with everything else..)

## Testing

- [x] Sparkle Test App
- [x] Unit Tests
- [ ] My own app
- [x] Other - generate_appcast

Added couple unit tests to make sure versions (element and attribute variants) are parseable.
Tested generate_appcast making sure it writes new entries with element variant, and that it can look up both attribute & element version variants from existing entries.

### DIFF
--- a/Resources/SampleAppcast.xml
+++ b/Resources/SampleAppcast.xml
@@ -6,11 +6,11 @@
         <language>en</language>
         <item>
             <title>Version 2.0 (2 bugs fixed; 3 new features)</title>
+            <link>http://sparkle-project.org</link>
             <sparkle:version>2.0</sparkle:version>
             <sparkle:releaseNotesLink>
                 http://you.com/app/2.0.html
             </sparkle:releaseNotesLink>
-            <link>http://sparkle-project.org</link>
             <pubDate>Wed, 09 Jan 2006 19:20:11 +0000</pubDate>
             <enclosure url="http://you.com/app/Your%20Great%20App%202.0.zip" length="1623481" type="application/octet-stream" sparkle:edSignature="WVyVJpOx+a5+vNWJVY79TRjFKveNk+VhGJf2iti4CZtJsJewIUGvh/1AKKEAFbH1qUwx+vro1ECuzOsMmumoBA==" />
             <sparkle:minimumSystemVersion>10.11</sparkle:minimumSystemVersion>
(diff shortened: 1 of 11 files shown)


zorgiepoo:
cc @vslavik

kornelski (approved):

vslavik:
To be clear, "favor" means in documentation here, the code still gives enclosure attributes higher priority as far as I can tell — i.e. item version is only used of the enclosure doesn't say otherwise.

That seems reasonable to me and is how WinSparkle behaves too (although mostly be accident, I think, and the support you refer to probably originates from those decade-old versions of Sparkle...

zorgiepoo:
Yep, that's exactly it.

PR #1879, opened 2021-06-28
Add support for channels

### LINKED ISSUE #51: Add option to have multiple updates in a single appcast that can be selectively chosen
_**livings124 (livings124)_\* [reported (on Launchpad)](https://bugs.launchpad.net/sparkle/+bug/316247) on' 2009-01-12:*

I'm not sure if "Answers" or here is the best place to add feature
requests, so I'm guessing it's here.

It would be useful to have multiple updates listed in a single appcast
file, with some sort of key separating them. This would be similar to
how it works now with minimum version numbers.

For an example, I could have 2 updates, one with no key and one with the
key "beta". There would be some sort of method (delegate maybe?) to
specify all the valid keys (as an array most likely). If no keys are
specified, the update with the "beta" key would be ignored and only the
other would be considered. If one of the keys specified in the code is
"beta", however, both would be considered, and the one with the higher
version number would be used. This would simplify things greatly for
users that want to support optional updates for betas, etc.; it's easier
to maintain and feels less hacky than having multiple feeds that have to
be synchronized.

zorgiepoo:
Add support for channels, mainly intended for supporting beta or nightly updates in the same appcast feed.

You specify a channel on an appcast item, and the updater will only look for updates on that channel if the delegate says it's allowed to. No specified channel means the appcast item is on the default channel, which updaters always have access to.

This is only intended to support versions that branch off to different channels and that eventually converge back to the default channel. It's not for maintaining completely parallel trains; separate appcast feeds still fit that use case (or the use case of migrating to newer appcast features, like this one).

This implementation is simple and almost identical to the [patch](https://launchpadlibrarian.net/21146422/tags.diff) proposed in #51 for Transmission back in the day.

I needed to also fix the logic that assumed no updates in an appcast meant there was something wrong with the feed. This may happen for various reasons, but in this case we must assume the user is just up to date.

Probably my last enhancement / missing gap to 2.x, then I intend to look into tooling (sparkle-cli, generate_appcast) and docs..

Fixes #51 

## Checklist:

## Type of change

## Testing

- [x] Sparkle Test App
- [x] Unit Tests
- [ ] My own app
- [x] Other - sparkle-cli

Tested adding beta channel to test app and made sure updates were picked (or not picked) appropriately based on the implemented delegate method.
Added unit tests.
Tested sparkle-cli support for channels on running test app.


kornelski (approved):

jeff-h:
I'm wondering if there's any way to only show beta release notes to those using my beta channel  (I'm using external release notes)?

zorgiepoo:
Since the appcast item will be only be available to those who subscribe to the beta channel can you just not use a different beta release notes file for them? People on the default channel shouldn’t see them.

jeff-h:
Sorry, I was tricked by the fact my generated appcast.xml only had one `sparkle:releaseNotesLink`, so I didn't realise it would work in other items also. Thanks for your time helping a newbie :)

PR #1880, opened 2021-07-04
Allow delegate to pick no valid items

### LINKED ISSUE #1399: Testing for Hardware configuration?
I currently use the Sparkle.Framework and would like to know how to best handle a hardware condition test. I can test for a minimum version in my app cast, but I need to now check for hardware that supports Metal. The following delegate method makes sense:

- (nullable SUAppcastItem *)bestValidUpdateInAppcast:(SUAppcast *)appcast forUpdater:(SUUpdater *)updater;

But, looking at the code (SUBasicUpdateDriver.m), if I return nil, then Sparkle will choose the best item. How have others handled this?

zorgiepoo:
We now allow returning NSNull.null in -bestValidUpdateInAppcast:forUpdater: to indicate no update should be picked.

We preserve returning nil to indicate Sparkle handles the best appcast selection instead.

People have expressed needs for both of these decisions (and deprecated SUUpdater relies on the nil behavior).

Fixes #1399 (allowing no fall back)

## Checklist:

## Type of change

## Testing

- [x] Sparkle Test App
- [ ] Unit Tests
- [ ] My own app
- [ ] Other (please specify)

Tested implementing delegate method in test app and that returning appcast item, returning nil, returning NSNull.null work expectably.

### DIFF
--- a/Sparkle/SPUUpdaterDelegate.h
+++ b/Sparkle/SPUUpdaterDelegate.h
@@ -209,9 +209,10 @@ typedef NS_ENUM(NSInteger, SPUUpdateCheck)
  
  \param appcast The appcast that was downloaded from the remote server.
  \param updater The updater instance.
- \return The best valid appcast item, or nil if you don't want to be delegated this task.
+ \return The best valid appcast item. Return NSNull.null if no appcast item is valid. Return nil if you don't want to be delegated this task and
+         want to let Sparkle handle picking the best valid update.
  */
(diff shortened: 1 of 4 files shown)


kornelski:
Maybe to preserve the return type there could be a global symbol for `NoUpdateSelected` that is a dummy `SUAppcastItem` and is checked with pointer identity?

kornelski (approved):

zorgiepoo:
Yeah good idea. I think something like `SUAppcastItem.emptyAppcastItem` may work.

kornelski (approved):

PR #1884, opened 2021-07-06
Set $USER and $HOME variables when running guided pkg installers

### LINKED ISSUE #1882: Sparkle 2 runs in root space environment
Im having an issue moving from sparkle 1.23 to Sparkle 2.0.
The environment is different and its a problem for us because we are using python and pip installation in post/pre scripts and we get a different locations.
For example the last line in the install log I attached is detecting the user_site:
(from the preinstall script: python -c "import site;print site.USER_SITE")
USER_SITE=/var/root/Library/Python/2.7/lib/python/site-packages 
While we used to get with Sparkle1:
USER_SITE=/Users/bongi/Library/Python/2.7/lib/python/site-packages

Every line in the log that has /var/root/.. used to be /Users/myUserName/..
I also noticed that the extracted location was ‘org.sparkle-project.Sparkle/PersistentDownloads/’ and with sparkle2 its ‘org.sparkle-project.Sparkle/Installation/’

We are using pkg. I define the xml:
sparkle:installationType="package"

Can I somehow work in the previous user space mode?
**[sparkle2_install_log.txt](https://github.com/sparkle-project/Sparkle/files/6760237/sparkle2_install_log.txt)**

zorgiepoo:
This allows pre/post install scripts to reference the user's environment correctly. The standard pkg installer GUI preserves these two variables.

Fixes #1882

## Checklist:

## Type of change

## Testing

- [ ] Sparkle Test App
- [ ] Unit Tests
- [ ] My own app
- [x] Other - custom app that updates via pkg

I tested that a post install script correctly uses appropriate $USER and $HOME (by echo'ing them to a file in /tmp). My pkg file didn't correctly install the app but that's probably because I didn't create the pkg correctly as the same happens when I run with `sudo installer` .. and they're a pain to figure out generating correctly.

### DIFF
--- a/Autoupdate/AppInstaller.h
+++ b/Autoupdate/AppInstaller.h
@@ -13,7 +13,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface AppInstaller : NSObject
 
-- (instancetype)initWithHostBundleIdentifier:(NSString *)hostBundleIdentifier;
+- (instancetype)initWithHostBundleIdentifier:(NSString *)hostBundleIdentifier homeDirectory:(NSString *)homeDirectory userName:(NSString *)userName;
 
 - (void)start;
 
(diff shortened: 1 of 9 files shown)


kornelski (approved):

kornelski:
That was easier to fix than I suspected

Issue #1355, opened 2019-02-03
Migrating from DSA to EdDSA


gnachman:
I'm trying to stop using DSA keys in favor of EdDSA. I'd like to minimize the amount of time that DSA keys are trusted. I have plenty of users in the field with old builds (using Sparkle 1.7.0).

It seems I cannot change the DSA key. If I change it, updates fail saying "DSA signature validation failed. The update has a public DSA key and is signed with a DSA key, but the new public key shipped with the update doesn't match the signature. The update will be rejected."

I cannot remove the DSA key. Doing so causes the update to fail with "A public DSA key is found in the old bundle but no public DSA key is found in the new update. For security reasons, the update will be rejected."

From this I conclude the migration path is to support both keys for time (perhaps as little as one version). I will need to keep the one of these releases in the appcast indefinitely.

Then I need to provide a way for users running this release with both keys to upgrade to the EdDSA-only version without letting users on the DSA-only releases know about it (lest they try to upgrade and get one of the aformentioned errors). I guess this means I need to use a different URL for appcasts for my Sparkle 1.21 users.

Is this correct?

kornelski:
I haven't put out much thought into migration path off it yet, so Sparkle is cautious and conservative about the new keys.

I think the best solution would be to change Sparkle itself to start ignoring old DSA completely.

gnachman:
I've tested using the plan I suggested and it works, but it is a pain. If you relax the requirement that an app with a DSA public key requires a DSA signature then it could get you down to two appcasts. I don't know the reasoning behind that decision, though. Is that to protect developers from their own mistakes or is there a specific vulnerability you were worried about?

kornelski:
There's no vulnerability per se. I've just been (too) careful not to create a logic bug.

For example, adding EdDSA signature and EdDSA key to the *new* app shouldn't be enough to pass EdDSA signature verification, because both signature and the key are new and therefore untrusted in this scenario. The app needs to check new signature against the old key to ensure the chain of trust is unbroken.

Technically there's no problem at all doing that. It's just that adding EdDSA already required many changes in the signature verification code, and supporting all of the combinations separately, securely and with helpful error handling, adds a ton of new code paths: {no DSA, old DSA, old EdDSA, old DSA + EdDSA} * {new DSA, new EdDSA, new DSA + EdDSA} * {app code signing, unsigned bundle, package}.

zorgiepoo:
[...]

PR #1888, opened 2021-07-10
Avoid DSA verification if EdDSA verification passes

zorgiepoo:
If EdDSA verification passes we don't need to have DSA verification pass, and don't need to require the new update to contain a DSA signature or keys.

We should also require EdDSA signature to be present in appcast if the app has EdDSA key. We ignore DSA unless the signature in the appcast is erroneously invalid.

The developer can omit the DSA signature in the appcast and if EdDSA verification passes, all is good.
The worst I see that can happen is that an unintentional/malicious change to appcast removes DSA signature on an update item, but that would just reject old app updates that don't have EdDSA keys.

Refs #1355

## Checklist:

- [x] My change is being backported to master branch (Sparkle 1.x). Please create a separate pull request for 1.x, should it be backported. Note 1.x is feature frozen and is only taking bug fixes, localization updates, and critical OS adoption enhancements. - *I probably will after this is approved*

## Type of change

I will get to documentation/migration as part of another change. I also want to emit a log indicating DSA-only is deprecated soon.

## Testing

- [x] Sparkle Test App
- [x] Unit Tests
- [ ] My own app
- [x] Other - custom test app to test DSA + EdDSA -> EdDSA

Tested Sparkle Test App (EdDSA -> EdDSA)
Tested custom app dropping DSA for app path (DSA + EdDSA -> EdDSA) (New allowed path)
Tested custom app dropping DSA for pkg path (DSA + EdDSA -> EdDSA) (New allowed path)
Tested sparkle-cli against an app that only used DSA (DSA -> DSA)
Tested sparkle-cli against an app that migrated to EdDSA (DSA -> DSA + EdDSA)
Updated unit tests testing input combinations of signatures and public keys to see what updates are valid

### DIFF
--- a/Tests/SUUpdateValidatorTest.swift
+++ b/Tests/SUUpdateValidatorTest.swift
@@ -62,14 +62,16 @@ class SUUpdateValidatorTest: XCTestCase {
         let dsaSig: String?
         switch config.dsa {
         case .none: dsaSig = nil
-        case .invalid: dsaSig = "MCwCFCIHCiYYkfZavNzTitTW5tlRp/k5AhQ40poFytqcVhIYdCxQznaXeJPJDQ=="
+        // Use some invalid base64 characters
+        case .invalid: dsaSig = "^%wCFCIHCIYYkfZavNzTitTW5tlRp/k5AhQ40poFytqcVhIYdCxQznaXeJPJDQ=="
         case .valid: dsaSig = "MCwCFCIHCIYYkfZavNzTitTW5tlRp/k5AhQ40poFytqcVhIYdCxQznaXeJPJDQ=="
         }
 
         let edSig: String?
         switch config.ed {
         case .none: edSig = nil
-        case .invalid: edSig = "wTcpXCgWoa4NrJpsfzS61FXJIbv963//12U2ef9xstzVOLPHYK2N4/ojgpDV5N1/NGG1uWMBgK+kEWp0Z5zMDQ=="
+        // Use some invalid base64 characters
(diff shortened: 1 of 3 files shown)


zorgiepoo:
Going to think over this a bit more, seeing some odd logic in the unit tests. Also kudos to @belkadan for the extensive unit tests for verification here.

Tests/SUUpdateValidatorTest.swift
```diff
@@ -62,14 +62,16 @@ class SUUpdateValidatorTest: XCTestCase {
         let dsaSig: String?
         switch config.dsa {
         case .none: dsaSig = nil
-        case .invalid: dsaSig = "MCwCFCIHCiYYkfZavNzTitTW5tlRp/k5AhQ40poFytqcVhIYdCxQznaXeJPJDQ=="
+        // Use some invalid base64 characters
+        case .invalid: dsaSig = "^%wCFCIHCIYYkfZavNzTitTW5tlRp/k5AhQ40poFytqcVhIYdCxQznaXeJPJDQ=="
```
zorgiepoo, line 66: This wasn't actually generating an invalid signature status before.

zorgiepoo: Never mind. Invalid here meant still formatted in valid way, but bad in verification. I added another configuration case for invalidFormat.

zorgiepoo:
Re-evaluated this more and I think changes make sense now.

kornelski:
If it's for 2.x, can we just drop DSA entirely?

kornelski (approved):

zorgiepoo:
I think it's too early right now. It will make migrating more difficult, 2.x added EdDSA support a lot later than 1.x. I still personally have one app using 2.x that I still need to migrate away from DSA. Until now, there wasn't a good migration plan to drop DSA. I want to also emit a log for apps only using DSA, and note it will be removed in a future release, first.

PR #1891, opened 2021-07-12
Add/port update relaunched API

zorgiepoo:
Add/port update relaunched API from #1115.

This inserts an environment variable on relaunch rather than mutating the host bundle defaults and favors exposing a class property in 2.x over exposing a delegate call in 1.x

I decided to port the ability to detect this as there may be legitimate use cases, but with a doc note that it should not be used in cases where manual updates must be handled.

Removed some dated __has_feature(objc_generics) checks I noticed too.

## Checklist:

## Type of change

## Testing

- [x] Sparkle Test App
- [x] Unit Tests (UI tests in this case)
- [ ] My own app
- [ ] Other (please specify)

Adjust UI test to take this API in account, tested it (and SUUpdater stub) with the test app.

### DIFF
--- a/Sparkle/InstallerProgress/InstallerProgressAppController.m
+++ b/Sparkle/InstallerProgress/InstallerProgressAppController.m
@@ -331,8 +331,19 @@ static const NSTimeInterval SUTerminationTimeDelay = 0.3;
             }
             
             // Note: we can launch application bundles or open plug-in bundles
-            if (![[NSWorkspace sharedWorkspace] openFile:pathToRelaunch]) {
-                SULog(SULogLevelError, @"Error: Failed to relaunch bundle at %@", pathToRelaunch);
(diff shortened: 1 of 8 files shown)


kornelski:
mainBundleRelaunchedFromUpdate is a footgun. Could it be kept private?

Or alternatively, make it always work even if relaunch wasn't from Sparkle by storing last known application version in nsuserdefaults and notifying when it changes.

zorgiepoo:
> Or alternatively, make it always work even if relaunch wasn't from Sparkle by storing last known application version in nsuserdefaults and notifying when it changes.

At this point, I don't think this is Sparkle's duty. On second thought though, this sounds like a right approach for an app to handle themselves. I was considering adopting this because 1.x had support for it but I think I'm going to end up declining this.

PR #1892, opened 2021-07-15
Backport security and DSA changes to 1.x

zorgiepoo:
Backport the recent DSA deprecation changes and allowing dropping of DSA signatures in new updates more easily.

Backport logging that not using any Sparkle signing keys is deprecated (more on this in a bit).

Tested:
DSA -> DSA
EdDSA -> EdDSA
DSA + EdDSA -> DSA + EdDSA
DSA + EdDSA -> EdDSA

### DIFF
--- a/Sparkle/SUSignatureVerifier.m
+++ b/Sparkle/SUSignatureVerifier.m
@@ -103,7 +103,8 @@
     const unsigned char *edPubKey = self.pubKeys.ed25519PubKey;
 
     if (edPubKey && !edSignature) {
-        SULog(SULogLevelDefault, @"There is no EdDSA signature in the update, but the app is capable of verifying them");
+        SULog(SULogLevelError, @"The app has an EdDSA public key, but there is no EdDSA signature in the update, so the update will be rejected.");
+        return NO;
     } if (!edPubKey && edSignature) {
         SULog(SULogLevelDefault, @"The update has an EdDSA signature, but it won't be used, because the old app doesn't have an EdDSA public key");
     } else if (edPubKey && edSignature) {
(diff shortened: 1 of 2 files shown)


zorgiepoo:
I think this would be an appropriate place to re-discuss signing policy both for 1.x / 2.x (#543). (I think 1.x and 2.x should converge towards same behavior here).

On one hand, if a developer's web server is hacked, Sparkle's support for Apple's code signing will prevent some types of changes to the app bundle. So it's definitely better than nothing but it will not reject:
* Changes to files in nested bundles (due to compatibility/efficiency reasons historically, Sparkle does only a shallow bundle signing check. This is likely the weakest link.)
* Replacements (and downgrades) of nested code dependencies (like frameworks) signed from the same author.
* Missing localization files or stripped architectures (like stripping out x86_64)
* Incorrect file metadata (permissions, missing executable bit, ACLs, etc)

It would be easy to 'break' the update, a bit harder to insert something malicious.

On the other hand, a lot of apps out there don't use DSA/EdDSA and don't want to use Sparkle's signing mechanism. Maybe some middle ground is instead of deprecating not using EdDSA, we change the documentation to recommend EdDSA for more secure updates, plus having the safety net if a developer needs to change their Dev ID certificate.

Another possibility is we enforce deep signing checks (kSecCSCheckNestedCode) if no EdDSA keys are used. The risk here is that if another major code signing check is introduced (like there was in macOS 10.9), this may be more likely to break. The other risk is a larger performance penalty for big apps.

kornelski (approved):

zorgiepoo:
I think the biggest problem is the shallow signing check provides a weaker guarantees than code signing is intended to provide in this case, and changing that has other potential consequences.

Going to merge as-is for now.

Issue #1883, opened 2021-07-05
Shipping pre-release builds of Sparkle 2.x via Swift Package Manager


tonyarnold:
It would be great to do away with some of the tooling that I'm currently maintaining for my projects by making pre-release builds of Sparkle 2.x available via Swift Package Manager (SPM). 

- I understand that the project is currently shipping Sparkle 1.x builds via SPM.
- I also understand that Sparkle 2.0 is _not_ ready to ship yet, and there are ongoing discussions about the API surface that affect any potential ETA. 

Still, it would be great if we could automate deployment of 2.x binary builds for SPM, and commit to tagging some early version of the branch as a pre-release. 

As always, I'm happy to help where I can and commit my time/efforts to this issue.

zorgiepoo:
Yeah we should figure out how to make prereleases for SPM and I think it's a good time to start doing that. Just haven't looked into it yet so feel free making PRs needed to do this.

cc @thebluepotato

tonyarnold:
Actually, is it even possible to ship the XPC Services via SPM? (Even as binaries)

zorgiepoo:
It might be possible, see https://github.com/sparkle-project/Sparkle/pull/1635#issuecomment-690288772

tonyarnold:
So it sounds like most of the hard work is done - I'd need to automate the release to GitHub via Actions to solve the immediate issue?

zorgiepoo:
I think... all we need to do is make a new tag `2.0.0-rc1`, run `make release`, push Package.swift / tag changes, and potentially add a pre-release draft / changelog on GitHub.

tonyarnold:
Excellent! So how's about an early alpha release of 2.0.0 to test that it's all working? I understand the API is still in flux for v2, but it'd be great to pick a point and start doing pre-releases.

zorgiepoo:
There are some recent DSA (potentially Apple Code Signing only too) deprecation / migration changes merged in 2.x that may be important enough to back port to 1.x and even make a small release there first (without having to think about if ordering will matter when tagging releases..)

There's also some documentation changes I'm working on, and some migration changes I should note for people migrating from earlier 2.x versions when creating a prerelease on GitHub.

Currently I don't have more API plans other than extending / taking a look at generate_appcast, but you never know..

tonyarnold:
> without having to think about if ordering will matter when tagging releases

[...]

PR #1751, opened 2021-02-04
Fix Package.swift

thebluepotato:
I'll do the change for the 2.x branch ASAP, but this fixes the showstopping bug in #1750. Major change is that Package.swift now uses the latest **tag** instead of the version number so it can be used for pre-release versions as well. It still includes the "final" version number even though it is not used in any way anymore.
**IMPORTANT**: it seems that during the release of `1.25.0-rc1` two things were overlooked:
- SPM **needs** the tag to follow semantic versioning to be able to resolve the git repo. `1.25.0rc1` does **not** follow semantic versioning whereas `1.25.0-rc1` does. I've updated make-release-package.sh to check whether the latest tag follows semantic versioning based on a modified version of the "official" RegEx: https://semver.org/. It outputs a warning if the tag does not follow semantic versioning. In that case, you should delete the tag, tag again and run `make release` again.
- The "force tag" stage at the end of `make release` defaults to **no** and seems to have been overlooked. As there are other safeguards in place (only moves the tag when nothing else has been committed), it now defaults to **yes**.

Fixes #1750 

## Checklist:

- [ ] My change is being tested and reviewed against the Sparkle 2.x branch. New changes must be developed on the 2.x development branch first.
- [x] My change is being backported to master branch (Sparkle 1.x), if deemed necessary. Please create a separate pull request for 1.x, should it be backported.
- [x] I have reviewed and commented my code, particularly in hard-to-understand areas.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] My change is or requires a documentation or localization update

## Testing

I tested and verified my change by using one or multiple of these methods:

- [ ] Sparkle Test App
- [ ] Unit Tests
- [ ] My own app
- [x] Other: Ran `make release` with tag `1.25.0-rc2`, successfully moves the tag and uses the correct url in Package.swift

macOS version tested: 11.2

PR #1908, opened 2021-07-26
Add a workflow that builds and publishes a release via GitHub Actions

tonyarnold:
This PR proposes to add a new GitHub Actions workflow that will build and publish a copy of the Sparkle XCFramework to the releases page on GitHub. 

The workflow can be triggered by creating a new release via the "Releases" page on GitHub.

I've no way of testing this properly, so someone from the core team will probably need to take what I've done here, try it and fix any issues that arise when they are doing the first Sparkle 2.x prerelease. 

I'm also not sure if there's more that we need to publish here - I can see that there are separate frameworks published for SPM in v1.x. Please feel free to comment or review and I'll add anything I've missed here.

Addresses #1883.


zorgiepoo:
Thanks for the PR attempt. I didn't have a chance to run it but noticed a few things.

First off, to clarify my aversion for using GitHub's artifact zip tool and using ditto instead, see eg https://github.com/actions/upload-artifact/issues/37 which apparently you have already noticed before :)

However if you use `actions/upload-artifact` on a `.zip` file, it's likely going to `.zip` the zipped file again (https://github.com/actions/upload-artifact/issues/39) when uploading it which is problematic..

The Distribution target invokes to `make-release-package` which already builds a SPM zip package, so we don't need to replicate that in a GitHub action. It currently uses `zip -rqyX -9`, we ought to change it to use `ditto`. It also includes more things we need to distribute other than just the xcframework (CLI tools, XPC Services, ..)

We haven't made the decision to drop the non-xcframework variant so we may still need to provide it.. although, it might make sense to, because there's some redundancy here in packaging both versions. Unfortunately though the xcframework variant for SPM needs to be a `.zip` and it does not presently compress nearly as nicely compared to the tar.xz version (> 4x size difference).

The last hurdle (from #1751).. is that the when `make-release-package` is not run in CI, it reads the latest git tag and computes/writes the SPM checksum of `Sparkle-for-Swift-Package-Manager.zip`, and when we build locally via `make release` from the Makefile, `release-move-tag.sh` is invoked which makes a final commit with the spm checksum of the zip included in the Package.swift, and the moves the latest tag (which we created prior) to that commit.

tonyarnold:
Yeah, I hadn't accounted for the double-zip bug in `actions/upload-artifact` - it doesn't make any difference when you're downloading directly, but there's got to be a better way to handle this. Perhaps we can avoid artefacts entirely in this instance.

> We haven't made the decision to drop the non-xcframework variant so we may still need to provide it

That's fine - no problem at all. I'll add the framework variant back. 

> the xcframework variant for SPM needs to be a .zip and it does not presently compress nearly as nicely compared to the tar.xz version

Not a problem - providing the frameworks compressed as both zip and a gzipped tape archive are fine 👌🏻 Maybe we could shift it over the bzip2 or lzma and potentially save even more space?

> it reads the latest git tag and computes/writes the SPM checksum 

Yeah, this will be tricky. Ideally, the CI user has no ability to change the repo it is building from, but this is definitely a problem.  I'll give it some thought, but the obvious (but perhaps not great) answer is to let the workflow calculate that checksum and update the package manifest in the repo directly (ie, follow through on the non-CI script and then push the changes back). 

I wonder how other projects are handling this?

tonyarnold:
Ok, I've updated the workflow to not use artefacts at all, and to publish both the .tar.xz and the .zip. 

It will now also push the moved tag back up to the server.

zorgiepoo:
`make release` actually outputs the build output in some temporary directory you can't determine ahead of time, so you may just need to run xcodebuild specifying the `build` directory and run `release-move-tag.sh` yourself.

Do you know what Xcode version is being used? In the pull request CI workflow we specify it by specifying the `DEVELOPER_DIR` (I think at this moment we want Xcode 12.5 on Big Sur which is the latest stable).

If I understand right, we will need to commit/push a tag we want first before running this workflow, so there will be a small window where the tag pushed is wrong. Is there any way to pass the tag we want as a parameter so we can just push the tag remotely once..? Otherwise this may be an acceptable consequence. (let's start with tagging `2.0.0-beta.1` by the way.. Note before we made a mistake of using a tag like 2.0.0beta1 which didn't follow semver with the missing dash so let's not repeat that mistake)

In make-release-package.sh we still likely need to change `zip -rqyX -9` to `ditto` as I mentioned in an earlier comment. In this script we actually extract the non-spm tar archive to a temp directory, run codesign --verify --deep on all binaries to make sure they were archived/generated correctly. We may want/need to do the same thing with the spm archive we create. GitHub Actions for example had a prior bug where they were generating corrupt archives (https://github.com/actions/virtual-environments/issues/2619) and this can catch that sort of issue some of the time..

make-release-package.sh currently uses a dummy git tag when ran in CI. It will need to use the actual tag now.. See the comments in https://github.com/sparkle-project/Sparkle/pull/1751#issuecomment-773582745 about unit tests failing. Because we specify `fetch-depth` as 0 now (and we didn't used to back then), we may be able to retrieve the latest tag now in CI; we just need to simply update the code to omit the `$CI` check.

> Not a problem - providing the frameworks compressed as both zip and a gzipped tape archive are fine 👌🏻 Maybe we could shift it over the bzip2 or lzma and potentially save even more space?

We used to use bzip2 I think, Kornel has experimented around with better compression.. SPM binary packages enforces us to use zip though as far as I know.

zorgiepoo:
Thanks for writing this by the way, this can potentially save us lot of work, if it works!

> I've no way of testing this properly, so someone from the core team will probably need to take what I've done here, try it and fix any issues that arise when they are doing the first Sparkle 2.x prerelease.

@kornelski feel free to give Tony access to test this and iron out issues. It may save us work and we should create a beta 1 tag anyway =)

zorgiepoo:
There is also cocoa pods but I think we should just do that push manually, and (..maybe..) ignore it for beta versions. The important bit is that `$CURRENT_PROJECT_VERSION` needs to match the version in the pod spec file because that is what make-release-package.sh checks. For beta releases we could change `$CURRENT_PROJECT_VERSION` to include a beta number and then it would sync to the pods file and Package.swift file.. although I don't know if it's strictly necessary, it might be..

tonyarnold:
I feel like after we've automated the releases via GH Actions, there are probably a number of other things we can rationalise and improve. 

> Is there any way to pass the tag we want as a parameter so we can just push the tag remotely once..?

Yes, we could make it a manual workflow that you execute directly and provide a string as the tag. 

Initially, someone would then need to go in and fill in the release notes, etc - but all of that _could_ be automated. Reading the changelog into the release is feasible, so long as it's kept up-to-date pretty religiously. 

Let's get this working as a manual workflow to start with, and then we can talk through improving it so that nobody needs to be doing any builds or uploads from their personal machines.

tonyarnold:
> make release actually outputs the build output in some temporary directory you can't determine ahead of time, so you may just need to run xcodebuild specifying the build directory and run release-move-tag.sh yourself.

It's actually fine - if you predefine the `BUILDDIR` environment variable before running, it will use that instead of the temporary directory. 

> Do you know what Xcode version is being used? In the pull request CI workflow we specify it by specifying the DEVELOPER_DIR (I think at this moment we want Xcode 12.5 on Big Sur which is the latest stable).

Whatever the default is - I believe it's Xcode 12.5.1 on GitHub Actions right now. For building releases, I feel like this is probably the right call, but yes, it should be locked so that we're not automatically updated to Xcode 13 it becomes the default.

.github/workflows/create-draft-release.yml
```diff
@@ -0,0 +1,53 @@
...
+            git add Package.swift
+            git commit -m "Update Package.swift for version ${{ github.event.inputs.tag }}"
+            git push
+
+      - name: "Draft a release"
+        uses: softprops/action-gh-release@v1
+        with:
+          draft: true
+          name: "Sparkle ${{ github.event.inputs.tag }}"
+          tag_name: ${{ github.event.inputs.tag }}
```
zorgiepoo, line 49: I didn't actually know you could pass an input. Does this get used or still WIP? move_tag() seems to get message from an existing tag.

zorgiepoo: Never mind, I see it used above.

zorgiepoo:
> Yes, we could make it a manual workflow that you execute directly and provide a string as the tag.

Curious how you pass an input in the drafts page, but neat that this is possible.

Some things I mentioned above:

[ ] ditto instead of zip in make-release-package
[ ] codesign verification of extracted binaries from SPM archive
[ ] give Tony access to run / make a tag

.github/workflows/create-draft-release.yml
```diff
@@ -0,0 +1,53 @@
...
+
+    steps:
+      - name: "Checkout sources"
+        uses: actions/checkout@v2
+        with:
+          submodules: true
+          fetch-depth: 0
+
+      - name: "Build release distribution"
+        run: make release
```
zorgiepoo, line 31: We may not need to create a tag yet, but we need to provide make-release-package.sh the tag name (see `$latest_git_tag` usage)

tonyarnold:
> Curious how you pass an input in the drafts page, but neat that this is possible

It's not happening on the drafts page. I believe the process is that you go to the "Actions" tab, and then select the "Create Draft Release" workflow - there should be a button in the detail of that page that allows you to trigger the workflow, and _that_ is where the input field for this value will be. 

The workflow will then build all of the binary bits, and make a very basic draft release that you can then visit to tag and publish. 

Long story short, the tag isn't created at all as part of this workflow - it won't be created until you publish the draft release that the workflow creates. This gives you the opportunity to attach release notes and double-check that the binaries are OK before making it final.

It does make me wonder if **maybe we should be using input value to populate the `CURRENT_MARKETING_VERSION` and use that in every place that we currently look up the tag on disk**?

zorgiepoo:
> It's not happening on the drafts page. I believe the process is that you go to the "Actions" tab, and then select the "Create Draft Release" workflow - there should be a button in the detail of that page that allows you to trigger the workflow, and _that_ is where the input field for this value will be.
> 
> The workflow will then build all of the binary bits, and make a very basic draft release that you can then visit to tag and publish.
> 
> Long story short, the tag isn't created at all as part of this workflow - it won't be created until you publish the draft release that the workflow creates. This gives you the opportunity to attach release notes and double-check that the binaries are OK before making it final.

This sounds good and reasonable.

> It does make me wonder if maybe we should be using input value to populate the CURRENT_MARKETING_VERSION and use that in every place that we currently look up the tag on disk?

I think we will need to pass an input for the CURRENT_TAG_NAME or something. The CURRENT_PROJECT_VERSION needs to be equal to it, or a CURRENT_TAG_NAME needs to have a prefix "$CURRENT_PROJECT_VERSION-" in case of prerelease identifiers (CURRENT_PROJECT_VERSION which is used for bundle versions shouldn't include prerelease identifier info)

We would then need to read the current tag name we pass, instead of trying to fetch the latest tag on disk as you say, to:
* Update git-version-info.sh so short version string of various bundles is set to the new tag name (git hash won't be needed)
* Updating version and tag in Package.swift
* Updating version in podspec file (and fixing the pre-validation check about version == CURRENT_PROJECT_VERSION)
* Updating the Sparkle*.tar.(xz|bz2) filename.

tonyarnold:
What we could consider doing is allowing the input of a value into the workflow that can be in SemVer format, ie: 2.0.0-beta.1, and use a script to strip off the prerelease information from that when it is set in the bundle. So the tag would be correct (`2.0.0-beta.1`), and the `CFBundleVersion`/`CFBundleShortVersionString` would be `2.0.0`. 

This does raise questions in my mind, though… I've always used a monotonically increasing integer for the `CFBundleVersion` value in my app's plists, but the docs are very clear about only using a strict non-prerelease version of SemVer (as you pointed out). 

How do we resolve this? It's fine to mark a build as pre-release on GitHub, but betas and the like should be clearly marked as thus in their metadata/Info.plist. How has this been handled for Sparkle in the past? I'd like to be consistent with the project's standards, if they're not broken.

zorgiepoo:
Sparkle has a post install script (set-git-version-info.sh) that modifies the short version string on the framework and various bundles to be the current project version ($CURRENT_PROJECT_VERSION) appended by the tag/hash information (after the first dash from  `git describe --tags --match '[12].*'`). If a tag was just created, this means the short version string will be exactly that tag name, I believe. Because the tag isn't created as part of the workflow you're advising, you would thus need to pass the tag name as input to this script.

In my opinion the CFBundleVersion and CURRENT_PROJECT_VERSION should not include pre-release info, leave it to the short version string which will show up when you get info in Finder. For 2.x's entire lifespan so far, the CFBundleVersion has basically never changed, but the short version string has meaningfully been updated on every commit.

The apple docs may say you shouldn't include prerelease info in the short version string, but I would really not worry about that too much, and it's less important for a framework bundle versus an app IMO.

> What we could consider doing is allowing the input of a value into the workflow that can be in SemVer format, ie: 2.0.0-beta.1, and use a script to strip off the prerelease information from that when it is set in the bundle.

This could be a reasonable approach for updating the CURRENT_PROJECT_VERSION in the project config. Or we could more manually require the CURRENT_PROJECT_VERSION to be updated beforehand when we need to update the CHANGELOG manually.

tonyarnold:
For my own projects, I use the result of the following as my `CURRENT_PROJECT_VERSION`: 

```bash
git rev-list --count HEAD | sed -e 's/^[ \t]*//'
```

This produces an integer that increases on every commit to the current branch. If we run this against Sparkle's 2.x branch right now, it produces a build of `3794`. 

Thoughts?

zorgiepoo:
That seems less valuable than the `git describe` invocation we use for the short version string which appends a hash and I don't think it's solving a significant problem. We do want a user to Get Info on a Sparkle framework bundle and be able to tell from the short version what build it is.

I think we need to maintain the current display version and base project version. So my first thought would be in our ConfigCommon.xcconfig file doing:

```
CURRENT_PROJECT_VERSION = 2.0.0 // used by dylib version, set-git-version-info script, CFBundleVersion, maybe(?) C preprocessor
CURRENT_PROJECT_DISPLAY_VERSION_SUFFIX = "-beta.1" // if it's empty, there's no pre-release identifier
CURRENT_PROJECT_DISPLAY_VERSION = $(CURRENT_PROJECT_VERSION)$(CURRENT_PROJECT_DISPLAY_VERSION_SUFFIX) // used for distribution filenames, CFBundleShortVersionString, SPM, cocoapods
```

tonyarnold:
Could I inject an argument that the CURRENT_PROJECT_VERSION _should_ increase over time with each build? I know it's not that big a deal for a framework to have incorrect details on disk, but it sets a bad example for the wider community who _do_ need to increase this value over time. They'll potentially look at examples like this and follow them. 

Ultimately, I'll follow whatever guidance you give here, but I felt like I needed to inject this. 

Also, the project should probably switch to using `CURRENT_PROJECT_VERSION` (for `CFBundleVersion`) and `MARKETING_VERSION` (for `CFBundleShortVersionString` ) consistently as they are used within Xcode's build settings.

zorgiepoo:
I think I'm ok with something like this:

```
// It doesn't matter what value this initially is, as long as we increment it monotonically and it starts at a value > 2
CURRENT_PROJECT_VERSION = 174

// leave empty if there's no prerelease, otherwise it is in format of eg -beta.1
MARKETING_VERSION_PRERELEASE_SUFFIX = "-beta.1"

MARKETING_VERSION = $(SPARKLE_VERSION_MAJOR).$(SPARKLE_VERSION_MINOR).$(SPARKLE_VERSION_PATCH)${MARKETING_VERSION_PRERELEASE_SUFFIX}
```

I don't care to update a monotonically increasing bundle version on every commit, only on releases that we tag; this is a proper approach that apps take during development. So basically this is an edge case for beta releases we may publish soon. I also don't want to add more to our post install scripts (I'd more like to remove them but alas we need the set-git-version info one at least to ensure the XPC Services / frameworks are aligned on a commit level). If we publish a release locally (I want to keep this workflow working as a fallback), we can manually update the project version; we'd have to update the marketing version and CHANGELOG alongside anyway.

It's not necessary to depend on git to increment the counter (can read current value and write value + 1). However, if it's easier that way, then I don't see why not. The set git version info script will still be needed as well as the counter/short git hash it outputs, which is a more reliable indicator to see if things are aligned or what build a user has. There's a chance making this change will break the git script on pre-releases and the SPUXPCValidateServiceIfBundleExists() function, so those likely will need to be updated.

tonyarnold:
> I don't care to update a monotonically increasing bundle version on every commit, only on releases that we tag; this is a proper approach that apps take during development. 

You're absolutely right: for simple projects, the count of the rev list is fine, but it's not reliable - especially if there is ever an intent to go back and release patch updates to older releases, or if you're branching all over the place.

It sounds to me that sticking with a manual versioning approach works better for how this project currently runs? That's fine - we can just leave all of that out of the GitHub Actions workflow, and version details will be the responsibility of the developer, not CI.

**I've updated the Actions workflow to extract the marketing version to use as the draft tag, and the name of the draft release on Github**, so that it all needs to be setup in the repo beforehand. The workflow won't ask for the tag details, as those should already be available in the repo.

> The set git version info script will still be needed as well as the counter/short git hash it outputs, which is a more reliable indicator to see if things are aligned or what build a user has. 

I've updated the script to extract the commit hash properly, even with a prerelease suffix present. Was there more in `SPUXPCValidateServiceIfBundleExists()` that needed to be done?

Sorry, I know this is dragging on - I'm keen to get it right.

zorgiepoo:
> You're absolutely right: for simple projects, the count of the rev list is fine, but it's not reliable - especially if there is ever an intent to go back and release patch updates to older releases, or if you're branching all over the place.

> It sounds to me that sticking with a manual versioning approach works better for how this project currently runs? That's fine - we can just leave all of that out of the GitHub Actions workflow, and version details will be the responsibility of the developer, not CI.

I think this sounds fine.

> I've updated the script to extract the commit hash properly, even with a prerelease suffix present. Was there more in SPUXPCValidateServiceIfBundleExists() that needed to be done?

Let me review other parts of the PR first and then I can address the logic there (based on what we end up doing aside from this)..

.github/workflows/create-draft-release.yml
```diff
@@ -0,0 +1,63 @@
...
+        env:
+          GITHUB_ACTOR: ${{ github.actor }}
+          GITHUB_REPOSITORY: ${{ github.repository }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: "Extract version information from project"
+        run: |
+          MARKETING_VERSION=$(xcrun xcodebuild -showBuildSettings -project Sparkle.xcodeproj | grep MARKETING_VERSION | cut -f 7 -d ' ')
```
zorgiepoo, line 35: Can you enlighten me on the understandings of this unix cut incantation, specifically the `-f 7` part.

.github/workflows/create-draft-release.yml
```diff
@@ -0,0 +1,63 @@
...
+          echo ::set-env name=MARKETING_VERSION::${MARKETING_VERSION}
```
zorgiepoo, line 36: I had just found this article on google this but is this going to be a problem setting env variables like this? https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

Configurations/ConfigCommon.xcconfig
```diff
@@ -20,9 +20,11 @@ SPARKLE_ICON_NAME = AppIcon
 SPARKLE_VERSION_MAJOR = 2
 SPARKLE_VERSION_MINOR = 0
 SPARKLE_VERSION_PATCH = 0
-SPARKLE_VERSION = $(SPARKLE_VERSION_MAJOR).$(SPARKLE_VERSION_MINOR).$(SPARKLE_VERSION_PATCH)
+SPARKLE_VERSION_SUFFIX =
+SPARKLE_VERSION = $(SPARKLE_VERSION_MAJOR).$(SPARKLE_VERSION_MINOR).$(SPARKLE_VERSION_PATCH)$(SPARKLE_VERSION_SUFFIX)
 
 CURRENT_PROJECT_VERSION = $(SPARKLE_VERSION)
```
zorgiepoo, line 26: You do want to update this to some hardcoded counter we can manually increase right? Maybe we can start it with `2000`..

Configurations/ConfigFramework.xcconfig
```diff
@@ -1,8 +1,8 @@
 // Framework only
 
 DYLIB_INSTALL_NAME_BASE = @rpath
-DYLIB_COMPATIBILITY_VERSION = 1.6
-DYLIB_CURRENT_VERSION = $(SPARKLE_VERSION_MAJOR).$(SPARKLE_VERSION_MINOR).$(SPARKLE_VERSION_PATCH)
+DYLIB_COMPATIBILITY_VERSION = $(SPARKLE_VERSION_MAJOR).0.0
+DYLIB_CURRENT_VERSION = $(SPARKLE_VERSION)
```
zorgiepoo, line 5: The dylib version is not supposed to contain $SPARKLE_VERSION_SUFFIX. `$(SPARKLE_VERSION_MAJOR).$(SPARKLE_VERSION_MINOR).$(SPARKLE_VERSION_PATCH)` was correct.

Configurations/ConfigFramework.xcconfig
```diff
@@ -1,8 +1,8 @@
 // Framework only
 
 DYLIB_INSTALL_NAME_BASE = @rpath
-DYLIB_COMPATIBILITY_VERSION = 1.6
-DYLIB_CURRENT_VERSION = $(SPARKLE_VERSION_MAJOR).$(SPARKLE_VERSION_MINOR).$(SPARKLE_VERSION_PATCH)
+DYLIB_COMPATIBILITY_VERSION = $(SPARKLE_VERSION_MAJOR).0.0
```
zorgiepoo, line 4: We should not change the compatibility version. On several occasions I debug apps built using Sparkle 1 with  using Sparkle 2 (dyld inserting the framework or otherwise). This will likely break that. Also Sparkle 2 maintains ABI compatibility with Sparkle 1 versions.

Configurations/ConfigFramework.xcconfig
zorgiepoo: Unless you meant for `SPARKLE_VERSION` to not include the suffix. I don't think this variable is used anywhere else.

Configurations/release-move-tag.sh
```diff
@@ -4,23 +4,28 @@ set -e
 # Convenience script to automatically commit Package.swift after updating the checksum and move the latest tag
 latest_git_tag=$(git describe --abbrev=0) # gets the latest tag name
```
zorgiepoo, line 5: We still have to update this right? Otherwise it's going to get the wrong tag name since we are creating the tag after the workflow? Recall the tag has to be created after this script updates the Package.swift

Configurations/make-release-package.sh
```diff
@@ -115,7 +115,7 @@ if [ "$ACTION" = "" ] ; then
     # Generate zip containing the xcframework for SPM
     cd "$CONFIGURATION_BUILD_DIR/staging-spm"
     #rm -rf "$CONFIGURATION_BUILD_DIR/Sparkle.xcarchive"
-    zip -rqyX -9 "../Sparkle-for-Swift-Package-Manager.zip" *
+    ditto -c -k --zlibCompressionLevel 9 --rsrc . "../Sparkle-for-Swift-Package-Manager.zip"
     
     # GitHub actions set the CI environment variable to true
     if [ "$CI" != true ]; then
```
zorgiepoo, line 121: When we are doing this specific workflow in CI, we will need to use the marketing version as the tag name. So Package.swift will be updated, committed, and pushed appropriately.

.github/workflows/create-draft-release.yml
zorgiepoo: As part of this workflow, we don't want to run release-move-tag.sh or any of the other commands except for the xcodebuild one in `make release` to my understanding.

Configurations/set-git-version-info.sh
```diff
@@ -22,8 +22,8 @@ if [ -z "$gitversion" ] ; then
     exit 0
 fi
 
-# remove everything before the first "-" to keep the hash part only
-versionsuffix=${gitversion#*-};
+# remove everything before the last "-" to keep the hash part only
+versionsuffix=${gitversion##*-};
 if [ "$versionsuffix" != "$gitversion" ]; then
     version="$version $versionsuffix"
```
zorgiepoo, line 28: This should probably be `$version-$versionsuffix` for consistency.

Configurations/set-git-version-info.sh
```diff
@@ -22,8 +22,8 @@ if [ -z "$gitversion" ] ; then
     exit 0
 fi
 
-# remove everything before the first "-" to keep the hash part only
-versionsuffix=${gitversion#*-};
+# remove everything before the last "-" to keep the hash part only
+versionsuffix=${gitversion##*-};
```
zorgiepoo, line 26: This comment was slightly wrong. We actually used to append both a counter (since the last tag) and a git short hash. For example on `master`, we currently get `8-gf7d830d9`. This counter does strengthen the uniqueness of this identifier a bit more, so we may not want to lose it.

zorgiepoo: In short, everything past the 2nd last dash.

zorgiepoo: But anyway.. we will need to actually feed this script with the marketing version when ran in this workflow. Because in this workflow we actually create the tag *afterwards*.

zorgiepoo:
So there's a couple of places that assume the tag was created before building things. Rather than creating a divergence between publishing releases locally versus in CI, let's just create a tag as part of this workflow, before running make release. However, we do *not need to* push the tag and we can still create it in the draft page.

.github/workflows/create-draft-release.yml
zorgiepoo: release-move-tag.sh also interactively asks the user a question so we can't allow that to happen in CI

Configurations/make-release-package.sh
zorgiepoo: Everything in this make-release-package.sh script needs to use `MARKETING_VERSION` now, not `CURRENT_PROJECT_VERSION`

Configurations/set-git-version-info.sh
zorgiepoo: We also need to use MARKETING_VERSION in this script now, not CURRENT_PROJECT_VERSION

zorgiepoo: Ahh also, when the counter is zero (when we're directly at the tag), git describe outputs exactly that tag name, with no hash info (eg `2.0.0-beta.1`). But as this example shows there can still be a dash due to the prerelease naming scheme. When there is a hash, you can be guaranteed it ends in `-g*` though.

zorgiepoo: Or replicate/check release-move-tag.sh logic if you want to detect if we are more than 0 commits away from latest tag.

.github/workflows/create-draft-release.yml
zorgiepoo: On second thought, I think we should still run release-move-tag.sh - but we should make it automatically make the commit / move the tag and not interactively ask the user.

.github/workflows/create-draft-release.yml
```diff
@@ -0,0 +1,63 @@
...
+          if [[ -z "$SPARKLE_VERSION_SUFFIX" ]]; then
+            echo ::set-env name=PRERELEASE_VERSION::true
+          else
+            echo ::set-env name=PRERELEASE_VERSION::false
+          fi
+
+      - name: "Commit and push updated Swift package description"
+        run: |
+            git add Package.swift
+            git commit -m "Update Package.swift for version ${{ env.MARKETING_VERSION }}"
```
zorgiepoo, line 50: The push is necessary but the add and commit are not necessary if release-move-tag.sh does this. In the future I may want to add/commit the podspec file as well, and release-move-tag.sh seems to be a centralized place to make such a change.

.github/workflows/create-draft-release.yml
zorgiepoo: One other approach is that we pass MARKETING_VERSION and PROJECT_CURRENT_VERSION as input in the GitHub workflow after all, and then the workflow overrides those values in CommonConfig.xcconfig and `git add`'s that change. I'm fine with either approach.

tonyarnold:
Thanks for the thorough review, @zorgiepoo. I'm currently recovering from pneumonia (in the middle of a pandemic… _awesome_!) so please give me a few days to work through the feedback you've provided, and the changes you've requested.

zorgiepoo:
Ahh that sucks. Have a good recovery ❤️; there's no rush here.

.github/workflows/create-draft-release.yml
tonyarnold: Good catch. I've switched to the new [environment files](https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#environment-files) approach. 

> is setting an env variable necessary or there is no other way to pass info from one step to another?

There are other methods, but this was the simplest way of getting simple key value pairs across steps in the workflow. 

I think the question I'd ask in response to the suggestions here is: do you want to allow unvalidated free text input to determine the contents of MARKETING_VERSION and PROJECT_CURRENT_VERSION? I guess it really doesn't make much difference - my main issue here is that I'm wary of committing changes from within CI workflows.

.github/workflows/create-draft-release.yml
tonyarnold: Right, so my understanding here is that the invocation of cut is targeting only spaces. So the value here is after the 7th space in the string. 

It's not great, but it's super fast. I should probably use a regex here so that it's a bit more reliable.

Configurations/ConfigFramework.xcconfig
tonyarnold: Right, that's very interesting. And brave 😂 

I've reverted this change in 26b8546d8be0df7e4dd1d1fa1a650215812f653d.

Configurations/ConfigFramework.xcconfig
tonyarnold: Nope, I got this wrong. I've reverted it in 85426bf03791c26c257c38e722df6b57125e7236.

Configurations/ConfigCommon.xcconfig
tonyarnold: I've set it to `2000`. If it were my own project, I'd want to automatically increment whatever the committed/current value of this build setting is when tagging a release, but you've explained your reluctance toward this and I understand and am happy to leave it here.

Configurations/set-git-version-info.sh
tonyarnold: 82d6ed9cf9d732b16b348287393b2781030b3a30 should extend the capture of the git hash and count information. 

I believe that the check for `"$versionsuffix" != "$gitversion"` should ensure that the tag name capture is OK, too.

.github/workflows/create-draft-release.yml
tonyarnold: I've updated this behaviour in 3e943bf5f7b45286ab50353032561748f8c33079, but is there still a need to move the tag at all?

.github/workflows/create-draft-release.yml
tonyarnold: This is addressed by 3e943bf5f7b45286ab50353032561748f8c33079.

.github/workflows/create-draft-release.yml
tonyarnold: Done in b2e07a51f8fdd769da8aaded3177968587f4de7d.

Configurations/make-release-package.sh
tonyarnold: Done in e2cd36410d0c3d5357c0128515f0ae19e8e54e39.

tonyarnold: I've also removed the CI check here, as you suggested: 13df5a238eb1049cec4be7b59d3ded19b119d79f

tonyarnold:
Ok, I've made most of the suggested changes, but I'm still not 100% clear on when you'd like the tag created. 

You made mention of creating a tag locally as part of the workflow, but not pushing it? Could you elaborate on what you're thinking there?

.github/workflows/create-draft-release.yml
```diff
@@ -0,0 +1,60 @@
...
+      - name: "Build release distribution"
+        run: make release
+        env:
+          GITHUB_ACTOR: ${{ github.actor }}
+          GITHUB_REPOSITORY: ${{ github.repository }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: "Extract version information from project"
+        run: |
+          MARKETING_VERSION=$(xcrun xcodebuild -showBuildSettings -project Sparkle.xcodeproj | grep MARKETING_VERSION | sed -E 's/.+=[[:space:]]+(.*)$/\1/')
```
zorgiepoo, line 35: I wish xcodebuild just had a command to properly extract individual build settings rather than regex-hacking it.

This could possibly break if MARKETING_VERSION = happens to be inserted in some other place, like `GCC_PREPROCESSOR_DEFINITIONS` for example, which is a feasible thing to happen in the future. So maybe you could make it more resilient at least so that the line (ignoring whitespace) needs to start with MARKETING_VERSION, and it would be a bad error if more than one match was found.

Otherwise at the very least we may need to add a comment in the CommonConfig file being careful about using/exporting MARKETING_VERSION elsewhere.

.github/workflows/create-draft-release.yml
```diff
@@ -0,0 +1,60 @@
...
+          echo "MARKETING_VERSION=${MARKETING_VERSION}" >> $GITHUB_ENV~
+
+      - name: "Determine if this is a pre-release version"
+        run: |
+          SPARKLE_VERSION_SUFFIX=$(xcrun xcodebuild -showBuildSettings -project Sparkle.xcodeproj | grep SPARKLE_VERSION_SUFFIX | cut -f 7 -d ' ')
```
zorgiepoo, line 40: Not sure if you wanted to use sed here too?

.github/workflows/create-draft-release.yml
```diff
@@ -0,0 +1,60 @@
...
+          if [[ -z "$SPARKLE_VERSION_SUFFIX" ]]; then
+            echo "PRERELEASE_VERSION=true" >> $GITHUB_ENV
```
zorgiepoo, line 42: Are you sure this is right? Google tells me -z returns true when the string is empty.

Configurations/set-git-version-info.sh
```diff
@@ -22,10 +22,10 @@ if [ -z "$gitversion" ] ; then
     exit 0
 fi
 
-# remove everything before the first "-" to keep the hash part only
-versionsuffix=${gitversion#*-};
+# remove everything before the second last "-" to keep the hash part only
+versionsuffix=$( echo "${gitversion}" | sed -E 's/.+((-[^.]+){2})$/\1/' )
 if [ "$versionsuffix" != "$gitversion" ]; then
```
zorgiepoo, line 27: Tested several strings locally and I think this works well 👌

Configurations/release-move-tag.sh
```diff
@@ -4,23 +4,30 @@ set -e
...
 commits_since_tag=$(git rev-list ${latest_git_tag}.. --count)
+
+function move_tag() {
+    long_message=$(git tag -n99 -l $latest_git_tag) # gets corresponding message
+    long_message=${long_message/$latest_git_tag} # trims tag name
+    long_message="$(echo -e "${long_message}" | sed -e 's/^[[:space:]]*//')" # trim leading whitespace
+    git add Package.swift
+    git commit -m "Update Package.swift for version ${latest_git_tag}"
+    git tag -fa $latest_git_tag -m "${long_message}"
```
zorgiepoo, line 14: To answer your question, creating/moving the tag (git tag) in CI is probably not necessary because we don't push it. But the git add and git commit parts are needed.

Also note make release and release-move-tag.sh are not invoked in our other normal CI workflow that triggers on pull requests, which is a good thing here.. because we wouldn't want this procedure to be invoked in that case.

Configurations/ConfigCommon.xcconfig
```diff
@@ -20,9 +20,10 @@ SPARKLE_ICON_NAME = AppIcon
 SPARKLE_VERSION_MAJOR = 2
 SPARKLE_VERSION_MINOR = 0
 SPARKLE_VERSION_PATCH = 0
-SPARKLE_VERSION = $(SPARKLE_VERSION_MAJOR).$(SPARKLE_VERSION_MINOR).$(SPARKLE_VERSION_PATCH)
+SPARKLE_VERSION_SUFFIX =
```
zorgiepoo, line 23: let's add a comment illustrating what the expected format of this should be when not empty (eg -beta.1)

zorgiepoo: also a comment that if we change the major/minor/patch/suffix, we should bump the project version.

.github/workflows/create-draft-release.yml
zorgiepoo: See https://github.com/sparkle-project/Sparkle/pull/1908/files#r687424758

zorgiepoo:
> Ok, I've made most of the suggested changes, but I'm still not 100% clear on when you'd like the tag created.
> 
> You made mention of creating a tag locally as part of the workflow, but not pushing it? Could you elaborate on what you're thinking there?

We need to have a step that runs this before running `make release`:
(something like:)

`git tag "$MARKETING_VERSION"`

Because our workflow and scripts in general assume the tag info is up to date before making the build.

However, we *do not need* to push the tag to the remote via:

`git push origin "$MARKETING_VERSION"` (or `git push --tags` which may be more familiar to people; tags are not automatically pushed via vanilla `git push`)

So that would still allow us to look at the binaries uploaded, and draft, and create a remote tag via github's interface when ready.

zorgiepoo:
The default `-[SPUUpdater userAgentString]` implementation needs to be updated now. Instead of reading from Sparkle's bundle it should just use `MARKETING_VERSION`. This may then mean you need to export it in the preprocessor definitions and fix any `sed` ambiguities as discussed earlier.

Configurations/make-release-package.sh
```diff
@@ -115,35 +115,31 @@ if [ "$ACTION" = "" ] ; then
     # Generate zip containing the xcframework for SPM
     cd "$CONFIGURATION_BUILD_DIR/staging-spm"
     #rm -rf "$CONFIGURATION_BUILD_DIR/Sparkle.xcarchive"
-    zip -rqyX -9 "../Sparkle-for-Swift-Package-Manager.zip" *
+    ditto -c -k --zlibCompressionLevel 9 --rsrc . "../Sparkle-for-Swift-Package-Manager.zip"
```
zorgiepoo, line 118: We have code (right above this) to extract the *.xz|bz2 archive, validate the code signature of all the binaries/bundles there. That was added due to history of archiving issues I've seen occur, particularly in GitHub Actions. Just in case, we may want to do the same thing with the SPM zip (and this partly reason I suggested changing to ditto). This may mean pulling that verification code into its own function.

tonyarnold:
> The default -[SPUUpdater userAgentString] implementation needs to be updated now. Instead of reading from Sparkle's bundle it should just use MARKETING_VERSION. This may then mean you need to export it in the preprocessor definitions and fix any sed ambiguities as discussed earlier.

Are you sure you want to involve preprocessor definitions in this? (assuming you're not already) This'll prevent any future move to Swift.

Configurations/make-release-package.sh
tonyarnold: Are the SPM binaries supposed to be code signed? 

I've updated the script locally, but it throws an error:

```
/var/folders/x1/pg2m1zvx2c32tm_7pckc8fzw0000gn/T/Sparkle.c1W0L3/Build/Products/Release/staging-spm/Sparkle.xcframework: code object is not signed at all
```

zorgiepoo: Yes they are ad-hoc signed, but I forgot that xcframework is not code bundle. Sparkle.framework is somewhere inside of it. Codesign verification on the xcframework won't make sense.

tonyarnold: I'll verify the individual child frameworks.

zorgiepoo:
> > The default -[SPUUpdater userAgentString] implementation needs to be updated now. Instead of reading from Sparkle's bundle it should just use MARKETING_VERSION. This may then mean you need to export it in the preprocessor definitions and fix any sed ambiguities as discussed earlier.
> 
> Are you sure you want to involve preprocessor definitions in this? (assuming you're not already) This'll prevent any future move to Swift.

I don't know why this is not compatible with Swift but we are exporting other preprocessor definitions like this anyway, and I don't imagine there being Swift code in the framework until at least 10.14.4 becomes a minimum requirement, which will be a long way from now.

tonyarnold:
> I don't imagine there being Swift code in the framework until at least 10.14.4 becomes a minimum requirement, which will be a long way from now

That was what I needed to know. All good - I'll add the preprocessor definition now.

Configurations/ConfigCommon.xcconfig
tonyarnold: Done in 9cff8f9f5cf6007a33e38dda6954331678f7241c.

.github/workflows/create-draft-release.yml
tonyarnold: You're right - fixed in 8f55ebc28c69e4f9e8f73d40de23b24534fe8210.

.github/workflows/create-draft-release.yml
tonyarnold: I did! Fixed in ed76aac8beadbc3d6647dd643394157751901540.

.github/workflows/create-draft-release.yml
tonyarnold: Yeah, I wish this were true, too. 

I can't think of a great safety check for other variables containing MARKETING_VERSION here, so I'll add a comment to the Xcode config so that we can keep moving.

tonyarnold: I've added the comment in 18b5da5db03e17bb20144c2fd6e85ee6f55d5ada.

.github/workflows/create-draft-release.yml
```diff
@@ -0,0 +1,63 @@
...
+        run: |
+          SPARKLE_VERSION_SUFFIX=$(xcrun xcodebuild -showBuildSettings -project Sparkle.xcodeproj | grep SPARKLE_VERSION_SUFFIX | sed -E 's/.+=[[:space:]]+(.*)$/\1/')
+          if [[ ! -z "$SPARKLE_VERSION_SUFFIX" ]]; then
+            echo "PRERELEASE_VERSION=true" >> $GITHUB_ENV
+          else
+            echo "PRERELEASE_VERSION=false" >> $GITHUB_ENV
+          fi
+
+      - name: "Create a temporary Git tag"
+        run: git tag "$MARKETING_VERSION"
```
zorgiepoo, line 48: This has to be done before `make release` which will mean you may need to extract the version information before make release step too.

tonyarnold: Ah, poop - yeah, right. Missing simple stuff here, sorry.

Configurations/make-release-package.sh
```diff
@@ -80,70 +104,60 @@ if [ "$ACTION" = "" ] ; then
...
     # Generate zip containing the xcframework for SPM
     cd "$CONFIGURATION_BUILD_DIR/staging-spm"
-    #rm -rf "$CONFIGURATION_BUILD_DIR/Sparkle.xcarchive"
-    zip -rqyX -9 "../Sparkle-for-Swift-Package-Manager.zip" *
+    rm -rf "$CONFIGURATION_BUILD_DIR/Sparkle.xcarchive"
+    ditto -c -k --zlibCompressionLevel 9 --rsrc . "../Sparkle-for-Swift-Package-Manager.zip"
+
+    # Test code signing validity of the extracted Swift package
+    # This guards against our archives being corrupt / created incorrectly
+    verify_code_signatures "$CONFIGURATION_BUILD_DIR/staging-spm"
```
zorgiepoo, line 136: We would want to verify the signatures of the extracted zip file, not the contents before being zipped. Should be a small change I think. To mirror above, I think this ought to do it:

```
ditto -x -k "../Sparkle-for-Swift-Package-Manager.zip" /tmp/sparkle-extract
verify_code_signatures "/tmp/sparkle-extract"

rm -rf "/tmp/sparkle-extract"
rm -rf "$CONFIGURATION_BUILD_DIR/staging-spm"

File: .github/workflows/create-draft-release.yml [hunk 22]
```diff
@@ -0,0 +1,65 @@
...
+
+    steps:
+      - name: "Checkout sources"
+        uses: actions/checkout@v2
+        with:
+          submodules: true
+          fetch-depth: 0
+
+      - name: "Create a temporary Git tag"
+        run: git tag "$MARKETING_VERSION"
```
zorgiepoo, line 27: We need to extract the MARKETING_VERSION before this step now.

.github/workflows/create-draft-release.yml
```diff
@@ -0,0 +1,65 @@
...
+
+      - name: "Extract version information from project"
+        run: |
+          MARKETING_VERSION=$(xcrun xcodebuild -showBuildSettings -project Sparkle.xcodeproj | grep MARKETING_VERSION | sed -E 's/.+=[[:space:]]+(.*)$/\1/')
```
zorgiepoo, line 31: I think multiple results are going to be matched now because of `MARKETING_VERSION=\"$(MARKETING_VERSION)\"` in the exported preprocessor definitions, but please re-check and handle this as necessary.

zorgiepoo: Also possibly the comment about being careful to use `MARKETING_VERSION`..

zorgiepoo: I think in your `grep` part you need to make sure the line starts with `MARKETING_VERSION`. At least that would be one way of doing it.

zorgiepoo: Eg

`grep -m 1 '^[[:space:]]*CURRENT_PROJECT_VERSION'`

zorgiepoo: Whatever improvement done here should likely also apply to extracting `SPARKLE_VERSION_SUFFIX`

zorgiepoo:
Suggested updated implementation for `SPUXPCValidateServiceIfBundleExists()`:

```objc
BOOL SPUXPCValidateServiceIfBundleExists(NSString *bundleName, NSBundle *sparkleBundle, NSError * __autoreleasing *error)
{
    NSBundle *xpcBundle = SPUXPCServiceBundle(bundleName);
    if (xpcBundle == nil) {
        return YES;
    }
    
#if !DEBUG
    // Post install scripts for appending git hash info to CFBundleShortVersionString are too unreliable to verify for debug/development
    // Fortunately debug builds of Sparkle are not usable in a production environment
    (void)sparkleBundle; // Mark parameter as used
#else
    // Make sure the display versions of Sparkle and the XPC bundle match
    // These versions may contain a git hash identifier if built from the Sparkle repository
    NSString *sparkleDisplayVersion = [sparkleBundle objectForInfoDictionaryKey:@"CFBundleShortVersionString"];
    NSString *xpcDisplayVersion = [xpcBundle objectForInfoDictionaryKey:@"CFBundleShortVersionString"];
    
    if (sparkleDisplayVersion == nil || xpcDisplayVersion == nil || ![sparkleDisplayVersion isEqualToString:xpcDisplayVersion]) {
        if (error != NULL) {
            *error = [NSError errorWithDomain:SUSparkleErrorDomain code:SUInvalidUpdaterError userInfo:@{ NSLocalizedDescriptionKey: [NSString stringWithFormat:@"XPC Version mismatch. Framework display version is %@ but XPC Service (%@) display version is %@. Bundled XPC Service cannot be used.", sparkleDisplayVersion, xpcBundle.bundlePath, xpcDisplayVersion] }];
        }
    }
#endif
    
    // Make sure the bundle versions of Sparkle and the XPC bundle match
    NSString *xpcBundleVersion = [xpcBundle objectForInfoDictionaryKey:(__bridge NSString *)kCFBundleVersionKey];
    
    if (xpcBundleVersion == nil || ![xpcBundleVersion isEqualToString:@""CURRENT_PROJECT_VERSION]) {
        if (error != NULL) {
            *error = [NSError errorWithDomain:SUSparkleErrorDomain code:SUInvalidUpdaterError userInfo:@{ NSLocalizedDescriptionKey: [NSString stringWithFormat:@"XPC Version mismatch. Framework bundle version is %@ but XPC Service (%@) bundle version is %@. Bundled XPC Service cannot be used.", @""CURRENT_PROJECT_VERSION, xpcBundle.bundlePath, xpcBundleVersion] }];
        }
        return NO;
    }
    
    return YES;
}
```

Configurations/release-move-tag.sh
```diff
@@ -4,23 +4,30 @@ set -e
 # Convenience script to automatically commit Package.swift after updating the checksum and move the latest tag
 latest_git_tag=$(git describe --abbrev=0) # gets the latest tag name
```
zorgiepoo, line 5: I just realized this isn't scanning lightweight tags. Can you add `--tags` to this `git describe` invocation? (other scripts already do this).

zorgiepoo:
Please apply this patch as well to update the cocoa pods spec :)

```diff
commit 4f9c89d8b7e1a17c7d2a6cbda1e98a9affae0ecc
Author: Zorg <zorgiepoo@gmail.com>
Date:   Sat Aug 21 13:37:19 2021 -0700

    Update podspec file automatically

diff --git a/Configurations/make-release-package.sh b/Configurations/make-release-package.sh
index b7a2ddf5..8c4093fd 100755
--- a/Configurations/make-release-package.sh
+++ b/Configurations/make-release-package.sh
@@ -26,15 +26,6 @@ function verify_code_signatures() {
 }
 
 if [ "$ACTION" = "" ] ; then
-    # If using cocoapods, sanity check that the Podspec version matches the Sparkle version
-    if [ -x "$(command -v pod)" ]; then
-        spec_version=$(printf "require 'cocoapods'\nspec = %s\nprint spec.version" "$(cat "$SRCROOT/Sparkle.podspec")" | LANG=en_US.UTF-8 ruby)
-        if [ "$spec_version" != "$MARKETING_VERSION" ] ; then
-            echo "podspec version '$spec_version' does not match the marketing version '$MARKETING_VERSION'" >&2
-            exit 1
-        fi
-    fi
-
     rm -rf "$CONFIGURATION_BUILD_DIR/staging"
     rm -rf "$CONFIGURATION_BUILD_DIR/staging-spm"
     rm -f "Sparkle-$MARKETING_VERSION.tar.xz"
@@ -147,9 +138,10 @@ if [ "$ACTION" = "" ] ; then
         exit 1
     fi
 
-    # Generate new Package manifest
+    # Generate new Package manifest and podspec
     cd "$CONFIGURATION_BUILD_DIR"
     cp "$SRCROOT/Package.swift" "$CONFIGURATION_BUILD_DIR"
+    cp "$SRCROOT/Sparkle.podspec" "$CONFIGURATION_BUILD_DIR"
     if [ "$XCODE_VERSION_MAJOR" -ge "1200" ]; then
         # is equivalent to shasum -a 256 FILE
         spm_checksum=$(swift package compute-checksum "Sparkle-for-Swift-Package-Manager.zip")
@@ -160,6 +152,11 @@ if [ "$ACTION" = "" ] ; then
         echo "Version: $MARKETING_VERSION"
         echo "Tag: $latest_git_tag"
         echo "Checksum: $spm_checksum"
+
+        sed -E -i '' -e "/s\.version.+=/ s/\".+\"/\"$MARKETING_VERSION\"/" "Sparkle.podspec"
+        cp "Sparkle.podspec" "$SRCROOT"
+        echo "Sparkle.podspec updated with following values:"
+        echo "Version: $MARKETING_VERSION"
     else
         echo "warning: Xcode version $XCODE_VERSION_ACTUAL does not support computing checksums for Swift Packages. Please update the Package manifest manually."
     fi
diff --git a/Configurations/release-move-tag.sh b/Configurations/release-move-tag.sh
index 0c63867a..818cf969 100755
--- a/Configurations/release-move-tag.sh
+++ b/Configurations/release-move-tag.sh
@@ -9,10 +9,10 @@ function move_tag() {
     long_message=$(git tag -n99 -l $latest_git_tag) # gets corresponding message
     long_message=${long_message/$latest_git_tag} # trims tag name
     long_message="$(echo -e "${long_message}" | sed -e 's/^[[:space:]]*//')" # trim leading whitespace
-    git add Package.swift
-    git commit -m "Update Package.swift for version ${latest_git_tag}"
+    git add Package.swift Sparkle.podspec
+    git commit -m "Update Package.swift and Sparkle.podspec for version ${latest_git_tag}"
     git tag -fa $latest_git_tag -m "${long_message}"
-    echo "Package.swift committed and tag '$latest_git_tag' moved."
+    echo "Package.swift and Sparkle.podspec committed and tag '$latest_git_tag' moved."
 }
 
 if [ "$commits_since_tag" -gt 0 ]; then
@@ -23,11 +23,11 @@ elif [ "$CI" == true ]; then
     move_tag
 else
 # TODO: add sanity check to see if version is actually being updated or not?
-    read -p "Do you want to commit changes to Package.swift and force move tag '$latest_git_tag'? (required for SPM release) [Y/n]" -n 1 -r
+    read -p "Do you want to commit changes to Package.swift and Sparkle.podspec, and force move tag '$latest_git_tag'? (required for official release) [Y/n]" -n 1 -r
         echo    # (optional) move to a new line
     if [[ $REPLY =~ ^[Yy]$ ]]; then
         move_tag
     else
-        echo "Package.swift has not been committed and tag has not been moved."
+        echo "Package.swift and Sparkle.podspec have not been committed and tag has not been moved."
     fi
 fi
```

.github/workflows/create-draft-release.yml
tonyarnold: Your grep command seems to work where my sed doesn't, so I've updated to use that: 2daa27b58e88a6d31235fa46bf7e270706f7711f

.github/workflows/create-draft-release.yml
tonyarnold: That was a silly mistake. Fixed in 388781e5714fee0f72d01718e719608cc83e8e8a.

Configurations/make-release-package.sh
tonyarnold: Done in e87eda83d207cd4b7b2359fa04150f1239817730.

Configurations/release-move-tag.sh
tonyarnold: Done in 363c05a1a87545fa06e42d0cb0892de7662ab8fd.

tonyarnold:
Ok, I've applied the requested patches, and I think I've covered the issues you've raised. 

I apologise for taking so long, and for the shoddy work in the middle there - I've got three children under the age of 10 homeschooling during a lockdown, so things are pretty chaotic. 

Thanks for being patient.

tonyarnold:
I've also rebased onto the latest code from the 2.x branch so the PR should apply cleanly.

Configurations/make-release-package.sh
```diff
@@ -80,72 +95,70 @@ if [ "$ACTION" = "" ] ; then
...
+    verify_code_signatures "/tmp/sparkle-extract"
 
     rm -rf "/tmp/sparkle-extract"
     rm -rf "$CONFIGURATION_BUILD_DIR/staging"
     
     # Generate zip containing the xcframework for SPM
     cd "$CONFIGURATION_BUILD_DIR/staging-spm"
-    #rm -rf "$CONFIGURATION_BUILD_DIR/Sparkle.xcarchive"
-    zip -rqyX -9 "../Sparkle-for-Swift-Package-Manager.zip" *
+    rm -rf "$CONFIGURATION_BUILD_DIR/Sparkle.xcarchive"
```
zorgiepoo, line 122: Not sure why we are removing the xcarchive now -- I don't think we should.

Configurations/make-release-package.sh
zorgiepoo: Not quite, this snippet I wrote here is extracting the .zip into a temp directory and verifying the code signatures of the extraction. Current code is still verifying code signatures inside staging-spm instead.

zorgiepoo:
Just had two small comments, one of them about the [ditto extraction](https://github.com/sparkle-project/Sparkle/pull/1908/files#r693739541), otherwise looks good to me (barring from testing the github workflow). There's no rush on this on our end.

Configurations/make-release-package.sh
tonyarnold: It was commented out when I got here, so I'll just leave it in there commented out: c99a592fbb1930d191fbef53a15501ed3e0ce1d8

zorgiepoo (approved):

zorgiepoo:
Thank you. Changes look good to me although I have not tried the workflow (and it seems like it would be more productive to have changes merged first to try it out). There are also some good project versioning changes here with having the CURRENT_PROJECT_VERSION being an increasing number, and MARKETING_VERSION as what we want to market, as well as making updating the Package.swift and cocoa pods spec files consistent.

I'll give a chance @kornelski to take a look, otherwise I'll just end up merging this. I won't get to trying the workflow myself until maybe the weekend although you are free to try; I'll invite you to the project. We will first need a PR that bumps the marketing version to 2.0.0-beta.1 (and current project version too consequently) though.

thebluepotato:
From a quick read-through of the changes, everything seems ok (and even improved!) on the SPM front

PR #1926, opened 2021-08-29
Fix issues with GitHub release workflow

zorgiepoo:
Fix not referencing MARKETING_VERSION in GitHub workflow properly. We reference it in other spots properly.
Fix project versions not being extracted properly due to lack of sed commands that were accidentally removed.
Fix tag version not being exported to ENV properly.


zorgiepoo:
Awesome. This workflow works properly now cc @tonyarnold 

I'll kick it off soon for real for a 2.x prerelease.

tonyarnold:
This is exciting!

PR #1586, opened 2020-05-11
fix: undesirable popups when checking for updates

lwouis:
Hi!

I use [this page](https://alt-tab-macos.netlify.app/changelog-bare) for release notes. It contains some JavaScript to load google translate if the user language is not English. This works great in a `WebView`, however it creates 3 popups:

image

After digging deeper, I found out that the google translate script loads a bunch of iframes. These have no `src` attribute. They triggers the `decidePolicyForNavigationAction` method in Sparkle, which calls `[[NSWorkspace sharedWorkspace] openURL:requestURL];` which result in the popups.

I don't think there is any point in ever opening a about:blank link for the user, thus this PR ignores these URLs.

---

If you decide to not merge this, could you please help me build on my fork? I ran `make release`, then made a zip out of the result, uploaded it to my fork's releases, then made [these changes to the podspec](https://github.com/lwouis/Sparkle/commit/fb1d08ecf51abc0a8da3982517289a0cbd0d6be3). Also note that I had to do [that tweak](https://github.com/lwouis/Sparkle/commit/17cc2cab09adb1a7c0877c752dc9ea1a1517f545) to build on my mac.

After this I updated the Podfile in my project to: `pod 'Sparkle', :git => 'https://github.com/lwouis/Sparkle.git', :branch => 'master'`.

My project doesn't build and complains that the Sparkle framework is missing. I don't quite understand how Sparkle is packaged/delivered.

Thanks a lot!

Louis Pontoise

PR #1938, opened 2021-09-03
fix: some iframes can use about:srcdoc, which spawns popups (follow-up to #1586)

lwouis:
Hi,

This is follow-up to https://github.com/sparkle-project/Sparkle/pull/1586

It turns out some iframes use `about:srcdoc` instead of `about:blank`, and this creates the same issue as in the original PR.

This PR is a one-liner, so I hope you forgive if I didn't fill the template

Thank you!

### DIFF
--- a/Sparkle/SUWebViewCommon.m
+++ b/Sparkle/SUWebViewCommon.m
@@ -11,7 +11,7 @@
 BOOL SUWebViewIsSafeURL(NSURL *url, BOOL *isAboutBlankURL)
 {
     NSString *scheme = url.scheme;
-    BOOL isAboutBlank = [url.absoluteString isEqualToString:@"about:blank"];
+    BOOL isAboutBlank = [@[@"about:blank", @"about:srcdoc"] containsObject:url.absoluteString];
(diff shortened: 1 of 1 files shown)


zorgiepoo:
@kornelski do you want to review over this one? I think the change is probably fine on its own, but I'm little weary about our safe URL handling logic, and I'm not sure if there are other HTML / iframe cases we're missing.

We also need a PR for 2.x branch for changes to land in the project.

zorgiepoo:
If you can, you may also want to test the change with the legacy webkit 1 implementation for older OS's / apps without sandboxed network entitlement, by temporarily adjusting the code so it takes the path of using `[[SULegacyWebView alloc] init...]`

kornelski:
Sorry, I don't have time to be involved in the project.

zorgiepoo:
Ok thanks for response.

So two things here.

1. Create a PR for 2.x branch
2. Test legacy webview implementation

Then I'll want to just look up a bit more on `about:srcdoc` and implications in the codebase here.

zorgiepoo:
I am dealing this with in #2006 and #2007. Thanks for bringing this up. Fyi, tagged pre-releases are published to Cocoapods now (will get the chance to publish a pre-release shortly).

PR #1961, opened 2021-09-22
Fix to prevent Sparkle manipulating the host app's high level WebView defaults

danielpunkass:
This patch prevents Sparkle from manipulating the default WebView preferences, which at least on Monterey beta returns a preferences instance with nil identifier and with autosaves set to YES. This has the unfortunate effect of causing Sparkle, when it uses SULegacyWebView, to pollute the host app's user defaults with high level WebKit defaults for user style sheet, font sizes, etc. Specifying a unique identifier as well as turning autosaves off should both ensure Sparkle doesn't impose app-wide defaults on the host app, and also avoid littering the host app's preference file with identifier-prefixed defaults.

- [x ] My change is being tested and reviewed against the Sparkle 2.x branch. New changes must be developed on the 2.x development branch first.

## Type of change

- [x ] Bug fix (non-breaking change which fixes an issue)

## Testing

- [x ] My own app

macOS version tested: 12.0 (21A5522h)

### DIFF
--- a/Sparkle/SULegacyWebView.m
+++ b/Sparkle/SULegacyWebView.m
@@ -30,8 +30,9 @@
     self = [super init];
     if (self != nil) {
         _webView = [[WebView alloc] initWithFrame:NSZeroRect];
-        
-        WebPreferences *preferences = _webView.preferences;
+
+        WebPreferences *preferences = [[WebPreferences alloc] initWithIdentifier:@"sparkle-project.org.legacy-web-view"];
+        preferences.autosaves = NO;
         preferences.javaScriptEnabled = javaScriptEnabled;
         preferences.javaEnabled = NO;
         preferences.plugInsEnabled = NO;


zorgiepoo:
I accidentally regressed when adding the newer WebKit implementation / refactoring the old :/

Thank you for pinpointing/fixing this + the autosave change.

zorgiepoo (approved):

danielpunkass:
Thanks @zorgiepoo ! I'm looking into another pull request that would change the logic for relying on the legacy Web View so that the specific entitlements in question are checked. I think that might be a better long-term solution to restrict the legacy Web View behavior only to the precise apps that need the workaround.

zorgiepoo:
The current check sees if the downloader XPC Service is available. I think only sandboxed apps that don't use the outgoing networking entitlement should be using the XPC Service. An entitlement check could be an improvement, but it depends on the APIs used to achieve that.. I'm also as of just recently re-considering XPC Service integration in #1883

danielpunkass:
Oh, that would be nice if you could bundle the XPCs in the framework. Much nicer for integration. You're probably right that clients with the networking entitlement shouldn't use the downloader XPC service. I think I just got in the habit of using as many XPC services as possible ;P

zorgiepoo:
For reference, this bug affects:
* 2.x: when applications use the downloader XPC Service
* 1.x: when the user running operating system is < 10.11 (fix was backported to `master` as well)

PR #1963, opened 2021-09-23
Use entitlements check to determine whether to use legacy WebView

danielpunkass:
Hi again! I know you mentioned the unlikelihood of a host app having the Downloader XPC and not needing to use the legacy WebView, but since I got down the path of implementing a solution that checks more specifically for the problem scenario, I thought I'd share it in case you are interested.

This PR factors out the logic for determining whether to use a legacy WebView or not, so that it may be determined by the SUHost object. Default implementation bases the determination on the presence of entitlements on the host app specifying it is sandboxed but does not have networking client capabilities.

- [x ] I have reviewed and commented my code, particularly in hard-to-understand areas.

## Type of change

Fixes the edge case scenario where a host app wants to include the Downloader XPC but doesn't want to have to use the legacy WebView.

## Testing

- [x] My own app

I have confirmed the code works as expected in these scenarios:

- Host app has no entitlements
- Host app is sandboxed and has the networking entitlement set to YES
- Host app is sandboxed and has the networking entitlement set to NO
- Host app is sandboxed and has no networking entitlement

macOS version tested: 12.0 (21A5522h)

### DIFF
--- a/Sparkle/SUHost.h
+++ b/Sparkle/SUHost.h
@@ -27,6 +27,7 @@ NS_ASSUME_NONNULL_BEGIN
 @property (getter=isRunningOnReadOnlyVolume, readonly) BOOL runningOnReadOnlyVolume;
 @property (getter=isRunningTranslocated, readonly) BOOL runningTranslocated;
 @property (readonly, nonatomic, copy, nullable) NSString *publicDSAKeyFileKey;
+@property (readonly) BOOL requiresLegacyWebView;
 
 - (nullable id)objectForInfoDictionaryKey:(NSString *)key;
 - (BOOL)boolForInfoDictionaryKey:(NSString *)key;
--- a/Sparkle/SUHost.m
+++ b/Sparkle/SUHost.m
@@ -178,6 +178,37 @@ NS_ASSUME_NONNULL_BEGIN
     return [self objectForInfoDictionaryKey:SUPublicDSAKeyFileKey];
 }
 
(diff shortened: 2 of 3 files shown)


zorgiepoo:
Thanks for the PR.

This has more correct behavior (at least in the common case) such that an app that uses the XPC service (but does not need to be using it) will opt into WebKit2. However I am weary about using these Security APIs - I don't know what they do under the hood and what's their cost is; e.g, if they can block or depend on some system service, etc.

There aren't really good options here. I would much prefer the framework did not have sandbox checks and WebKit would just fix their bug (I would encourage filing duplicate feedback reports..)

Consider also an edge case like an application inheriting a sandbox from a parent process (like a screen saver plug-in for example) - is this code still correct, or if it is, does it come at a cost?

As an aside SUHost operations are for a specific bundle; the code is correct about caring about the 'currently running process' (I believe), but it's not associated with SUHost correctly.

If we make changes to how XPC Services are integrated and its integration guide on Sparkle's website we may also be able to reset some old assumptions about "integrating all the services"

danielpunkass:
Yep, that all makes sense. FWIW I think from auditing the Security framework that these checks are fast (essentially consuslting a CFDictionary internally), and I did bracket the check with dispatch_once to head off some possible performance impacts. But what you say about integrating the services alleviating the problem makes a lot of sense to. I mostly wanted to get this "on the books" in case the approach proves useful later. I'm satisfied if you want to close the PR.

zorgiepoo:
At least in the case of a process that inherits a sandbox environment from a parent process, this code is not correct and will not yield entitlement information. So we can't use it.

PR #1965, opened 2021-09-25
WIP: Move XPC Services inside Sparkle framework

### LINKED ISSUE #1883: Shipping pre-release builds of Sparkle 2.x via Swift Package Manager

zorgiepoo:
Move XPC Services inside Sparkle framework to simplify framework integration and SPM integration reported from #1883. System frameworks are examples of frameworks bundling XPC Services, so technically this should be possible..

This is an experimental change that needs more testing.

For some reason, when building the Sparkle Test App in Debug, the test app cannot create an XPC connection because launchd cannot find it - however the issue is mitigated when the app's path is changed to literally anything else (moving the app elsewhere or renaming the parent directory from Debug to something else). This issue does not reproduce when building/running it from Release, and I am not sure yet it will reproduce in configurations where people embed Sparkle as a subproject. It doesn't reproduce when copying the Sparkle.framework to another project and therefore theoretically shouldn't repro with SPM or cocoa pods.

I need to test an older operating system and Monterey beta as well.

This change allows configuring which XPC Services to bundle in the framework when building it by default, and Info.plist keys allowing developer to specify which services to enable, which would require a further website documentation update. By default the installer launcher and downloader services are bundled, and are not enabled.

The Installer launcher service with these changes do not bundle Autoupdate / Updater.app now and rather use the framework's copies of them, reducing the file size for developers that sandbox their app. However since framework is bundling the services, file size of framework by default is raised about 500 KB or so. 

TODO:
- [ ] Make change to prevent XPC Services from being used / bundled outside of the framework

Fixes #1883

## Checklist:

## Type of change

## Testing

- [x] Sparkle Test App
- [ ] Unit Tests
- [x] My own app
- [ ] Other (please specify)

- [x] Tested Sparkle Test App (after building in Release or moving the Debug built app elsewhere)
- [x] Tested Sparkle Test App using downloader XPC Service (with some temporary changes not shown in this PR)
- [ ] Investigate Test App + Debug + Installed location (derived data / products / debug) issue + report issue
- [ ] Testing CI now.
- [ ] Need to test older operating system
- [ ] Need to test Monterey beta
- [ ] Reviewal from developers integrating Sparkle

macOS version tested: 11.6 (20G165)


zorgiepoo:
Sample nightly build https://github.com/sparkle-project/Sparkle/actions/runs/1273926947

zorgiepoo:
@danielpunkass @1024jp @tonyarnold

zorgiepoo:
I identified the system cache issue I was encountering earlier and put in a workaround. Essentially when an old version (before this PR) of Sparkle launches an XPC Service located directly inside the app bundle's Contents/XPCServices, the system caches in some way or notices it's not spawning XPC Services located inside the framework, and the system uses the file path of Sparkle as a key. When a newer version of an app replaces the old version at the same path, this cache is not invalidated. So when the newer version of the app uses a version of Sparkle that embeds the XPC Services inside the framework bundle, the system fails to find/launch those XPC Services. Which is why workarounds were: a. moving the app to another location or b. rebooting the system. As a more effective workaround, I'm changing the install path of Sparkle by changing the FRAMEWORK_VERSION so this type of cache conflict does not happen when upgrading to a newer version of Sparkle (https://github.com/sparkle-project/Sparkle/pull/1965/commits/e48302516c673c65fa479423531b7523db794c77)

zorgiepoo:
Latest artifact build https://github.com/sparkle-project/Sparkle/suites/3940556327/artifacts/98770944

Let me know if anyone can give this a try.

tonyarnold:
The code looks good to me, @zorgiepoo, however I'm not going to have a chance to test this until next week at the earliest. I'll be extremely excited if bundling the XPC services into the framework actually works!

digitalmoksha:
I was finally able to test this out.  Works like a charm 🚀 

However, since I don't use XCode directly, I ended up needing to re-sign the entire package, not just the XPC services, as you talk about in Code Signing (https://sparkle-project.org/documentation/sandboxing/).  

I kept getting errors like this during Apple notarization

```json
    {
      "severity": "error",
      "code": null,
      "path": "Versatil_Markdown.zip/Versatil Markdown.app/Contents/Frameworks/Sparkle.framework/Versions/B/Autoupdate",
      "message": "The executable does not have the hardened runtime enabled.",
      "docUrl": null,
      "architecture": "x86_64"
    },
```

so I used `--deep` and signed the entire package, like this:

```
codesign --deep -f -s "$CODE_SIGN_IDENTITY" -o runtime --preserve-metadata=entitlements Sparkle.framework
```

This may be atypical to my scenario, but wanted to at least mention it here.

zorgiepoo:
That doesn't make sense. We have hardened runtime enabled in our project and for that executable. You can even verify the runtime option is set yourself from the `Sparkle-2.0.0-beta.4.tar.xz` package -- see (adhoc,**runtime**)

```
msp@Sleek ~> codesign -dvvv /Users/msp/Downloads/Sparkle-2.0.0-beta.4\ 2/Sparkle.framework/Versions/B/Autoupdate 
Executable=/Users/msp/Downloads/Sparkle-2.0.0-beta.4 2/Sparkle.framework/Versions/B/Autoupdate
Identifier=Autoupdate-5555494417663fb24f963c489bd6f5503232f70b
Format=Mach-O universal (x86_64 arm64)
CodeDirectory v=20500 size=3476 flags=0x10002(adhoc,runtime) hashes=102+2 location=embedded
```

zorgiepoo:
However you will need to re-sign it and any other auxiliary tools / services anyway because it's signed with an adhoc signature, which does make sense. Xcode's standard workflow does this already.

I *may* want to add get-task-allow entitlements to the XPC Services for development purposes (I haven't decided yet).. in which case `--preserve-metadata=entitlements` would not be what you'd want. So keep an eye out for that.

digitalmoksha:
Yeah, they have the `runtime` flag set - I was surprised at it too. But it wasn't until I tried to notarize the app that it showed up.

>  in which case --preserve-metadata=entitlements would not be what you'd want

ah, thanks for the heads up.  I only added that because you had it in your sample code.  If it's not necessary, then I may just go ahead and remove it.

zorgiepoo:
You need to preserve entitlements for the downloader XPC Service (if you use it) currently. Everything else currently doesn't have entitlements / isn't sandboxed and doesn't matter. If I make a change to add task-get-allow entitlements to XPC Services we distribute, that then means you don't want to preserve the task-get-allow entitlement specifically (but want to preserve sandboxing entitlement for downloader service still).

Generally because different executables can have different entitlements is why --deep signing isn't all that encouraged, although it works when you want to apply everything the same way, and not using --deep means you have to sign things inside-out order like Xcode does.

zorgiepoo:
I updated the [sandbox guide](https://sparkle-project.org/documentation/sandboxing) for code signing with custom non-xcode workflows for the latest beta. 

For the standard / common workflow, you probably don't need to disable debugging XPC Services anymore and can test Sparkle in a live debugging session.

Issue #1986, opened 2021-10-28
Clients polling rapidly - began ~Oct 26...


herf:
I am seeing a big uptick in clients polling our Appcast in a loop (growing rapidly over the last three days, beginning around Oct 26). A single client can generate 500,000 requests per day (we drop these clients at the firewall). Right now, we are blocking hundreds of IPs. My shipping version is Sparkle/1.24.

The symptoms:
- About 4% of clients have a user agent of "Sparkle/?" implying "displayVersion" is failing somehow.
- Previously I had about 1.8% of clients with "Sparkle/?", but they did not poll, they just acted like normal clients. Now there are more of them and they poll.
- Darwin versions for the high-traffic clients include 20.6.0, 20.5.0, 20.3.0, and 21.1.0 (so Big Sur and Monterey) but older versions still sometimes send "Sparkle/?" as UA.

These "Sparkle/?" requests seem to be using separate cookies than the ones that were sending a version, so it's like they're suddenly getting sandboxed out of the filesystem or something like that. Also, some of the clients sending system configuration (the random sample) are sending "Sparkle/?" so I think it's less likely that a third party (like Homebrew cask) is at fault, because the system information looks an awful lot like Sparkle.

I've tried running my app from Downloads/ while denying it access to Downloads, but Sparkle just popped up a nice error. I'm thinking something like this security-wise has changed, but looking for any clues (or maybe bugs that are fixed in newer versions!)

Thanks for any help.

A much older version of this bug was #1061

zorgiepoo:
#1992 and #1991 for that. I still don't have an idea of underlying issue but it could help.

herf:
When I "slowloris" the updates, I am still seeing multiple connections, overlapping.

I think the paranoid approach is to assume something in macOS security has totally disconnected the filesystem - so NSDate and NSUserDefaults cannot be touched in the worst-case "retry" path. Can we force delays that don't depend on the filesystem?

zorgiepoo:
We already don't trust NSUserDefaults in the re-try path. I just pushed changes in to not trust NSDate in the common re-try-loop path too (#1991).

We still need to rely on a timer of some sort to schedule the next update. The alternative to that is you can disable automatic checks for updates (`SUEnableAutomaticChecks` to `NO`) but then you'd have to be in charge for when to make update checks in the background (`checkForUpdatesInBackground`). Or you can decide to start the updater itself at some point later if you think it will help you (if it's an early initialization problem).

[...]

PR #2000, opened 2021-11-01
Prevent potential rapid update checks from sleep/wake events

zorgiepoo:
Prevent potential rapid update checks from sleep/wake events. This is a "paranoid" check.

Related: #1986

I may have to port this to 2.x which means I may have to listen for sleep/wake notifications there (which sucks to begin with), or it could just mean I adjust the dispatch source leeway interval 🤔. (If the suspected problem is too many successive wake notifications, then no need to port. If the suspected issue is system being in bad state after wake, then technically there should be a similar problem after system boots and we aren't really solving anything here)

This is only useful if it's believed that system wake notifications are often successive (or in a bad networking state) AND date checking code is unreliable. 

I am not convinced that this change should be merged yet (and 2.x would need appropriate change IF I believe this is an actual problem, otherwise this won't be merged)

## Checklist:

- [ ] My change is being tested and reviewed against the Sparkle 2.x branch. New changes must be developed on the 2.x development branch first.
- [x] My change is being backported to master branch (Sparkle 1.x). Please create a separate pull request for 1.x, should it be backported. Note 1.x is feature frozen and is only taking bug fixes, localization updates, and critical OS adoption enhancements.

## Type of change

## Testing

- [x] Sparkle Test App
- [ ] Unit Tests
- [ ] My own app
- [ ] Other (please specify)

Tested sleep / wake while monotonic time interval has not fully passed yet, but wall time has passed. In that case, leftover delay from monotonic time takes effect.

Tested sleep / wake while monotonic time interval has fully passed, but wall time has not passed. In that case leftover delay from wall time takes effect.

Tested sleep / wake while monotonic time interval has fully passed, and wall time has fully passed. In that case, there's 0 delay.

Tested when app starts updater on launch, no minimum delay from monotonic clock is enforced.

macOS version tested: 12.0.1 (21A559)

### DIFF
--- a/Sparkle/SUConstants.h
+++ b/Sparkle/SUConstants.h
@@ -18,6 +18,7 @@
 
 extern const NSTimeInterval SUMinimumUpdateCheckInterval;
 extern const NSTimeInterval SUDefaultUpdateCheckInterval;
+extern const NSTimeInterval SUMinimumMonotonicUpdateCheckInterval;
 
 extern NSString *const SUBundleIdentifier;
 
--- a/Sparkle/SUConstants.m
+++ b/Sparkle/SUConstants.m
@@ -18,6 +18,7 @@
 // Define some minimum intervals to avoid DoS-like checking attacks
 const NSTimeInterval SUMinimumUpdateCheckInterval = DEBUG ? 60 : (60 * 60);
 const NSTimeInterval SUDefaultUpdateCheckInterval = DEBUG ? 60 : (60 * 60 * 24);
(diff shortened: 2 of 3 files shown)


zorgiepoo:
This causes a delay after normal wake of when the prompt appears. I don't think it should be done like this. Rather there should be a cooldown preventing successive wakes in a short interval.

zorgiepoo:
I updated this to enforce a minimum elapsed monotonic delay.

Issue #857, opened 2016-08-03
Inflexible API


danielchasehooper:
Sparkle is great if all you want is the vanilla update process, but even slight changes to it's behavior are difficult to impossible.

While trying to use sparkle to add downgrade to support to Principleformac.com I made some notes:
- All or nothing API. For the customization that is there, you either have to use it as is, or completely roll your own, even the parts that you aren't changing. Most of the bullet points below are  incarnations of this problem.
  For example:  Implementing bestValidUpdateInAppcast:forUpdater: requires you reimplement the max/min supported OS logic in addition to your novel verification logic. It would be nice if this logic were either already applied to the items that get sent to this delegate method and/or if SUAppcastItems had a -supportsOSVersion method
- Doesn't allow extension without building from source
  If someone wanted to build their own update driver, they'd most likely have to end up building the framework from source, or do some nasty runtime hacking. It'd be nice if more headers were visible, even if just in a "Optional Headers" folder in the framework so that it's there for the people that need it. In my case, I'm wanting to use the basicUpdateDriver since it doesn't show any of it's own UI, but since the framework doesn't expose that class, so I'm getting the class by name at runtime - I don't want to add all Sparkle source to our project/build time.
- Almost all SUUpdater delegate methods give you no control and act only as notifications. There have been lots of instances where the driver did mostly what I wanted, but I wanted to choose to show my own UI, show the alerts as a sheet (since they don't work when there is already a modal window), prevent the UI from showing all together. Example: -updater: shouldShowAlert: OfType:, which would allow you to control over how/if alerts are shown. It'd be nice if the delegate was delegated-to more.
- huge number of delegate callback paths
  If you have logic that you need to guarantee runs after the update process has run you have to account for all the edge cases: user cancels, valid update couldn't be found, couldn't download update, etc. Looking at the source, I'm not even sure all paths are accounted for even if you implement all delegate methods. Some sort of will/didRunUpdateProcess might be helpful. 
- orthogonal concepts combined
  SUUpdateDriver combines all the logic relating to how the update is checked, installed and relaunched and it is also in charge of the user-facing aspect. There isn't a way to have one without the other.  It may be worth considering splitting the two concepts into a SUUpdateDriver & SUUpdateUIProvider, for example - composition-based drivers would do this better than a class hierarchy.

zorgiepoo:
[...]

PR #2005, opened 2021-11-06
Use NSPanel for Sparkle windows to allow untitled windows form Dock

GetToSet:
## Background

It takes time to download updates and users tend to continue using an app while it's downloading updates.

Document based apps often override `-[NSApplicationDelegate applicationShouldOpenUntitledFile:]`` & `-[NSApplicationDelegate applicationOpenUntitledFile:]` to allow opening new document when the Dock icon being clicked when no window exists.

An interesting behavior that I discover is that while Sparkle has its window on screen (e.g. downloading updates), AppKit no longer calls `-[NSApplicationDelegate applicationOpenUntitledFile:]`. This behavior especially bothers users for apps such as [iTerm2](https://iterm2.com/), since closing an unneeded pseudo terminal and reopen a new one from Dock is an extremely frequent interaction. Users have to right-click or control-click the Dock icon to achieve the desired behavior.

This PR uses NSPanel (rather than `NSWindow`) for Sparkle windows  to allow untitled windows form Dock.

## Deeper Dive

Here is a typical call stack that AppKit handles opening untitled file:

Screenshot 2021-11-07 上午2 09 25

Disassembling AppKit binary (12.0.1, 21A559) and looking for symbol `-[NSApplication(NSAppleEventHandling) _handleAEReopen:] signifies AppKit prevents `-[NSApplication _doOpenUntitled]` call when `-[NSApplication _appHasOpenNSWindow]` is `YES`:

Screenshot 2021-11-07 上午2 14 57

For its implementation (notice the third argument differs):

Screenshot 2021-11-07 上午2 22 16

Screenshot 2021-11-07 上午2 22 25

`-[NSApplication _appHasOpenNSWindow]` has a dedicated check that excludes `NSPanel` class:

Screenshot 2021-11-07 上午2 26 18

and here comes this PR.

## Checklist:

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Testing

- [x] Sparkle Test App
- [x] Unit Tests
- [ ] My own app
- [x] Other (please specify)
  - [x] [CotEditor](https://github.com/coteditor/CotEditor)

1. Make sure target app implements `-[NSApplicationDelegate applicationShouldOpenUntitledFile:]`
2. Close all windows except any Sparkle windows, click the Dock icon.
3. `-[NSApplicationDelegate applicationShouldOpenUntitledFile:]` should be called.

macOS version tested: All latest major versions from 10.11-12.0 have been tested with the Test App. A version of [CotEditor](https://github.com/coteditor/CotEditor) using the branch version of Sparkle has been tested on all latest major versions from 10.15-12.0

### DIFF
--- a/Sparkle/Base.lproj/SUUpdateAlert.xib
+++ b/Sparkle/Base.lproj/SUUpdateAlert.xib
@@ -20,7 +20,7 @@
         </customObject>
         <customObject id="-1" userLabel="First Responder" customClass="FirstResponder"/>
         <customObject id="-3" userLabel="Application" customClass="NSObject"/>
-        <window identifier="SUUpdateAlert" title="Software Update" allowsToolTipsWhenApplicationIsInactive="NO" autorecalculatesKeyViewLoop="NO" releasedWhenClosed="NO" visibleAtLaunch="NO" frameAutosaveName="SUUpdateAlert" animationBehavior="default" id="5" userLabel="Update Alert (release notes)">
+        <window identifier="SUUpdateAlert" title="Software Update" allowsToolTipsWhenApplicationIsInactive="NO" autorecalculatesKeyViewLoop="NO" releasedWhenClosed="NO" visibleAtLaunch="NO" frameAutosaveName="SUUpdateAlert" animationBehavior="default" id="5" userLabel="Update Alert (release notes)" customClass="NSPanel">
             <windowStyleMask key="styleMask" titled="YES" closable="YES" miniaturizable="YES" resizable="YES"/>
             <windowCollectionBehavior key="collectionBehavior" fullScreenAuxiliary="YES"/>
             <windowPositionMask key="initialPositionMask" topStrut="YES" bottomStrut="YES"/>
--- a/Sparkle/Base.lproj/SUUpdatePermissionPrompt.xib
+++ b/Sparkle/Base.lproj/SUUpdatePermissionPrompt.xib
@@ -23,7 +23,7 @@
         </customObject>
         <customObject id="-1" userLabel="First Responder" customClass="FirstResponder"/>
         <customObject id="-3" userLabel="Application" customClass="NSObject"/>
-        <window allowsToolTipsWhenApplicationIsInactive="NO" autorecalculatesKeyViewLoop="NO" releasedWhenClosed="NO" visibleAtLaunch="NO" animationBehavior="default" id="5" userLabel="Profile Info">
+        <window allowsToolTipsWhenApplicationIsInactive="NO" autorecalculatesKeyViewLoop="NO" releasedWhenClosed="NO" visibleAtLaunch="NO" animationBehavior="default" id="5" userLabel="Profile Info" customClass="NSPanel">
             <windowStyleMask key="styleMask" titled="YES"/>
             <windowCollectionBehavior key="collectionBehavior" fullScreenAuxiliary="YES"/>
             <windowPositionMask key="initialPositionMask" leftStrut="YES" rightStrut="YES" topStrut="YES" bottomStrut="YES"/>
--- a/Sparkle/SUStatus.xib
+++ b/Sparkle/SUStatus.xib
@@ -16,7 +16,7 @@
         </customObject>
         <customObject id="-1" userLabel="First Responder" customClass="FirstResponder"/>
         <customObject id="-3" userLabel="Application" customClass="NSObject"/>
-        <window identifier="SUStatus" title="Set in Code" allowsToolTipsWhenApplicationIsInactive="NO" autorecalculatesKeyViewLoop="NO" releasedWhenClosed="NO" visibleAtLaunch="NO" animationBehavior="default" id="5" userLabel="Window">
+        <window identifier="SUStatus" title="Set in Code" allowsToolTipsWhenApplicationIsInactive="NO" autorecalculatesKeyViewLoop="NO" releasedWhenClosed="NO" visibleAtLaunch="NO" animationBehavior="default" id="5" userLabel="Window" customClass="NSPanel">
             <windowStyleMask key="styleMask" titled="YES"/>
             <windowPositionMask key="initialPositionMask" leftStrut="YES" rightStrut="YES" topStrut="YES" bottomStrut="YES"/>
             <rect key="contentRect" x="200" y="222" width="400" height="107"/>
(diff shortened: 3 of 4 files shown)


zorgiepoo:
Thank you for the PR and in depth analysis and testing. It's probably good. I assume this change to NSPanel does not make any changes in behavior or activation state when it's in the background.

Please update the UI tests in `SUTestApplicationTest.swift` as they're currently failing. You'll want to use `app.dialogs` instead of `app.windows` in these two lines:

```swift
app.windows["SUUpdateAlert"].buttons["Install Update"].click()
/*...*/
app.windows["SUStatus"].buttons["Install and Relaunch"].click()
```

zorgiepoo:
This change will also make this work correctly:

```objc
- (BOOL)applicationShouldTerminateAfterLastWindowClosed:(NSApplication *)sender
{
    return YES;
}
```

As panels don't appear to count here, so Sparkle's status or update windows won't prevent the application from terminating. I vaguely remember maybe iTerm running into these sort of issues (cc @gnachman) but can't find the exact issue.

aonez:
> I was actually thinking of Keka @aonez

@zorgiepoo my issue was just [the inverse](https://github.com/aonez/Keka/issues/408#issuecomment-766871807), the app terminated before being updated.

GetToSet:
> This change will also make this work correctly:
> 
> ```objc
> - (BOOL)applicationShouldTerminateAfterLastWindowClosed:(NSApplication *)sender
> {
>     return YES;
> }
> ```
> 
> As panels don't appear to count here, so Sparkle's status or update windows won't prevent the application from terminating. I vaguely remember maybe iTerm running into these sort of issues (cc @gnachman) but can't find the exact issue. (edit: my bad, I was actually thinking of Keka @aonez)
> 
> Will likely cherry-pick back to 1.x if I don't see issues.

Sorry that I didn't even notice there's a UI Test.

Using`NSPanel` can result a different behavior when apps implementing `-[NSApplicationDelegate applicationShouldTerminateAfterLastWindowClosed:]` to return `YES` and may introduce other side-effects that require developers' attention.

It might be better to let developers to have more real-world testing on beta branch, rather than cherry-picking it to 1.x.

zorgiepoo (approved):

zorgiepoo:
I'll put it in a beta tag for 1.x for a short while. I generally like back porting bug-fixes for the time being. The default behavior previously was that an app would not close if a Sparkle panel appeared and other windows were closed by the user but when you tried to install an update, the app would quit.

zorgiepoo:
I took more time to play with these changes but I think I prematurely merged this and came to some wrong conclusions here. Sorry.

> The default behavior previously was that an app would not quit if a Sparkle panel appeared and other windows were closed by the user (if they opted into this delegate without considering Sparkle) but when you tried to install an update, the app would quit

This is wrong according to my recent tests (on macOS 12.0.1 (21A559)). When I try to implement `-applicationShouldTerminateAfterLastWindowClosed:` and always return `YES` and I only have a Sparkle update alert window open prior to these changes, if I try to install an update, it ends up working and the app doesn't get terminated unexpectedly in between. Which I think is actually the desired behavior. Now with these changes if the status window or update alert window is a panel and shown onscreen, and closing other windows terminates the application, that is *unexpected*.

To the original issue: I think clicking the app in the dock should bring up the window that was last in key focus. If Sparkle has an update alert window or status window shown (often there is some user initiation here), clicking the dock icon *should* bring up Sparkle's window if it was the one in key focus. This is the case when all other document windows are closed. And this is what I expect when I want to eg switch back to iTerm - I don't expect an untitled window should open if Sparkle's status or alert window is on screen. And a developer may want to prioritize the user to give attention to updating the app.. The dock is about bringing the app back more-so than opening new documents (if you're already in the app you can open a new document from menu bar or command n or similar).

So it's a bit more complicated than what I thought. Likely I am going to revert these changes but we can continue discussion here as needed.

GetToSet:
> I took more time to play with these changes but I think I prematurely merged this and came to some wrong conclusions here. Sorry.
> 
> > The default behavior previously was that an app would not quit if a Sparkle panel appeared and other windows were closed by the user (if they opted into this delegate without considering Sparkle) but when you tried to install an update, the app would quit
> 
> This is wrong according to my recent tests (on macOS 12.0.1 (21A559)). When I try to implement `-applicationShouldTerminateAfterLastWindowClosed:` and always return `YES` and I only have a Sparkle update alert window open prior to these changes, if I try to install an update, it ends up working and the app doesn't get terminated unexpectedly in between. Which I think is actually the desired behavior. Now with these changes if the status window or update alert window is a panel and shown onscreen, and closing other windows terminates the application (without consideration of Sparkle), that is _unexpected_.
> 
> To the original issue: I think clicking the app in the dock should bring up the window that was last in key focus. If Sparkle has an update alert window or status window shown (often there is some user initiation here), clicking the dock icon _should_ bring up Sparkle's window if it was the one in key focus. This is the case when all other document windows are closed. And this is what I expect when I want to eg switch back to iTerm - I don't expect an untitled window should open if Sparkle's status or alert window is on screen. And a developer may want to prioritize the user to give attention to updating the app.. The dock is about bringing the app back more-so than opening new documents (if you're already in the app you can open a new document from menu bar or command n or similar). I don't think Sparkle is "auxiliary" enough to have its windows be panels.
> 
> So it's a bit more complicated than what I thought. Likely I am going to revert these changes but we can continue discussion here as needed.

Yeah it seems more complicated than I thought. Maybe there can be a less intrusive alternative to the original issue, or 
 making it possible for developers opt-in this behavior. I'll continue to investigate for a better approach.

aonez:
> making it possible for developers opt-in this behavior

This can be a good option. For Keka is use a combination of those for opening a new window, so I really don't care if there're other windows opened:

```objc
- (BOOL)applicationShouldHandleReopen:(NSApplication *)theApplication hasVisibleWindows:(BOOL)flag;
- (void)applicationDidBecomeActive:(NSNotification *)aNotification;
- (void)applicationDidResignActive:(NSNotification *)aNotification;
```

zorgiepoo:
The first one of those delegate methods sounds like it may give the app control of what to do, maybe if it detects Sparkle `sessionInProgress && canCheckForUpdates` (a checkable session is in progress) or so...

I don't think there's much to do here from Sparkle's end as I think its default behavior behaves expectably. NSApplication delegate customization is outside the realm of Sparkle. Another approach is to make a custom UI that more inlines the update process to a single window (eg launch the Sparkle Test App by holding down Shift, or check what Nova does https://github.com/sparkle-project/Sparkle/issues/857#issuecomment-808015267)

PR #2021, opened 2021-11-13
Preserve custom icons for the app during the update

core-code:
Some users like to set custom icons for apps (using Finder's Get Info window).

Obviously these custom icons vanish when the app updates itself using Sparkle.

This patch tries to preserve the custom icons during the update.

### DIFF
--- a/Autoupdate/SUPlainInstaller.m
+++ b/Autoupdate/SUPlainInstaller.m
@@ -155,6 +155,15 @@
         [newTempURL setResourceValue:resourceTags forKey:NSURLTagNamesKey error:NULL];
     }
     
+    // Try to preserve custom icons set by the user in the Finder
+    if ([[oldTempURL URLByAppendingPathComponent:@"Icon\r"] checkResourceIsReachableAndReturnError:nil])
+    {
+        NSString *oldTempPath = oldTempURL.path;
+        NSString *destinationAppPath = destinationAppURL.path;
+        NSImage *customIcon = [NSWorkspace.sharedWorkspace iconForFile:oldTempPath];
+        [NSWorkspace.sharedWorkspace setIcon:customIcon forFile:destinationAppPath options:NSExcludeQuickDrawElementsIconCreationOption]; // ignore return status - failure ain't fatal
+    }
+    
     // We must leave moving the app to its destination as the final step in installing it, so that
     // it's not possible our new app can be left in an incomplete state at the final destination
     


zorgiepoo:
I don't think I want to support preserving these icons overridden in Finder (which may be low-resolution and possibly not scale). We also cannot use AppKit in the installer code which may be executed from a daemon. I assume these calls are not expensive but it would also be something I'd be interested in profiling if I were interested in preserving these..

core-code:
> I don't think I want to support preserving these icons overridden in Finder (which may be low-resolution and possibly not scale). 

im not sure i understand you here. a custom icon will have the quality that the user has created for the icon. in any case, it is the icon the user wants to see for the app. it shouldn't vanish just because the app is updated.

> We also cannot use AppKit in the installer code which may be executed from a daemon. 

hm. an alternative approach would be copying the custom icon and setting the required xattr manually.

> I assume these calls are not expensive but it would also be something I'd be interested in profiling if I were interested in preserving these.

not sure about this. even if the calls were expensive they'd not be noticeable compared to the time required to download and apply an update. increasing the time here by a millionth part shouldn't matter.

furthermore and even more important, no time will be spent at all unless the user has actually applied a custom icon. so the code path will only be applied for like one in 1000 - 10000 users.

> When a developer updates their app icon we'd also be making a choice that might not be right here.

thats a valid concern. i believe if a user applies a custom icon, he will still want to keep his custom icon in most cases even if the app ships with a new icon. but it might not be the case for everyone and under all circumstances.

what would the solution here be? checking if the icon has changed in the app might be possible but not trivial

zorgiepoo:
> im not sure i understand you here. a custom icon will have the quality that the user has created for the icon. in any case, it is the icon the user wants to see for the app. it shouldn't vanish just because the app is updated.

A regular app icon at least scales and may change depending on the size or resolution it is shown in. But icons set this way *might* be set just for a specific size. I'm not really that sure on the details and limitations of this format to be honest. My impression is that this is not as good quality as an app icon or an app that supports custom dock icons.

> not sure about this. even if the calls were expensive they'd not be noticeable compared to the time required to download and apply an update. increasing the time here by a millionth part shouldn't matter.

Performance wise the downloading and extraction time doesn't matter here if this code is in the installation-path (when you try to install/relaunch the app). A long period of idleness can go between downloading/extracting and installing. I have seen a noticeable bottleneck before in another program I worked on trying to retrieve application icons from bundles. You are right this is guarded by an uncommon use case ("Icon\r"), although checking if the icon changed would make this worse.

The App Store does not preserve custom icons set this way (and thus presumably neither does `-[NSFileManager replaceItemAtURL:withItemAtURL:backupItemName:options:resultingItemURL:error:]` but I have not confirmed this), so there is a precedent here of it being standard not to preserve this. They do preserve labels and comments though as you'd expect.

core-code:
>  I'm not really that sure on the details and limitations of this format to be honest. 

the implementation is very weird, as the icon is stored in the resource fork (!) of that empty `Icon\r` file. definitely a classic MacOS < 10 leftover.

> But icons set this way might be set just for a specific size. 

well, it depends on which resolution the icon the user applies has. it was half a megabyte in my test case.

> My impression is that this is not as good quality as an app icon or an app that supports custom dock icons 

i don't think thats relevant. if the user wants to see a specific icon and goes to the trouble of applying the icon, thats his choice. saying 'you are not allowed to see the icon you want to see because we believe it might be low-quality' isn't friendly.

> The App Store does not preserve custom icons set this way (and thus presumably neither does -[NSFileManager replaceItemAtURL:withItemAtURL:backupItemName:options:resultingItemURL:error:] but I have not confirmed this), so there is a precedent here of it being standard not to preserve this.

definitely. though on the one hand, just because something is done in a way, doesn't mean it is good. and on the other hand, the Mac App Store is ultra-low quality and really quite user-hostile in many ways. 

in any case, i'd never apply a custom icon myself, and i struggle to understand users that do this. i'd be happy to close this and only re-open this bug-report if enough actual users that would like to see some progress here speak up.

zorgiepoo:
Yeah let's defer this then. My long interest is adopting the replaceItem: method one day (after I verify it's safe to do so) and having less one-off cases, appkit usage, or hardcoded/private attribute usage.

PR #2023, opened 2021-11-15
Fix unarchiving job being blocked when passed bad input

### LINKED ISSUE #1433: 'SUPipedUnarchiver' can go into infinite loop for some ZIP files, causing app to freeze.
for easy reproduction i've taken the code from 'SUPipedUnarchiver' and embedded into a standalone app so you can quickly confirm this issue.

1.) download the attached "ZIPUncompresser" xcode project which uses SUPipedUnarchiver to uncompress ZIP files
2.) build and run the app using Xcode
3.) drag any ZIP file onto ZIPUncompresser, everything works fine
4.) download this zip file: https://updates.signal.org/desktop/signal-desktop-mac-1.25.3.zip
5.) drop it onto ZIPUncompresser
6.) SUPipedUnarchiver is stuck in an infinite loop in [archiveOutput writeData:data];

this problem will presumably send any app into a infinite loop that tries to update using ZIP files that are similar to 'signal-desktop-mac-1.25.3.zip'

bug happens on 10.14.5 and presumably other macOS versions

what is the rationale to use a pipe instead of directly pointing 'ditto' at the 'archivePath' anyway - which would presumably also be much faster?

zorgiepoo:
Fix unarchiving job being blocked when passed bad input.

We were passing `[pipe fileHandleForReading]` to `standardInput` of the task but according to the API documentation, this is wrong because if you do this, the pipe will not be automatically closed. So we have to pass the pipe itself. The reason we use a pipe is to propagate unarchive file progress back to the application (I did do a quick test with a 300 MB zip to verify the file progress *is* actually meaningful over time, and it's not any slower than not piping).

One complication is when bad input is passed, the calling process can receive a SIGPIPE signal when writing to the pipe which by default will terminate the process. In Autoupdate and in the (bad case) unit tests we ignore SIGPIPE to prevent them from being terminated suddenly (I found no other way around this, even with setting writeabilityHandler and trying other APIs..). This also means this change is not as easy to correctly port to 1.x because unarchiving is not done out of process there (and would therefore have a side effect affecting the developer's running application).

I also adopted the newer NSTask / NSFileHandle methods that take an NSError object instead of raising exceptions.

Fixes #1433

## Checklist:

## Type of change

## Testing

- [x] Sparkle Test App
- [x] Unit Tests
- [x] My own app
- [ ] Other (please specify)

Tested following changes by:

Created a unit test with extraneous bytes at end of file.
Created a unit test with bad header (missing proper zip header).
Tested both of the above cases in a real sample app.
Older macOS version (with both bad zip cases, and good zip case)
Pkg package installation (via sparkle-cli)
Tested installing large-ish regular zip update (300 MB)
Tested app dmg installation

macOS version tested:
12.0.1 (21A559)
10.12.6 (16G29) VM


core-code:
just as documentation and FYI since this is slightly connected: we've used SUDiskImageUnarchiver code in a project but had some weird issues which occurred for around one in a thousand users. namely it would just freeze on DMG decompression, inside Sparkle's code, specifically the `readDataToEndOfFile`.

we've spend a TSI working with Apple on getting that resolved. the problem seemed to have been a general issue in NSTask that did occur only for some rare users in practice. in any case the resolution was getting rid of `readDataToEndOfFile` and `availableData` and instead relying on `setReadabilityHandler` `setTerminationHandler` and a semaphore. if you want the source to our modifed SUDiskImageUnarchiver version, happy to share.

zorgiepoo:
Yeah that would be helpful to share your implementation, by the sounds of it I'm sure our logic is faulty.

core-code:
> that would be helpful to share your implementation,

sure thing, its here:
https://gist.github.com/core-code/2002790687de4def4c6ebee7db75c195

> by the sounds of it I'm sure our logic is faulty.

actually no. it seems like any app calling `readDataToEndOfFile` could be stuck hanging forever, but this will almost never occur in practice. which made this an awful issue to resolve, costing us weeks. here is the conclusion of our discussion with an Apple engineer:

> > the way I read your
response is that any invocation of NSTask / readDataToEndOfFile might
deadlock due to the buffer issue and its basically a gamble whether it
> > will ever return into our code depending on the amount and timing of
the text output written by the called process.  

> Yes, that's basically correct, though the details also depend on exactly
how a specific tools is implemented.

zorgiepoo:
Thanks for that snipped, I'll deal with that issue in #2027

PR #2048, opened 2021-12-30
Update pull request templates and Xcode CI version to 13.2.1

zorgiepoo:


zorgiepoo:
Warnings should be fixed by #2047 I believe so I'll need to re-run this.

PR #2050, opened 2022-01-01
Fixed #2049: Check http statusCode in didFinishDownloadingToURL

### LINKED ISSUE #2049: Check http status code on feed downloads for a more precise error message
## Summary

Sparkle 2.0.0 is displaying **"An error occurred while parsing the update feed"** when the problem is actually a permission error from the web server. Checking the http statusCode would allow for a better error message. 

Details:
Our feed is protected by CloudFront. If the user's access token expires, CloudFront returns a 503 error along with an HTML page explaining the issue. Sparkle's NSURLSessionDownloadTask delegate doesn't check the http status code on the download and attempts to parse the HTML error page from CloudFront. This results in the slightly erroneous but confusing error message above. 

## Possible Fix

I believe this can be fixed in SPUDownloader.m: didFinishDownloadingToURL by checking the statusCode and using a new error message if it indicates an error. A (Swift - sorry) example on checking the statusCode is shown below:

```
    func urlSession(_ session: URLSession, 
      downloadTask: URLSessionDownloadTask, 
      didFinishDownloadingTo location: URL) {

      if let response = downloadTask.response as? HTTPURLResponse {
        if response.statusCode < 200 || response.statusCode >= 400 {
          error = "Status code \(response.statusCode)"
          return
        }
      }
      // download was successful
    }

```
An improved message might be something as simple as "A network error occurred while downloading the update feed (503)."

SparkleUpdateError

## Version

2.0.0

EricShapiro:
Check the http statusCode in didFinishDownloadingToURL to ensure that errors are detected on download rather than ignored and trying to parse the error description later. 

Fixes # (issue)
2049

## Testing

- [X ] My own app

I used a URL protected by CloudFront with an invalid token. In this case, CloudFront returns a 503 error with an HTML page description the issue. 

Note that the error message I added is correct and I included the statusCode, but that error doesn't get propagated to the user. I don't consider that a big issue. Instead the more generic error "An error occurred while downloading the update. Please try again later." is displayed, which is fine.

12.1


zorgiepoo:
This logic looks good barring a couple minor comments.

zorgiepoo:
I think you have some incidental changes here now (the GitHub pull request workflow is such that a separate branch is used for pull requests, and that 2.x remains the same as upstream).

zorgiepoo:
I handled this in #2073 (just made minor adjustments). Thanks.

I was looking through your commit history btw, if you want an update check to be done on every launch you can call -checkForUpdatesInBackground on app launch.

Issue #1902, opened 2021-07-23
Upgrading binary delta support


zorgiepoo:
From #1899, it is noted xar is deprecated, and our current delta implementation is lacking.

One suggestion is:

> I think we could drop xar and bsdiff and replace them with zstd and some basic container format.
> zstd, besides having decent compression and great speed, has an option to specify a "prefix" for compression/decompression. This prefix can be set to be the old file version, and then it automatically becomes an optimal file diff. (Prefix is a one-time dictionary. zstd also has an option to generate reusable dictionaries, but AFAIK this isn't suitable for our case, since the dictionaries are slow to generate and are supposed to be reusable for many compressions/decompressions)

zstd sounds promising. https://github.com/facebook/zstd/wiki/Zstandard-as-a-patching-engine has more info about using it for patching files like Sparkle currently does. I personally don't enjoy the slow bsdiff generation times so I'm happy to try something else..

We need to also (perhaps more importantly) switch to a different container format than xar, ideally one that is small(?) in surface and not a hassle to embed / maintain (we never had to embed xar). What are our options here?

Maybe we could upgrade and introduce new major versions to the format in piecemeal, starting first with changing the container format.

We also need to upgrade generate_appcast so it picks the correct version to generate patches from based on the Sparkle bundle in the old app (if it doesn't do that already).

We can drop support for the version 1 format that nobody should be using anymore (like from Sparkle 1.9 or something).

Tentatively, I'm going to classify this as a long term goal with some major undertaking, to be in "2.1" milestone, but would like to be proven otherwise :)

kornelski:
One more thing that occurred to me is that we could support deltas that work for multiple versions at the same time. The diffs could use files or file ranges that are common in multiple versions. This way there would be fewer delta files to upload and serve.

zorgiepoo:
In order to keep download files as small as possible doesn't that imply we'd have to download a byte range of the file and servers would have to support that?

Anyway, I think we should first decide on a container format to move away from xar, and get a development branch that uses a new container and have the unit tests pass.

Is choosing the container format going to be intertwined with the delta compression we choose? I guess we will want to compress the entire container. When we generate binary diffs for specific set of individual files, I'm not sure if those will be compressed or not.

kornelski:
[...]

PR #2051, opened 2022-01-02
Add new version 3 binary delta archive format

zorgiepoo:
This introduces a new format (version 3) to our binary delta patches.

Version 2 delta files use the legacy XAR archive/container format which has been deprecated. It's tied to Apple's whim and recent changes temporarily broke our implementation because we weren't creating valid structured XAR files. It's completely undocumented and hard to parse API in terms of correct usage, and it has warnings everywhere about not using it anymore. So it has got to go.

Version 3 introduces a new custom format designed with having data a bit more tightly packed together. The file content blobs and the relative file paths table are kept together in the archive, and the format/implementation is designed such that we read/write/stream the entire delta file in a single pass.

This change also introduces a new `Clone` command which tracks large unchanged files in an app being replicated to new places. This consumes file renames as well. Instead of writing out new file data, we store an index referencing to an old file in the file path table. These commands are only used with large files as I've found it to not make a significant difference for small ones (sometimes it is worse when turned on for small files due to how the whole archive is compressed interesting enough).

See [SPUSparkleDeltaArchive.h](https://github.com/sparkle-project/Sparkle/pull/2051/files#diff-d731376297ce19d0dfbc1592d5d07fed318fe6acc3b76b87f7e8fd4fd125f886) / [SPUSparkleDeltaArchive.m](https://github.com/sparkle-project/Sparkle/pull/2051/files#diff-c68ecc4b1109d218ed61713aa11f8369a94fcf0b809031bac26694b25deb32b6) for more details about the new format/implementation.

Otherwise the binary delta apply/create drivers (aside from supporting clones) remains relatively unchanged. And the unit tests were altered to test both version 2 and version 3 patches. One other misc change though is that we remove the check that disallowed file permissions that aren't 0755 or 0644 as this was an arbitrary safeguard that blocks people.

This work is part of #1902

Changes to generate_appcast and website documentation will come later. For now, the default version when creating new patches is still version 2. The change to make version 3 the default will come later. For now this should be treated as experimental and subject to change.

To use version 3, pass `--version=3` to `./BinaryDelta` when creating new patches.

Additionally, two environment variable options to `BinaryDelta` were introduced for generating both version 2 or 3 patches. These options will likely move to command line options later, so the env variables are temporary:

`SPARKLE_DELTA_COMPRESSION=lzma` (default) or `SPARKLE_DELTA_COMPRESSION=none` to control the compression. Other options supported are bzip2, zlib, lzfse, lz4.

`SPARKLE_DELTA_COMPRESSION_LEVEL=9` (default) to control the compression level, only applicable for bzip2. This value is from 1-9.

## Misc Checklist:

## Testing

- [ ] Sparkle Test App
- [x] Unit Tests
- [ ] My own app
- [x] Other (please specify)

All the unit tests (which were very useful here) should pass, and additional ones were added for testing file cloning.

I've been testing with BinaryDelta manually with a few different apps including some of mine. One of the more interesting tests was between two different builds of Docker.

Generally I have seen delta files to be smaller by about 0.5 - 2 MB. Deltas that include large files being renamed will of course have better savings, but generally I've not seen that many large files are renamed. Small savings are mostly due to metadata and accumulation of it for larger patches.

For the two Docker builds I've been using, I see version 3 patches take a few seconds faster to apply (33 vs 37 seconds) and faster to create (14 minutes vs 26 minutes). I don't know why XAR patch creation is that much slower. The patch file difference in this case (which is compressed the same way; bzip2 with level 9) is 65.7 MB vs 66.8 MB (~1 MB savings). Some other apps you may see smaller savings or larger by a bit (but not exponentially).

The improvements are in small quantities but should always be better than version 2 / XAR across the board.
bsdiff / bspatch is still used and remains unchanged. Our directory tree hash function is also unchanged and the same as version 2.

[EDIT]: we now support lib compression based formats and are switching over to lzma which gives us even better savings while remaining competitive on patch creation/apply times.

macOS version tested: 12.1 (21C52)


zorgiepoo:
I could maybe play with liblzma later (if it had any documentation) but I don't know if it'll be promising.

zorgiepoo:
libcompression based formats are now supported :). Creation now defaults to lzma (level 6) which looks as competitive / around the same as bzip2 on time to create and apply the patch. But savings are even better, by a few MB. For Docker this is 2 MB (66 MB down to 64 MB) and on of my apps I see a 4 MB savings (15 MB to 11 MB).

PR #2087, opened 2022-02-07
Use strcoll_l() for locale-independent comparisons

dan-rigdon-bel:
The update code uses the compareFiles function to determine the sort order of file names in a directory, which is then compared with a list of files in the Sparkle update info. If the list of files of a directory from Sparkle's update is different than what is on disk, the Sparkle abandons the update and does a full download. This bug would prevent Sparkle from updating properly when the file lists are calculated on an English language system, but then checked on a destination Japanese system.

The fix is to use strcoll_l(), which is locale-independent, when comparing file names for sorting. This will ignore any locale changes the host application has made that could potentially affect Sparkle's sort order.

(Insert summary of your pull request here)

Fixes # (issue)

## Misc Checklist:

## Testing

- [ ] Sparkle Test App
- [ ] Unit Tests
- [x] My own app
- [ ] Other (please specify)

### DIFF
--- a/Autoupdate/SUBinaryDeltaCommon.m
+++ b/Autoupdate/SUBinaryDeltaCommon.m
@@ -16,6 +16,7 @@
 #include <sys/mman.h>
 #include <sys/param.h>
 #include <sys/stat.h>
+#include <xlocale.h>
 
 #include "AppKitPrevention.h"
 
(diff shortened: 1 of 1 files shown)


zorgiepoo:
Thank you. I assume cases where this fails is when files have some sort of non-english or accent characters or some such in them. I'm also guessing using the c locale is somewhat equivalent to using strcmp(). Either way, change looks good.

dan-rigdon-bel:
> I assume cases where this fails is when files have some sort of non-english or accent characters or some such in them

Exactly right.  We hit this bug with internationalized file names in Japanese.

zorgiepoo:
I should have asked, do you still have an example of two filenames that don't sort the same way on en_US vs ja_JP?

dan-rigdon-bel:
> I should have asked, do you still have an example of two filenames that don't sort the same way on en_US vs ja_JP?

I'm sorry to say that I do not.  Reversing our changes to get the list would be challenging.

zorgiepoo:
I could probably figure it out if the app was publicly available, but it's not a big deal either way.

PR #2091, opened 2022-02-18
Prevent background update popup

mherrmann:
This PR is necessary for https://github.com/brave/brave-core/pull/12335/.

Fixes # (issue)

## Misc Checklist:

## Testing

- [ ] Sparkle Test App
- [ ] Unit Tests
- [ ] My own app
- [ ] Other (please specify)


mherrmann:
Wrong project, sorry.

zorgiepoo:
As a side note, Sparkle 2 probably better supports what you want to do (it shows an admin auth prompt before app termination, updates can be downloaded but not installed automatically, or they can be disabled for when auth would be required like sparkle-cli does, or you can integrate a custom UI for showing updates inside your settings), but in either case you'll probably want a UI to manually trigger installing for updates because otherwise an administrator dialog may randomly pop up to the user. In Sparkle 1 that will be when the app is terminated.

mherrmann:
Yes, thank you very much for this input :-)

PR #2096, opened 2022-02-28
Change XPC Service product names to be more human readable

### LINKED ISSUE #2095: InstallerLauncher bundle identifier visible on macOS Monterey
### Description of the problem
`org.sparkle-project.InstallerLauncher` bundle identifier is visible in authorization (AuthorizationRightSet) prompt (on macOS Monterey only)

https://prnt.sc/zvNF4baRXVCM

### Do you use Sandboxing in your app?
Yes

### Version of `Sparkle.framework` in the latest version of your app
2.0.0-rc1

zorgiepoo:
The XPC Service names should not be bundle IDs because they may surface up in the UI.

Particularly, the Installer XPC Service may show up in the authorization prompt when authorization is required to install an update.

This will need an accompanying documentation PR change, and a notice that these services changed to whoever may be using custom scripts that process these services.

I doubt the AuthorizationRef APIs allow customizing this name that shows up (if it did, I would use the original app name). And last I tested, the `CFBundleName` and `CFBundleDisplayName` keys didn't affect the name shown here. Apparently the app's name only shows up on Monterey and not previous macOS releases like Big Sur.

Before:
Screen Shot 2022-02-27 at 6 45 19 PM

After:
Screen Shot 2022-02-27 at 6 44 20 PM

Fixes #2095

## Misc Checklist:

## Testing

- [x] Sparkle Test App
- [ ] Unit Tests
- [ ] My own app
- [ ] Other (please specify)

After the app is launched and before an update check is made I run:
```
sudo chown root <my-app-path>
```

Tested the authorization dialog shows "Installer" instead of "org.sparkle-project.InstallerLauncher" by default.
Tested app upgrading to new Sparkle version works
Tested that non-sandbox app still uses the original app's name.

macOS version tested: 12.1 (21C52)


zorgiepoo:
Hm for a non-sandbox app, it does use `CFBundleName` ..

zorgiepoo:
I'll leave this here a bit to make sure there aren't any better solutions.

DivineDominion:
@zorgiepoo I noticed these titles in some apps, too -- thanks for figuring out where they come from.

So for the optimal readable title, one would

- not need to change anything for non-Sandboxed apps b/c they use the app title anyway
- maybe change "Installer" to e.g. "MyAppName Installer" to disambiguate for Sandboxed apps?

danielpunkass:
@DivineDominion If I understand the challenge right, the host app's name can't be systematically swapped in at runtime, so to have the "MyAppName Installer" solution it would have to be done as a packaging process by the host app's own build process. I have done some similar trickerey in my own apps to rename helpers to customize them for the host app.

PR #2112, opened 2022-04-23
Replace deprecated code with newer APIs

Eitot:
I ran the static analyser with the deployment target set to macOS 12.3 to see if any deprecation warnings show up. These commits address some of those warnings (e.g. not the C-based Keychain APIs) by using newer APIs where available.

## Misc Checklist:

## Testing

- [X] Sparkle Test App
- [X] Unit Tests
- [ ] My own app
- [x] Other (Clang-Analyser)

macOS version tested: 12.3.1


Sparkle/SUApplicationInfo.m
```diff
@@ -12,6 +12,7 @@
 #import "SUBundleIcon.h"
 #import "SUHost.h"
 #import <AppKit/AppKit.h>
+#import <UniformTypeIdentifiers/UniformTypeIdentifiers.h>
```
zorgiepoo, line 15: UniformTypeIdentifiers may need to be added in linked libraries for Sparkle framework and Installer Progress targets and as `Optional` to make sure it's weak linked so it's fine on systems less than macOS 11.

zorgiepoo:
Looks good besides one weak linkage concern. After that I will probably give it a basic test run on one of my VMs running < macOS 11.

Sparkle/SUApplicationInfo.m
zorgiepoo: Alright Xcode already made this weak automatically so this is not an issue, never mind!

Eitot: I was under the impression that Clang does this automatically when CLANG_MODULES_AUTOLINK is enabled (which is the case for Sparkle and Installer Progress).

`otool` indicates this as well (if I interpret this correctly).

zorgiepoo: Yep you're right.

zorgiepoo:
Merge conflict from previous changes.

zorgiepoo:
Gave a run on macOS 10.12, seems good, thanks!

zorgiepoo:
(The deprecated keychain APIs are only used for DSA keys btw which we've deprecated ourselves).

PR #2170, opened 2022-06-20
Deprecate -s flag and add --ed-key-file option to generate_appcast

### LINKED ISSUE #2168: Reassess options that allow passing signing key as command line argument to signing tools
Our preferable way of maintaining keys is through the keychain.

To support the CI platforms and encrypted secrets keys (see [this discussion](https://github.com/sparkle-project/Sparkle/issues/1766#issuecomment-782012814)), we provide an option to pass the private EdDSA key as a command line argument to sign_update or generate_appcast. But this is not likely secure (can be sniffed from other processes via `ps` initially brought up from  #2161)

It should be better for our tools to support a specific environment variable (edit: passing as stdin is even better) instead, and that seems to be what [GitHub suggests doing](https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow):

> Avoid passing secrets between processes from the command line, whenever possible. Command-line processes may be visible to other users (using the ps command) or captured by [security audit events](https://docs.microsoft.com/windows-server/identity/ad-ds/manage/component-updates/command-line-process-auditing). To help protect secrets, consider using environment variables, STDIN, or other mechanisms supported by the target process

Note we also have an option to pass the signing key as a file to sign_update (but not generate_appcast -- something we should also fix).

zorgiepoo:
Passing a raw secret to command line argument (even when using a 'secret' environment variable) is discouraged and unsafe. After reading up several sources online, it's recommended to pass such a secret that CI platforms may provide as standard input to the program, using a built-in like `echo`. 

So we deprecate the `-s` flag in sign_update and generate_appcast, add a `--ed-key-file` option to `generate_appcast` that takes a file just like `sign_update -f` (or `--ed-key-file` now), and interpret `-` as the standard input file to read the key from.

I'm pretty certain now this is the right way forward but I'll leave this PR open a bit in case anyone wants to chime in. 

As our documentation generally recommends using the keychain which is the default path for these tools, I don't think we need to update any website documentation (edit: I should update the migration page). The help pages for these tools will be updated.

Fixes #2168

## Misc Checklist

## Testing

- [ ] Sparkle Test App
- [ ] Unit Tests
- [ ] My own app
- [x] Other (please specify)

For `sign_update` and `generate_appcast` tested that:
* `--ed-key-file path-to-file` works
* `--ed-key-file -` works
* `-s` still works and prints deprecation warning to stderr for `sign_update`, and to stdout for `generate_appcast`

macOS version tested: 12.4 (21F79)

### DIFF
--- a/generate_appcast/main.swift
+++ b/generate_appcast/main.swift
@@ -59,14 +59,17 @@ struct GenerateAppcast: ParsableCommand {
     @Option(help: ArgumentHelp("The account name in your keychain associated with your private EdDSA (ed25519) key to use for signing new updates."))
     var account : String = "ed25519"
     
-    @Option(name: .customShort("s"), help: ArgumentHelp("The private EdDSA string (128 characters). If not specified, the private EdDSA key will be read from the Keychain instead.", valueName: "private-EdDSA-key"))
-    var privateEdString : String?
+    @Option(name: .customLong("ed-key-file"), help: ArgumentHelp("Path to the private EdDSA key file. If not specified, the private EdDSA key will be read from the Keychain instead. '-' can be used to echo the EdDSA key from a 'secret' environment variable to the standard input stream. For example: echo \"$PRIVATE_KEY_SECRET\" | ./\(programName) --ed-key-file -", valueName: "private-EdDSA-key-file"))
+    var privateEdKeyPath: String?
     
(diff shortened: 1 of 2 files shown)


jeremyhu (approved):
Thanks.

PR #2275, opened 2022-10-24
Update Czech translation 

isametry:
I translated the strings that are currently missing (as of 2.4). 

Many of them are just variations of what had already been there, so there I just used the same language. Still, I would love if someone else could briefly check the translations.

@vojtakonarik has contributed Czech here before, but not sure if that account is still active… I also know that @junkycoder and @jezevec10 have done this elsewhere before.

Thanks :)


vojtakonarik:
Hi, I will check it ASAP, thanks!

jezevec10 (approved):

zorgiepoo:
Thank you.

PR #2373, opened 2023-05-05
Expose installed version string to update notes HTML page

Cykelero:
In the HTML page that contains the release notes:
- Inject a `sparkleInstalledVersion` global variable containing the version string of the current app install
- Add the `sparkle-installed-version` class on elements with a matching `data-sparkle-version` attribute

Initial discussion in #2367.

## Misc Checklist

- [X] My change requires a documentation update on [Sparkle's website repository](https://github.com/sparkle-project/sparkle-project.github.io)

## Testing

- [X] Sparkle Test App (using tweaked sample release notes from my own app)
	- HTML notes contain 0, 1, or 2 matching `data-sparkle-version` elements
- [ ] Unit Tests
- [ ] My own app
- [ ] Other (please specify)

macOS version tested: 13.3.1 (22E261)

### DIFF
--- a/Sparkle/SUUpdateAlert.m
+++ b/Sparkle/SUUpdateAlert.m
@@ -334,7 +334,7 @@ static NSString *const SUUpdateAlertTouchBarIndentifier = @"" SPARKLE_BUNDLE_IDE
         } else
 #endif
         {
-            _releaseNotesView = [[SUWKWebView alloc] initWithColorStyleSheetLocation:colorStyleURL fontFamily:defaultFontFamily fontPointSize:defaultFontSize javaScriptEnabled:javaScriptEnabled];
+            _releaseNotesView = [[SUWKWebView alloc] initWithColorStyleSheetLocation:colorStyleURL fontFamily:defaultFontFamily fontPointSize:defaultFontSize javaScriptEnabled:javaScriptEnabled installedVersion: _host.version];
         }
     }
     
--- a/Sparkle/SUWKWebView.h
+++ b/Sparkle/SUWKWebView.h
@@ -15,7 +15,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 SPU_OBJC_DIRECT_MEMBERS @interface SUWKWebView : NSObject <SUReleaseNotesView>
 
-- (instancetype)initWithColorStyleSheetLocation:(NSURL *)colorStyleSheetLocation fontFamily:(NSString *)fontFamily fontPointSize:(int)fontPointSize javaScriptEnabled:(BOOL)javaScriptEnabled;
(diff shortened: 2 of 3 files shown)


Cykelero:
I chose `sparkleInstalledVersion` and `sparkle-installed-version` for brevity, and to namespace the identifiers to avoid collisions (and make their special nature clearer in author HTML).

Also, I'd be happy to open a PR to add a bit of description and sample code to the docs, too. Do you have a suggestion of which page should cover the subject?

Cykelero:
My bad, I'll make the variable local.

As for safely passing in the version string: right now the code escapes `` ` `` and `"` (I need to add `\`), which I feel would be enough to prevent breakage—although I can't be sure.
Do you have an alternative in mind?

Cykelero:
No worries!

I just made the changes you requested. For the `nil` checks, I chose to make the version injection log just a warning; since this is a very optional feature, and existing codebases might have version numbers that don't pass the check.

ychin:
Hi, I was playing with this in the 2.5 beta release and the behavior doesn't really make sense to me. Seems like `sparkle-installed-version` is a direct match instead of based on a ">=" check based on semantic versioning?

Let's say I'm currently on r177. A new release 178 has just been released. My appcast shows the last 3 releases so it contains 176, 177, 178, with the XML looking like this:

```xml
      <title>MacVim release-178</title>
      <description><![CDATA[
      <style>
          div.sparkle-installed-version { opacity: 0.5; }
      </style>
      <div data-sparkle-version="176"> This is r176 </div>
      <div data-sparkle-version="177"> This is r177 </div>
      <div data-sparkle-version="178"> This is r178 </div>
      ]]></description>
      <sparkle:version>178</sparkle:version>
```

I see this in the release notes:

image

Seems like only r177 is dimmed. I would imagine most people would want r176 / r177 dimmed, compared to r178? There are tricks we could play if we nest the div's but that seems like a big hack.

ychin:
From the initial discussion (https://github.com/sparkle-project/Sparkle/discussions/2367) seems like using a ">=" check was the initial desire as well, from the screenshot, but seems like that got changed (intentional or not) during the implementation?

FWIW I am partial to a JavaScript solution (discussed as an alternative in the discussion). It's more flexible and allows custom logic. I turned on JavaScript for my release notes anyway mostly since otherwise HTML videos wouldn't work (unfortunately), but I do understand that the default is to not have it turned on for security reasons so maybe that can't be a global solution, and if the CSS solution works I'm fine with that too.

Note: I was actually wondering if I could just encode the current version in the URL query parameters using `feedParametersForUpdater` (something like https://example.com/appcast.xml?currentVersion=177), so that the JS code could parse `document.URL` and retrieve that information (since my release notes are just hosted off a GitHub page with no custom logic), but seems like Sparkle is creating the page in an about:blank page so that doesn't work.

ychin:
> Release notes are normally presented from newest (top) to oldest (bottom) -- descending.

> The sample there shows how to stylize the current installed version and how to stylize all versions that show below the current installed one.

Ah ok right. Now that makes sense. I missed the tilde that specifies it.

PR #2564, opened 2024-05-20
Adjust the layout of anonymous system profile info to align better with the rest of the panel's UI

danielpunkass:
These are minor tweaks to finesse the appearance of the SUUpdatePermissionPrompt when the "Include anonymous system profile" disclosure button is selected.

## Misc Checklist

## Testing

- [x] Sparkle Test App
- [ ] Unit Tests
- [x] My own app
- [ ] Other (please specify)

I compared the appearance of the panel by forcing "shouldPrompt" to YES and visually examining the resulting panel. I also created comparison images to show the effect of the change both with -NSShowAllViews 1, and without:

BeforeAndAfter

macOS version tested: macOS Sonoma 14.5


zorgiepoo:
Thanks, a somewhat subtle change. I definitely did change the alignment a bit when I made this window use a stack view in 2.x. I do wonder if there should be a left-padding for the table view, from the explanation text above it. There was padding in 1.x there and I think it was intentional (I didn't design it though), so the table doesn't look like it's too much to the "left" of the explanation text above it. I'm not a pixel/UI expert, though.

danielpunkass:
I imagine there are probably lots of "right" ways to design this, but the thing that looked particularly off to me was that neither the text nor the table view looked indented enough to look intentional. So in the absence of somebody putting the effort in to do a more substantial redesign, I think this is at least a good refinement of the status quo.

danielpunkass:
I think it is common though for table views to be aligned flush with the text above them. Just one example I found by looking in Safari's preferences: 
image

zorgiepoo:
Fair enough, sounds good, thanks!

PR #2576, opened 2024-06-04
fix: add RLM chars to he locale (#2575)

### LINKED ISSUE #2575: Layout issues in RTL languages
## Summary

Please note that some of the problems will only be presented after the completion of the Hebrew translation (in #2573)

Some languages, such as Hebrew and Arabic, are written from right to left (RTL). Mixed sentences are written from right to left, with terms embedded in their original form, for example:

אני משתמש ב-Github כל יום
Github הוא כלי חשוב

However, when there is a block of text that is entirely in English, it should be spread from left to right.
See also Apple's guide: https://developer.apple.com/design/human-interface-guidelines/right-to-left

I found a few things that don't match in Sparkle:

1. This is an RTL sentence, but it starts with an English term (LTR) which causes it to render incorrectly (note that the question mark escapes to the beginning of the line).
CleanShot 2024-06-03 at 15 55 54@2x
2. Same as 1, in "You're up to date" window:
CleanShot 2024-06-03 at 19 50 59@2x
3. The system information is always in English, so the table needs to be LTR to be displayed correctly:
CleanShot 2024-06-03 at 16 05 58@2x

## Version

The Sparkle Test App

ShlomoCode:
https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPInternational/SupportingRight-To-LeftLanguages/SupportingRight-To-LeftLanguages.html

(Insert summary of your pull request here)

Fixes #2575 (Partially)

## Misc Checklist

## Testing

- [x] Sparkle Test App
- [ ] Unit Tests
- [ ] My own app
- [ ] Other (please specify)

macOS version tested: 14.5


zorgiepoo:
Thank you.

PR #2712, opened 2025-03-28
Set STRINGS_FILE_OUTPUT_ENCODING build setting to "binary"

nickasd:
This reduces the final framework bundle size by 193 KB.

## Testing

- [ ] Sparkle Test App
- [x] Unit Tests
- [ ] My own app
- [ ] Other (please specify)

macOS version tested: macOS 15.3.2


zorgiepoo:
Thanks. Tested this works when changing language on 15.4 (24E248), 10.14.6 VM.

PR #2720, opened 2025-04-28
Rework the logic for avoiding re-sending the system profile more frequently than once a week

### LINKED ISSUE #278: Remove binary compare in parameterizedFeedURL method
This line of code in parameterizedFeedURL fails for me (line 4800;

sendingSystemProfile &= (-[lastSubmitDate timeIntervalSincenow] >= oneWeek);

This is always NO for me even when sendingSystemProfile is YES and lastSubmitDate is distandPast.  Turns out the problem is the binary compare &=.  If I break out the compare against oneWeek into a separate BOOL say:

BOOL timeLapsed = (-[lastSubmitDate timeIntervalSincenow] >= oneWeek);

and then change the logic to be:

sendSystemProfile = sendSystemProfile && timeLapsed;

it works ok.  I've had this issue before in C code where the binary compare is not reliable.  You need to use the Boolean compare.

danielpunkass:
Rework the logic for avoiding re-sending the system profile more frequently than once a week, so that it will not use a potentially problematic binary compare operation.

Fixes #278 

## Testing

- [ ] Sparkle Test App
- [ ] Unit Tests
- [ ] My own app
- [x] Line by line debugger evaluation with synthesized date values.

macOS version tested: macOS 15.4.1 (24E263)

### DIFF
--- a/Sparkle/SPUUpdater.m
+++ b/Sparkle/SPUUpdater.m
@@ -1094,12 +1094,17 @@ static NSString *escapeURLComponent(NSString *str) {
     BOOL sendingSystemProfile = [self sendsSystemProfile];
 
     // Let's only send the system profiling information once per week at most, so we normalize daily-checkers vs. biweekly-checkers and the such.
-    NSDate *lastSubmitDate = [_host objectForUserDefaultsKey:SULastProfileSubmitDateKey];
-    if (!lastSubmitDate) {
-        lastSubmitDate = [NSDate distantPast];
+    if (sendingSystemProfile) {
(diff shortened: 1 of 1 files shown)


zorgiepoo:
Thanks. Did you actually see Sparkle re-send the system profile data more frequently than once a week from the original reporter's claim and it being caused by this, or did you just feel like improving this? That will impact how important I find this change and what release to put it in.

danielpunkass:
I did not see the behavior, and couldn't reproduce it. So, low priority. I just happened to be scrolling through Sparkle bugs and got curious what the oldest one was ;) By the time I had played around with it trying to reproduce, I thought I might as well submit a PR.

zorgiepoo:
Gotcha. Works well over here.

PR #2726, opened 2025-05-14
InstallerProgressAppController: Workaround a corner case in which the bundle path of a running application contains Contents/MacOS/Executable

### LINKED ISSUE #2725: Sparkle does not restart app after update completes (sandboxed app with SUEnableInstallerLauncherService=true)
### Description of the problem

Sparkle does not restart app after update completes.

### Do you use Sandboxing in your app?

My app is sandboxed.  It has SUEnableInstallerLauncherService=true set in Info.plist and sanbox exceptions to connect to the following ports:

        <string>$(PRODUCT_BUNDLE_IDENTIFIER)-spks</string>
        <string>$(PRODUCT_BUNDLE_IDENTIFIER)-spki</string>

The update works fine.  But users need to manually quit and relaunch the app.

### Version of `Sparkle.framework` in the latest version of your app

2.6.4

### Version of `Sparkle.framework` in the old version of app that your users have (or N/A)

2.6.4

### Sparkle's output from Console.app
```
2025-05-14 01:20:05.169512-0700 0x22177f   Info        0xca1a25             77180  0    Redacted: (SparkleUpdater) [Redacted:Redacted] Sparkle Update will install update
2025-05-14 01:20:10.097188-0700 0x22177f   Info        0xca1a25             77180  0    Redacted: (SparkleUpdater) [Redacted:Redacted] Sparkle Update finished update cycle for update error: nil
2025-05-14 01:21:25.097836-0700 0x226b65   Info        0xcb51a0             77939  0    Redacted: (SparkleUpdater) [Redacted:Redacted] Sparkle Update finished downloading
2025-05-14 01:21:27.403964-0700 0x226f8b   Default     0xcb51a0             77971  0    Autoupdate: [org.sparkle-project.Sparkle:Sparkle] Extracting using '/usr/bin/ditto' '-x' '-k' '-' < '/var/root/Library/Caches/Redacted/org.sparkle-project.Sparkle/Installation/rz09IAFKF/Redacted..zip' '/var/root/Library/Caches/Redacted/org.sparkle-project.Sparkle/Installation/rz09IAFKF/c8893zK1s'
2025-05-14 01:21:29.805432-0700 0x226f89   Default     0xcb51a0             77971  0    Autoupdate: [org.sparkle-project.Sparkle:Sparkle] OK: EdDSA signature is correct
2025-05-14 01:21:29.826133-0700 0x226f89   Error       0xcb51a0             77971  0    Autoupdate: [org.sparkle-project.Sparkle:Sparkle] Code signature of the new version doesn't match the old version: identifier RedactedTrampoline and certificate root = H"0c2b0a9baf21fd7ce52e50555bfd347deafbcad5". Please ensure that old and new app is signed using exactly the same certificate.
2025-05-14 01:21:29.830408-0700 0x226f89   Default     0xcb51a0             77971  0    Autoupdate: [org.sparkle-project.Sparkle:Sparkle] old info: {
<redacted>
}
2025-05-14 01:21:29.831057-0700 0x226f89   Default     0xcb51a0             77971  0    Autoupdate: [org.sparkle-project.Sparkle:Sparkle] new info: {
<redacted>
}
2025-05-14 01:21:30.043080-0700 0x226fe2   Default     0xcb51a0             77971  0    Autoupdate: [org.sparkle-project.Sparkle:Sparkle] Skipping atomic rename/swap and gatekeeper scan because Autoupdate is not signed with same identity as the new update Redacted.app
2025-05-14 01:21:30.064454-0700 0x226b65   Info        0xcb51a0             77939  0    Redacted: (SparkleUpdater) [Redacted:Redacted] Sparkle Update will install update
2025-05-14 01:21:38.898940-0700 0x226b65   Info        0xcb51a0             77939  0    Redacted: (SparkleUpdater) [Redacted:Redacted] Sparkle Update finished update cycle for update error: nil%
```

Note the SparkleUpdater logs there are from our `SPUUpdaterDelegate` (which was only emitting logging to help track things down here...)

### Steps to reproduce the behavior

I check for updates and am presented with the update release notes.  
I click Install Update in the release notes window.

Expected (based on behavior I see in other apps):
I should get a dialog with a "Install and Relaunch" button.
Tapping on "Install and Relaunch" should prompt for authentication if required, install, and relaunch the app.

Observed:
I am immediately prompted to auth for the install
I see an Extracting Update dialog and the soon after an 'Update Installed' dialog.  The app does not relaunch, and the user is not told to relaunch it.

--

I debugged this a bit in lldb and what I observed is that we're getting a `SPUInstallationFinishedStage3` message while `_relaunch` is set to NO.

It looks like `_relaunch` is only set to `YES` in `-installWithToolAndRelaunch:displayingUserInterface:`, but that isn't getting called.

That in turn is because finishInstallationWithResponse:displayingUserInterface: isn't being called...

That looks to be usually called in -basicDriverDidFindUpdateWithAppcastItem:secondaryAppcastItem: when state is SPUUserUpdateStageInstalling , but the only time I ever see that called is for SPUUserUpdateStageNotDownloaded.

It looks like this is just a gap in how SUEnableInstallerLauncherService=true support was implemented.  That seems to completely bypass these function calls, which results in _relaunch never getting set to YES, so we end up passing relaunch=no when receiving SPUInstallationFinishedStage3...

jeremyhu:
This handles a corner case in which [NSRunningApplication runningApplicationsWithBundleIdentifier:bundleIdentifier] returns us an instance in which the bundle path contains "Contents/MacOS/<executable>" by ignoring those trailing path components.

Fixes #2725

### DIFF
--- a/Sparkle/InstallerProgress/InstallerProgressAppController.m
+++ b/Sparkle/InstallerProgress/InstallerProgressAppController.m
@@ -224,6 +224,7 @@ static const NSTimeInterval SUTerminationTimeDelay = 0.3;
     NSMutableArray<NSRunningApplication *> *matchedRunningApplications = [[NSMutableArray alloc] init];
     
     if (bundleIdentifier != nil && bundlePathComponents != nil) {
+        NSFileManager *fileManager = [NSFileManager defaultManager];
         NSArray *runningApplications = [NSRunningApplication runningApplicationsWithBundleIdentifier:bundleIdentifier];
         
         // If we find any running application that is translocated and looks like the bundle, we should record those too
@@ -235,6 +236,20 @@ static const NSTimeInterval SUTerminationTimeDelay = 0.3;
             NSString *candidatePath = runningApplication.bundleURL.URLByResolvingSymlinksInPath.path;
             if (candidatePath != nil) {
                 NSArray<NSString *> *candidatePathComponents = candidatePath.pathComponents;
+                NSUInteger candidatePathComponentsCount = candidatePathComponents.count;
+
+                // Workaround cases where macOS appends Contents/MacOS/<executable> to the bundle path which can happen in corner cases (eg: mishandled bundles or helper executables).
+                // https://github.com/sparkle-project/Sparkle/issues/2725
+                BOOL candidatePathIsDir = YES;
+                if ([fileManager fileExistsAtPath:candidatePath isDirectory:&candidatePathIsDir] && !candidatePathIsDir) {
+                    if (candidatePathComponentsCount > 3 && [candidatePathComponents[candidatePathComponentsCount - 3] isEqualToString:@"Contents"] && [candidatePathComponents[candidatePathComponentsCount - 2] isEqualToString:@"MacOS"]) {
+                        NSMutableArray<NSString *> *trimmedBundlePath = [candidatePathComponents mutableCopy];
+                        [trimmedBundlePath removeObjectsInRange:NSMakeRange(candidatePathComponentsCount - 3, 3)];
+                        candidatePathComponents = trimmedBundlePath;
+                        candidatePathComponentsCount -= 3;
+                    }
+                }
+
                 if ([candidatePathComponents isEqualToArray:bundlePathComponents]) {
                     [matchedRunningApplications addObject:runningApplication];
                 } else if (matchedRunningApplications.count == 0 && candidatePathComponents.count > 0 && bundlePathComponents.count > 0) {


Sparkle/InstallerProgress/InstallerProgressAppController.m
```diff
@@ -235,6 +235,14 @@ - (void)cleanupAndExitWithStatus:(int)status error:(NSError * _Nullable)error __
             NSString *candidatePath = runningApplication.bundleURL.URLByResolvingSymlinksInPath.path;
             if (candidatePath != nil) {
                 NSArray<NSString *> *candidatePathComponents = candidatePath.pathComponents;
+
+                // Workaround cases where macOS appends Contents/MacOS/<executable> to the bundle path which can happen in corner cases
+                // https://github.com/sparkle-project/Sparkle/issues/2725
+                NSUInteger candidateCount = candidatePathComponents.count;
+                if ([candidatePathComponents[candidateCount - 3] isEqualToString:@"Contents"] && [candidatePathComponents[candidateCount - 2] isEqualToString:@"MacOS"]) {
```
zorgiepoo: Can we also test if `candidatePathComponents[candidateCount - 1].pathExtension.length == 0 &&`?
It's unlikely but I could technically create `Contents/MacOS/` folder anywhere on my system and place an .app in there, or an app could be embedded in some other app's `Contents/MacOS/`... However it's unusual if the bundle path is not a bundle.

zorgiepoo: For my sanity please add a `candidateCount >= 3 &&` bounds-safety check.

edit: corrected, meant `>=`, not `>`

Sparkle/InstallerProgress/InstallerProgressAppController.m
```diff
@@ -235,6 +235,14 @@ - (void)cleanupAndExitWithStatus:(int)status error:(NSError * _Nullable)error __
...
+                    candidatePathComponents = [candidatePathComponents objectsAtIndexes:[NSIndexSet indexSetWithIndexesInRange:NSMakeRange(0, candidateCount - 3)]];
```
zorgiepoo: `-[NSArray subarrayWithRange:]` is easier.

Sparkle/InstallerProgress/InstallerProgressAppController.m
```diff
@@ -235,6 +235,14 @@ - (void)cleanupAndExitWithStatus:(int)status error:(NSError * _Nullable)error __
...
+                }
+
                 if ([candidatePathComponents isEqualToArray:bundlePathComponents]) {
                     [matchedRunningApplications addObject:runningApplication];
```
zorgiepoo, line 254: I wonder if it's ever possible to have both:

```
runningApp1: /Applications/MyApp.app/Contents/MacOS/SuperSpecialHelperTool
runningApp2: /Applications/MyApp.app
```

It may be safer to insert abnormal cases towards the end of the `matchedRunningApplications` array (but still not add any translocated instances if there are these cases). Although all instances are used for sending quit events, Sparkle currently only uses the first running app instance for monitoring when the app dies.

Sparkle/InstallerProgress/InstallerProgressAppController.m
jeremyhu: Good catch.  Yes.

jeremyhu: Checking `candidatePathComponents[candidateCount - 1].pathExtension.length == 0` isn't right either.  It's perfectly legal to have a '.' in the name of the executable.  I'd suspect that is more likely than someone creating /Applications/Contents/MacOS/Foo.app, so I'm going to leave that check out.

> It's unusual if the bundle path is not a bundle.

Yeah, I get that...  This is a bug in macOS dealing incorrectly with our bundle layout and process bootstrapping.  I can go into more details if you want, but it's more than I want to explain unless someone is actually interested in the gritty details.

Sparkle/InstallerProgress/InstallerProgressAppController.m
jeremyhu: Is this what you were thinking?

Sparkle/InstallerProgress/InstallerProgressAppController.m
zorgiepoo: Bar.app/Contents/MacOS/Foo.app is common for helper app Foo.app inside parent app Bar.app

Can we check if the path extension isn't ".app" which narrows it a bit more? We're trying to fix a narrow case while not introducing surprises for other cases which I hope people are not using (updating a helper app inside a parent app, and not having the parent app terminate), but don't want to later find out. 

> Yeah, I get that... This is a bug in macOS dealing incorrectly with our bundle layout and process bootstrapping. I can go into more details if you want, but it's more than I want to explain unless someone is actually interested in the gritty details.

The info that would allow me to create a repro case would be helpful, although I suspect I may have that if I try to create an app whose main binary exec's a helper binary? But if this is too much of a pain then so be it.

Sparkle/InstallerProgress/InstallerProgressAppController.m
```diff
@@ -243,15 +249,27 @@ - (void)cleanupAndExitWithStatus:(int)status error:(NSError * _Nullable)error __
                     if (lastBundlePathComponent != nil && lastCandidatePathComponent != nil && [lastBundlePathComponent isEqualToString:lastCandidatePathComponent] && [candidatePathComponents containsObject:@"AppTranslocation"]) {
                         [potentialMatchingTranslocatedRunningApplications addObject:runningApplication];
                     }
+                } else if (candidatePathComponentsCount > 3 && [candidatePathComponents[candidatePathComponentsCount - 3] isEqualToString:@"Contents"] && [candidatePathComponents[candidatePathComponentsCount - 2] isEqualToString:@"MacOS"]) {
```
zorgiepoo: I would put this before the app translocation case.

Sparkle/InstallerProgress/InstallerProgressAppController.m
```diff
@@ -229,12 +229,18 @@ - (void)cleanupAndExitWithStatus:(int)status error:(NSError * _Nullable)error __
...
         for (NSRunningApplication *runningApplication in runningApplications) {
             // Comparing the URLs hasn't worked well for me in practice, so I'm comparing the file paths instead
             NSString *candidatePath = runningApplication.bundleURL.URLByResolvingSymlinksInPath.path;
             if (candidatePath != nil) {
                 NSArray<NSString *> *candidatePathComponents = candidatePath.pathComponents;
+                NSUInteger candidatePathComponentsCount = candidatePathComponents.count;
+
                 if ([candidatePathComponents isEqualToArray:bundlePathComponents]) {
                     [matchedRunningApplications addObject:runningApplication];
                 } else if (matchedRunningApplications.count == 0 && candidatePathComponents.count > 0 && bundlePathComponents.count > 0) {
```
zorgiepoo, line 255: The `matchedRunningApplications.count == 0 ` was an optimization but now needs to also account for `&& potentialMatchingHelperOrBuggyBundleApplications.count == 0`, or I guess removing this if the optimization is premature (but I'm more leaning on extending it).

Sparkle/InstallerProgress/InstallerProgressAppController.m
```diff
@@ -243,15 +249,27 @@ - (void)cleanupAndExitWithStatus:(int)status error:(NSError * _Nullable)error __
...
         // Non-translocated apps take priority first
         // And we only use translocated version of apps if there are no regular apps matched
         if (matchedRunningApplications.count == 0) {
             [matchedRunningApplications addObjectsFromArray:potentialMatchingTranslocatedRunningApplications];
         }
+
+        // Append any helper applications or processes with buggy bundle paths to the list,
+        // so we kill them as well.  These come at the end, so they are terminated but not
+        // the ones monitored for termination (unless the list is otherwise empty).
+        if (potentialMatchingHelperOrBuggyBundleApplications.count != 0) {
```
zorgiepoo: This should also be before the app translocation case.

Sparkle/InstallerProgress/InstallerProgressAppController.m
zorgiepoo: Yes just with a couple small comments about handling it before translocation if's

Sparkle/InstallerProgress/InstallerProgressAppController.m
zorgiepoo: I don't reproduce the issue with [Audacity](https://www.audacityteam.org/download/mac/) ([source](https://github.com/audacity/audacity/blob/master/au3/mac/Wrapper.c)) by the way. Its app has CFBundleExecutable set to Wrapper, which is a helper executable that then exec's main executable as far as I know. I also wasn't able to create a basic test app that simulated this / tripped up NSRunningApplication bundleURL. There's probably some more nuance here.

edit: Steam also does much weirder things by having its extensionless app bundle redirected in ~/Library/Application Support/ but the NSRunningApplication paths kind of make sense

jeremyhu: > Bar.app/Contents/MacOS/Foo.app is common for helper app Foo.app inside parent app Bar.app
> 
> Can we check if the path extension isn't ".app" which narrows it a bit more? We're trying to fix a narrow case while not introducing surprises for other cases which I hope people are not using (updating a helper app inside a parent app, and not having the parent app terminate), but don't want to later find out.

That feels a bit arbitrary... how about I check if the bundle path is a file vs a directory?

> The info that would allow me to create a repro case would be helpful, although I suspect I may have that if I try to create an app whose main binary exec's a helper binary? But if this is too much of a pain then so be it.

So the case here is that we have two executables in Contents/MacOS, eg:

Foo.app/Contents/MacOS/Foo
Foo.app/Contents/MacOS/FooTrampoline

The bundle's Info.plist has CFBundleExecutable=FooTrampoline

Foo has an embedded __info_plist section in its mach-o that is identical to the bundle's Info.plist except CFBundleExecutable=Foo.  Note if you're trying to make an Xcode project to do this, the app bundle target in Xcode does not honor CREATE_INFOPLIST_SECTION_IN_BINARY, so you need to do it via OTHER_LDFLAGS , eg: `OTHER_LDFLAGS = $(inherited) -Xlinker -sectcreate -Xlinker __TEXT -Xlinker __info_plist -Xlinker $(DERIVED_FILE_DIR)/Foo-Info.plist` ... also be aware that this really needs to be -Xlinker because of a bug in XCBuild that doesn't handle -Wl correctly when doing dependency tracking.

FooTrampoline in our case has main that kicks off some async work via GCD and then calls `[NSRunLoop.mainRunLoop run]`.  That asynchronous work does some setup / install bits and then re-execs the main executable with posix_spawn, eg:

```
static void reexec_as_app(char * _Nonnull argv[], char * _Nonnull envp[]) {
    // Determine the path to our executable
    uint32_t executablePathSize = 0;
    (void)_NSGetExecutablePath(NULL, &executablePathSize);
    NSCAssert(executablePathSize > 0, @"Failed to determine path for main executable.");

    char * const executablePath = malloc(executablePathSize);
    int result = _NSGetExecutablePath(executablePath, &executablePathSize);
    NSCAssert(result == 0, @"Failed to determine path for main executable.");

    // chop off the "Trampoline" suffix to get the path to main executable
    executablePath[strlen(executablePath) - strlen("Trampoline")] = '\0';

    // Set argv[0] to be the executable path
    argv[0] = executablePath;

    // Spawn the real main executable
    posix_spawnattr_t attr;
    os_assumes_zero(posix_spawnattr_init(&attr));

    short flags = POSIX_SPAWN_SETEXEC;
    os_assumes_zero(posix_spawnattr_setflags(&attr, flags));

    pid_t pid;
    errno_t const err = posix_spawnp(&pid, executablePath, NULL /* &file_actions */, &attr, argv, envp);
    NSCAssert(err == 0, @"Failed to spawn main executable (%s): %d %s", executablePath, err, strerror(errno));
    __builtin_trap();
}
```

> I don't reproduce the issue with [Audacity](https://www.audacityteam.org/download/mac/) ([source](https://github.com/audacity/audacity/blob/master/au3/mac/Wrapper.c)) by the way. Its app has CFBundleExecutable set to Wrapper, which is a helper executable that then exec's main executable as far as I know. I also wasn't able to create a basic test app that simulated this / tripped up NSRunningApplication bundleURL. There's probably some more nuance here.

Does Audacity set the __info_plist section in their main binary?  I needed to do that in order to get other things to function correctly in the system, but it's possible Audacity didn't hit those particular quirks.

jeremyhu: Also, we fixed the macOS bug in Tahoe, so if you're trying to hit this, be sure you're on Sequoia.

zorgiepoo: > That feels a bit arbitrary... how about I check if the bundle path is a file vs a directory?

It's an IO check but this should be fine.

> Does Audacity set the __info_plist section in their main binary? I needed to do that in order to get other things to function correctly in the system, but it's possible Audacity didn't hit those particular quirks.

I don't think so, and I think Audacity uses `execve()` rather than posix_spawn -- https://github.com/audacity/audacity/blob/master/au3/mac/Wrapper.c

> Also, we fixed the macOS bug in Tahoe, so if you're trying to hit this, be sure you're on Sequoia.

That's great. I'd like to disable this workaround on macOS 26+ then, in similar spirit to other OS bug workarounds I've had to place in Sparkle code. Assuming rest of the PR is good, I may do this in a followup.

Sparkle/InstallerProgress/InstallerProgressAppController.m
```diff
@@ -235,6 +236,20 @@ - (void)cleanupAndExitWithStatus:(int)status error:(NSError * _Nullable)error __
...
+
+                // Workaround cases where macOS appends Contents/MacOS/<executable> to the bundle path which can happen in corner cases (eg: mishandled bundles or helper executables).
+                // https://github.com/sparkle-project/Sparkle/issues/2725
+                BOOL candidatePathIsDir = YES;
+                if ([fileManager fileExistsAtPath:candidatePath isDirectory:&candidatePathIsDir] && !candidatePathIsDir) {
+                    if (candidatePathComponentsCount > 3 && [candidatePathComponents[candidatePathComponentsCount - 3] isEqualToString:@"Contents"] && [candidatePathComponents[candidatePathComponentsCount - 2] isEqualToString:@"MacOS"]) {
+                        NSMutableArray<NSString *> *trimmedBundlePath = [candidatePathComponents mutableCopy];
+                        [trimmedBundlePath removeObjectsInRange:NSMakeRange(candidatePathComponentsCount - 3, 3)];
+                        candidatePathComponents = trimmedBundlePath;
+                        candidatePathComponentsCount -= 3;
```
zorgiepoo, line 249: Product > Analyze generates this warning: `Value stored to 'candidatePathComponentsCount' is never read`

This variable doesn't seem to be used after fileExistsAtPath: check

Unfortunately CI will not pass if an analyzer warning exists. An easy change is moving candidatePathComponentsCount declaration inside fileExistsAtPath check and removing this line that -= 3.
(Normally the bot would make a comment here but there's permission issues if someone creates a branch outside of the repo)

zorgiepoo: I see we also removed placing running app bundles that hit this case towards the end of the array, once we start doing a check for the bundle being a directory, simplifying the workaround. I'm alright with that.

zorgiepoo:
I further polished and merge this fix in #2747. Thanks!

PR #2748, opened 2025-08-09
Fix an issue while searching a cloneable file

bSr43:
Hi !

Thank you for this amazing project that I've been using for years now !
I'm the developer of a program called "Hopper Disassembler." I have been using an old version of Sparkle (1.27.1), which worked perfectly fine for years. My CI compiles the new versions of my application and builds 3 delta updates for each variant (demo, full, etc.). Everything was fine, until I upgraded to Sparkle 2.7.1 for my upcoming version 6.x.

When building delta updates between my old version (which uses Sparkle 1.27.1) and my new version (using Sparkle 2.7.1), BinaryDelta fails to create the patch between the two versions of the embedded Sparkle framework because a file named `SUUpdateAlert.nib` is compared to a directory with the same name and path (after the `.framework/Versions/{A -> B}` substitution)  in the old version.

## Fixes

The fixe is trivial: in the `cloneableRelativePath` function, in the part which handled the `.framework/Versions/{A -> B}` substitution, I check that the candidate is also a file.

## Misc Checklist

## Testing

- [ ] Sparkle Test App
- [ ] Unit Tests
- [X] My own app
- [ ] Other (please specify)

macOS version tested: 15.6

### DIFF
--- a/Autoupdate/SUBinaryDeltaCommon.m
+++ b/Autoupdate/SUBinaryDeltaCommon.m
@@ -115,9 +115,9 @@ uint16_t latestMinorVersionForMajorVersion(SUBinaryDeltaMajorVersion majorVersio
         case SUBinaryDeltaMajorVersion2:
             return 5;
         case SUBinaryDeltaMajorVersion3:
-            return 2;
+            return 3;
         case SUBinaryDeltaMajorVersion4:
-            return 1;
+            return 2;
     }
     return 0;
 }
--- a/Autoupdate/SUBinaryDeltaCreate.m
+++ b/Autoupdate/SUBinaryDeltaCreate.m
@@ -345,7 +345,12 @@ static NSString *cloneableRelativePath(NSDictionary<NSString *, NSData *> *after
         if (oldCloneInfo == nil) {
             continue;
         }
-        
+
+        // The old file must have the same type as the new one
+        if ([(NSNumber *)oldCloneInfo[INFO_TYPE_KEY] unsignedShortValue] != FTS_F) {
+            continue;
+        }
+
         NSNumber *newPermissions = newInfo[INFO_PERMISSIONS_KEY];
         if (outNewPermissions != NULL) {
             *outNewPermissions = newPermissions;


zorgiepoo:
Thank you. (I also have used Hopper plenty btw, great app!)

This fix looks good to me. I feel like we should add a check to this case too:

```
// Look out for any changes that involve the same named file moving to another directory
```

(Even though it's not strictly necessary because the old size will be 0 for a directory I think)

For these kind of fixes I also like to update `latestMinorVersionForMajorVersion()` -- in this case bump `SUBinaryDeltaMajorVersion4` case from `1` to `2` so we can tell if a generated patch file has this fix.

edit: The `SUBinaryDeltaMajorVersion3` case should be bumped from `2` to `3` too since it supports cloneable diffs.

Let me know if you are able to update this or not.

> When building delta updates between my old version (which uses Sparkle 1.27.1) and my new version (using Sparkle 2.7.1), BinaryDelta fails to create the patch between the two versions of the embedded Sparkle framework because a file named SUUpdateAlert.nib is compared to a directory with the same name and path (after the .framework/Versions/{A -> B} substitution) in the old version

In this specific case, please also check the [Compatibility](https://sparkle-project.org/documentation/delta-updates/#compatibility) section for delta updates. In this instance, an app with Sparkle 1.27.1 only understands how to apply version 2 delta patches (which do not support clones) so you'd pass `./BinaryDelta --version 2` in that case when creating a delta update. Then later use `--version 4` (current default) for later updates.

zorgiepoo:
Feel free to also pick up this patch for additional tests for this case:
https://github.com/sparkle-project/Sparkle/commit/d339afa053e8e29eb4229b4506f09208ae99ede0

Or I can land these myself afterwards.

bSr43:
Sorry for the very late response, I'm not at home at the moment. I have modified the minor versions numbers, as requested.

zorgiepoo (approved):

zorgiepoo:
Thank you, I think this looks good.

zorgiepoo:
@bSr43 Btw, I downloaded a new version Hopper to try out. If you run `defaults delete com.cryptic-apps.hopper-web-4` and on second launch, it seems like Sparkle's "Check for updates automatically?" permission prompt window gets put behind your main app window, with the sheet asking to enter a license. One way to fix that is to defer starting the updater until after that sheet is dismissed. Not sure how that'll behave if there's an update to install too, and if the issue is only present when a license isn't entered.

[2.8.0](https://github.com/sparkle-project/Sparkle/discussions/2761) (Tahoe refresh) and 2.7.3 (security fixes) were also released recently.

bSr43:
Thank you @zorgiepoo, I'll improve that point in the upcoming update.

PR #2789, opened 2025-10-20
Add missing localizations to zh-CN

francisfeng:
This PR adds missing localizations to simplified Chinese and also modifies some to better fit the modern context of macOS.


zorgiepoo:
Thank you.

PR #2826, opened 2026-01-02
Remove sparkle-cli (sparkle.app) from binary distribution

zorgiepoo:
External updater software usually (and often should) use `SPUUpdater` directly. sparkle-cli is a show case for this. Although the utility app is useful and software may want to rely on it, I think if they want to do that, they should also build/deploy with a custom `PRODUCT_BUNDLE_IDENTIFIER` that doesn't clash other client's usages.

Note this app is already not in the SPM distribution.

## Misc Checklist

## Testing

- [ ] Sparkle Test App
- [ ] Unit Tests
- [ ] My own app
- [x] Other (please specify)

Tested `make release` locally and verified sparkle.app is not in there.

macOS version tested: 26.2 (25C56)

### DIFF
--- a/Configurations/make-release-package.sh
+++ b/Configurations/make-release-package.sh
@@ -16,7 +16,6 @@ function verify_code_signatures() {
     find "${verification_directory}" -name "Sparkle.framework" -type d -exec codesign --verify -vvv --deep {} \;
     
     if [ "$check_aux_apps" = true ] ; then
-        codesign --verify -vvv --deep "${verification_directory}/sparkle.app"
         codesign --verify -vvv --deep "${verification_directory}/Sparkle Test App.app"
     fi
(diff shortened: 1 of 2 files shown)


sandermeis:
Unfortunately we relied on the external sparkle binary for updating our macos cli (cpp). Solved by installing xcode on our runners and building from source. Do you recommend that we now build this target ourselves instead of relying on a brew tap?

zorgiepoo:
Recommendations are what is described in the description of this PR + https://sparkle-project.org/documentation/sparkle-cli/ , namely to build sparkle-cli with a custom bundle identifier (not something we can distribute ourselves) so you don't clash with other folks using the tool, or to use Sparkle.framework directly if you are also distributing an app that also uses Sparkle for updating itself (thereby avoiding redundancy of bundling sparkle twice).

PR #2848, opened 2026-02-16
Appcast order deltas

### LINKED ISSUE #2846: In generated appcast, make delta order stable
## Summary

When updating an appcast with `generate_appcast`, the order of deltas (within each `<sparkle: deltas>` tag) sometimes changes.

As a result, producing a text diff of two versions of the appcast (say, to check if you've accidentally messed something up) produces a lot of irrelevant changes.

## Possible Fix

The deltas could be sorted by their `deltaFrom` value.
I'm willing to give the implementation a try, if this idea is accepted.

Cykelero:
Makes order of deltas stable in appcasts generated by generate_appcast.

Also (probably?) makes the insertion of new deltas thread-safe.

Fixes #2846.

- [X] My own app

Checked updating the Retcon appcast with the new code.

macOS version tested: 26.3

### DIFF
--- a/generate_appcast/Appcast.swift
+++ b/generate_appcast/Appcast.swift
@@ -349,13 +349,13 @@ func makeAppcasts(archivesSourceDir: URL, outputPathURL: URL?, cacheDirectory ca
                     delta = DeltaUpdate(fromVersion: item.version, archivePath: deltaPath, sparkleExecutableFileSize: item.sparkleExecutableFileSize, sparkleLocales: item.sparkleLocales)
                 }
 
-                latestItem.deltas.insert(delta, at: 0)
-
                 // Require delta to be a bit smaller
                 if delta.fileSize / 7 > latestItem.fileSize / 8 {
                     markDeltaAsIgnored(delta: delta, markerPath: ignoreMarkerPath)
                     continue
                 }
+                
+                latestItem.deltas.append(delta)
 
                 group.enter()
                 DispatchQueue.global().async {
--- a/generate_appcast/Appcast.swift
+++ b/generate_appcast/Appcast.swift
@@ -406,6 +406,21 @@ func makeAppcasts(archivesSourceDir: URL, outputPathURL: URL?, cacheDirectory ca
             throw signingError
         }
     }
+    
+    for appcast in appcastByFeed.values {
+        for archiveItem in appcast.archives.values {
+            archiveItem.deltas.sort {
+                guard
+                    let leftIntVersion = Int($0.fromVersion),
+                    let rightIntVersion = Int($1.fromVersion)
+                else {
+                    return $0.fromVersion > $1.fromVersion
+                }
+                
+                return leftIntVersion > rightIntVersion
+            }
+        }
+    }
 
     return appcastByFeed
 }
--- a/generate_appcast/Appcast.swift
+++ b/generate_appcast/Appcast.swift
@@ -227,6 +227,7 @@ func makeAppcasts(archivesSourceDir: URL, outputPathURL: URL?, cacheDirectory ca
             // but we still wanted to record the used delta updates for a batch of recent updates
             // This is to support rollback in case the top newly generated update isn't exactly what the user wants
             let generatingDeltas = latestVersionPerBranch.contains(version)
+            let itemDeltasLock = NSLock()
             var numDeltas = 0
             let appBaseName = latestItem.appPath.deletingPathExtension().lastPathComponent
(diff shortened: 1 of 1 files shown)


zorgiepoo:
So the update items are already sorted before doing the concurrent dispatch. It's kind of brute force approach to lose order by inserting delta updates in any order and then re-sort all of them again later by enumerating through all potential objects. Also the code cannot assume the versions are integers (see `descendingVersionComparator` usage earlier in the file).

A simpler approach I think is just to opportunistically add the delta item before the dispatch. Remove it if we need to mark the delta as ignored (by finding the firstIndex() and then remove()). That will have to be synchronized in the DispatchQueue.main.async. No need to re-sort again.

Cykelero:
Oh yeah, this is much nicer!

I ended up using `removeAll()` instead of `firstIndex()`+`remove()`.

I test on the Retcon appcast again, and also tried artificially making some deltas not have a signature; both the ordering and filtering seem to work.

generate_appcast/Appcast.swift
```diff
@@ -379,13 +381,17 @@ func makeAppcasts(archivesSourceDir: URL, outputPathURL: URL?, cacheDirectory ca
                         hasAnyDSASignature = hasAnyDSASignature || (delta.dsaSignature != nil)
 #endif
                         if hasAnyDSASignature {
-                            latestItem.deltas.append(delta)
+                            group.leave()
                         } else {
                             markDeltaAsIgnored(delta: delta, markerPath: ignoreMarkerPath)
                             print("Delta \(delta.archivePath.path) ignored, because it could not be signed")
+                            
+                            DispatchQueue.main.async {
```
zorgiepoo: One more thing I noticed here is I think we need to use a lock as synchronization,  rather than the main thread queue. The `group.wait()` is called on the main thread and this can likely be blocked from executing.

generate_appcast/Appcast.swift
```diff
@@ -349,6 +349,8 @@ func makeAppcasts(archivesSourceDir: URL, outputPathURL: URL?, cacheDirectory ca
                     delta = DeltaUpdate(fromVersion: item.version, archivePath: deltaPath, sparkleExecutableFileSize: item.sparkleExecutableFileSize, sparkleLocales: item.sparkleLocales)
                 }
 
+                latestItem.deltas.insert(delta, at: 0)
```
zorgiepoo: This needs to be after the "// Require delta to be a bit smaller" branch

zorgiepoo: Also I thought it should append the delta at the end of the list rather than the beginning. The first element of the deltas should be the greatest version, and we should be processing greater versions first. (The output in the XML should have the greatest from version listed first)

Cykelero: Weird, I did check the order and thought it was right. I changed the insert to `latestItem.deltas.append(delta)` and the ordering is indeed correct now, but please check the behavior if you can, in case there's something else unexpected going on.

Cykelero:
OK, changes applied.

generate_appcast/Appcast.swift
```diff
@@ -379,13 +382,17 @@ func makeAppcasts(archivesSourceDir: URL, outputPathURL: URL?, cacheDirectory ca
...
                             markDeltaAsIgnored(delta: delta, markerPath: ignoreMarkerPath)
                             print("Delta \(delta.archivePath.path) ignored, because it could not be signed")
+                            
+                            itemDeltasLock.lock()
+                            latestItem.deltas.removeAll { $0 === delta }
+                            group.leave()
+                            itemDeltasLock.unlock()
                         }
                     }
-                    group.leave()
```
zorgiepoo, line 388: This can go back where it was now I believe. I'd also prefer if the group.leave was after the unlock call anyway (though I don't think it matters).

generate_appcast/Appcast.swift
```diff
@@ -354,6 +355,8 @@ func makeAppcasts(archivesSourceDir: URL, outputPathURL: URL?, cacheDirectory ca
                     markDeltaAsIgnored(delta: delta, markerPath: ignoreMarkerPath)
                     continue
                 }
+                
+                latestItem.deltas.append(delta)
```
zorgiepoo, line 360: We need a lock/unlock around this

zorgiepoo:
Works for me locally. Merging. Thanks!

Cykelero:
Great! Thank you for making these final changes, I was a bit busy.

PR #2864, opened 2026-03-23
Configurations: Add SKIP_INSTALL=YES to for XPCServices targets

jeremyhu:
In most scenarios, they were skipped because INSTALL_PATH is unset.

In XQuartz, we build Sparkle using `xcodebuild install INSTALL_PATH=...`, so these XPCServices would get installed into the frameworks path of our app bundle.  This change fixes that by explicitly setting SKIP_INSTALL.


zorgiepoo:
Thanks!

PR #2875, opened 2026-05-14
Polish Spanish localization in Sparkle.strings

perez987:
A few minor updates to the Spanish localization:

- Gender agreement: 'Ready to Install' translation adjusted to feminine 'Lista para instalarse'
- Verb tense improvement: '¿Quisieras' -> '¿Querrías' (two strings); 'corriendo -> 'ejecutando'
- Accent mark added: 'sólo' (read only)
- Only one string addresses the user formally; the rest use informal 'tú'. It is proposed to change the formal to informal address so that all strings use the same form of address: 'póngase en contacto' -> 'ponte en contacto'.

No English source keys were modified.

Thanks for your work!

## Testing

- [ x] My own app

macOS version tested: [macOS 26.5 Tahoe]

### DIFF
--- a/Sparkle/es.lproj/Sparkle.strings
+++ b/Sparkle/es.lproj/Sparkle.strings
@@ -2,13 +2,13 @@
 "%@ %@ is currently the newest version available." = "%1$@ %2$@ es la versión más nueva disponible.";
 
 /* No comment provided by engineer. */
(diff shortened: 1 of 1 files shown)


zorgiepoo:
Thank you.

PR #2879, opened 2026-05-21
Introduce a timing buffer for the Check for Updates flow to avoid jarring UI.

### LINKED ISSUE #2878: Visual flicker/noise when progress panel only appears for a fraction of a second
## Summary

The "Checking for updates…" window briefly flashes onto the screen before being immediately replaced by the subsequent panel (update found, no update, or an error). The flicker is most reproducible by pointing `SUFeedURL` at a garbage URL: the network failure comes back so quickly that the checking window is visible for only a frame or two before being torn down to make way for the error alert. Even with a valid feed, a response on a fast network connection can lead to the same problem.

https://github.com/user-attachments/assets/a5b0f5bc-b204-4d52-9f22-01f56aa25677

## Possible Fix

Pull request forthcoming.

## Version

Sparkle 2.x branch.

danielpunkass:
When a check completes or fails very quickly, the "Checking for Updates" window flashes briefly before being replaced by the next alert.

This PR introduces a buffering mechanism whereby SPUStandardUserDriver defers presenting the progress window for 0.3 seconds. If a result is obtained within that time frame, the progress window is never displayed.

When and if the checking window has actually appeared on screen, a minimum visible duration of 0.7 seconds must elapse before the panel is dismissed.

Fixes #2878 

## Misc Checklist

## Testing

- [ ] Sparkle Test App
- [ ] Unit Tests
- [x] My own app
- [ ] Other (please specify)

To replicate/confirm the various states for this UI flow, you can edit the SUCheckingWindowShowDelay and SUCheckingWindowMinDisplayTime const variables in SPUStandardUserDriver.m.

macOS version tested: macOS 26.5 (25F71)

### DIFF
--- a/Sparkle/SPUStandardUserDriver.m
+++ b/Sparkle/SPUStandardUserDriver.m
@@ -725,6 +725,9 @@ static const NSTimeInterval SUCheckingWindowMinDisplayTime = 0.7;
         if (strongSelf != nil) {
             [strongSelf _proceedWithUpdaterError:error acknowledgement:acknowledgement];
         }
+        else {
+            acknowledgement();
+        }
     } cancellation:^{
         acknowledgement();
     }];
--- a/Sparkle/SPUStandardUserDriver.m
+++ b/Sparkle/SPUStandardUserDriver.m
@@ -41,14 +41,6 @@
 @end
 #endif
 
-// Buffer the "Checking for updates…" window so it doesn't flicker when a check
-// completes (or fails) almost immediately. We suppress the window entirely if a
-// result comes back inside SUCheckingWindowShowDelay seconds, and we enforce a
-// minimum visible duration of SUCheckingWindowMinDisplayTime seconds once it
-// has actually appeared on screen.
-static const NSTimeInterval SUCheckingWindowShowDelay = 0.3;
-static const NSTimeInterval SUCheckingWindowMinDisplayTime = 0.7;
-
 @interface SPUStandardUserDriver () <SPUGentleUserDriverReminders>
 
 // Note: we expose a private interface for activeUpdateAlert property in SPUStandardUserDriver+Private.h as NSWindowController
@@ -85,15 +77,11 @@ static const NSTimeInterval SUCheckingWindowMinDisplayTime = 0.7;
     uint64_t _expectedContentLength;
     uint64_t _bytesDownloaded;
     double _timeSinceOpportuneUpdateNotice;
-    double _checkingWindowShownTime;
-    void (^_pendingPostCheckingCompletion)(void);
-    void (^_pendingPostCheckingCancellation)(void);
-    
+
     BOOL _updateAlertWindowWasInactive;
     BOOL _loggedGentleUpdateReminderWarning;
     BOOL _regularApplicationUpdate;
     BOOL _updateReceivedUserAttention;
-    BOOL _checkingWindowPendingShow;
 }
 
 @synthesize activeUpdateAlert = _activeUpdateAlert;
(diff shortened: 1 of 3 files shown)


zorgiepoo:
I do wonder a bit if this behavior should be UI driver agnostic or not. In other words, can we imagine any type of custom UI where it's beneficial to not have this sort of timing buffer logic?

For example `sparkle-cli --check-immediately` always prints "Checking for Updates...\n" immediately before "No new update available!\n" against a bundle that's up to date, but it'd likely be nice for it to have time-buffered logic too (so Checking for Updates is not printed if check is very fast).

Other custom UIs may use the same UI element (button, window, ..) for both checking and show update states.

On the other hand, some custom UI implementors may have done their own time buffering deferral and may have to think about compatibility.

zorgiepoo:
Well I didn't have to go far to find an app with custom UI that a timing buffer deferral approach would _not_ be suitable, so scratch that.

https://github.com/user-attachments/assets/7c42b889-6604-411f-b606-b56332fb6b98

Sparkle/SPUStandardUserDriver.m
```diff
@@ -477,6 +505,7 @@ - (void)showUpdateInFocus
         [_statusController showWindow:nil];
         mayNeedToActivateApp = YES;
     } else if (_checkingController != nil) {
+        [self _presentCheckingWindowIfPending];
         [_checkingController showWindow:nil];
```
zorgiepoo, line 497: This can call showWindow: twice potentially, you can add a return value on _presentCheckingWindowIfPending and only call showWindow: if that fails.

Sparkle/SPUStandardUserDriver.m
```diff
@@ -568,10 +620,92 @@ - (void)closeCheckingWindow SPU_OBJC_DIRECT
         _checkingController = nil;
         _cancellation = nil;
     }
+    _checkingWindowPendingShow = NO;
+    _checkingWindowShownTime = 0.0;
+    _pendingPostCheckingCompletion = nil;
+    _pendingPostCheckingCancellation = nil;
+}
+
+- (BOOL)_finishPendingPostCheckingTransition SPU_OBJC_DIRECT
```
zorgiepoo: Return value is unused from caller.

Sparkle/SPUStandardUserDriver.m
zorgiepoo: Actually I think this code is wrong. The point of `showUpdateInFocus` is to show UI elements back in focus. This shouldn't be used to bypass the timing buffer if the user spams checking for updates within the 0.3 seconds. So I believe we should show the window if `!_checkingWindowPendingShow` and otherwise let the transition handle it on its own. Then `_presentCheckingWindowIfPending` no longer needs to be its own method.

Additionally `_activeUpdateAlert` can now be non-nil and not shown while `_checkingController` is non-nil so I think you also have to move the `_checkingController != nil` check before the `_activeUpdateAlert` one.

Sparkle/SPUStandardUserDriver.m
zorgiepoo: Actually this method is only used in one place and probably shouldn't be a method on its own. It's not helpful to separate it out. Also `_pendingPostCheckingCompletion` probably is not needed to store as an ivar.

Sparkle/SPUStandardUserDriver.m
```diff
@@ -568,10 +620,92 @@ - (void)closeCheckingWindow SPU_OBJC_DIRECT
...
+    }
+
+    _pendingPostCheckingCompletion = nil;
+    _pendingPostCheckingCancellation = nil;
+    [self closeCheckingWindow];
+    pendingCompletion();
+    return YES;
+}
+
+- (BOOL)_cancelPendingPostCheckingTransition SPU_OBJC_DIRECT
```
zorgiepoo: Only used from one place, remove the method.

Sparkle/SPUStandardUserDriver.m
```diff
@@ -585,8 +719,22 @@ - (void)showUpdaterError:(NSError *)error acknowledgement:(void (^)(void))acknow
...
     assert(NSThread.isMainThread);
     
-    [self closeCheckingWindow];
+    __weak __typeof__(self) weakSelf = self;
+    [self _transitionFromCheckingWindowWithCompletion:^{
+        __typeof__(self) strongSelf = weakSelf;
+        if (strongSelf != nil) {
+            [strongSelf _proceedWithUpdaterError:error acknowledgement:acknowledgement];
+        }
+        else {
```
zorgiepoo: Brace style should be } else {
(ditto in showUpdateNotFoundWithError:acknowledgement:)

Sparkle/SPUStandardUserDriver.m
danielpunkass: Thanks I'll take a closer look!

danielpunkass: I've update the PR with a pretty substantial overhaul to the approach, moving the timing buffering into the SUStatusController. This seemed to make more sense both to avoid over-complicating things at the driver level, and to ensure that the same buffering is applied for the status/install-progress controller window too.

It got a bit more complicated than I wished but hopefully it'll make sense. It seems to work well in my tests.

Sparkle/SPUStandardUserDriver.m
danielpunkass: Superseded by new approach.

Sparkle/SPUStandardUserDriver.m
danielpunkass: Superseded by new approach.

Sparkle/SPUStandardUserDriver.m
```diff
@@ -352,12 +352,37 @@ - (void)showUpdateFoundWithAppcastItem:(SUAppcastItem *)appcastItem state:(SPUUs
...
         SULog(SULogLevelError, @"Error: -[%@ %@] should not be called when _activeUpdateAlert != nil:\n%@", NSStringFromClass([self class]), NSStringFromSelector(_cmd), NSThread.callStackSymbols);
     }
     
+    _regularApplicationUpdate = [appcastItem.installationType isEqualToString:SPUInstallationTypeApplication];
+
+    // For user initiated checks, let the delegate know we'll be showing an update.
+    // For scheduled checks, -setUpActiveUpdateAlertForUpdate:state: below will handle this.
+    id<SPUStandardUserDriverDelegate> delegate = _delegate;
+    if (state.userInitiated && [delegate respondsToSelector:@selector(standardUserDriverWillHandleShowingUpdate:forUpdate:state:)]) {
+        [delegate standardUserDriverWillHandleShowingUpdate:YES forUpdate:appcastItem state:state];
```
zorgiepoo: I think this should be after the checking window is closed (inside the completion block below)

zorgiepoo:
I believe showUpdateInFocus needs to be able to show the status window immediately when _retryTerminatingApplication != nil. An edge case I can think of is when you "Install and Relaunch" (which closes the window) but the app cancels the quit request, and the user checks for updates again and can retry that request.

Similarly the status window controller is used from the Updater.app (installer progress agent). That is already on a delayed timing buffer (~0.7 seconds `SUDisplayProgressTimeDelay`). This change would increase it by another 0.3 seconds, but it shouldn't.

I will probably need to look at this PR multiple times.

Sparkle/SPUStandardUserDriver.m
danielpunkass: Good catch, fixing.

danielpunkass:
> I believe showUpdateInFocus needs to be able to show the status window immediately when _retryTerminatingApplication != nil. An edge case I can think of is when you "Install and Relaunch" (which closes the window) but the app cancels the quit request, and the user checks for updates again and can retry that request.
> 
> Similarly the status window controller is used from the Updater.app (installer progress agent). That is already on a delayed timing buffer (~0.7 seconds `SUDisplayProgressTimeDelay`). This change would increase it by another 0.3 seconds, but it shouldn't.

My comparative lack of understanding of Sparkle nuances is showing! Sorry about missing the subtleties. I'll take a closer look again. 

> I will probably need to look at this PR multiple times.

Understandable, and if you think the gain is not worth the effort I totally get it.

danielpunkass:
I had missed that the progress agent is imposing its own display delay, and it seems like the best solution would be if it can leverage the same internal  buffering that `SUStatusController` would now have. Either the buffering delay/decay could be configurable, or the Updater.app could accept the default timing imposed by the `SUStatusController`

Sparkle/SPUStandardUserDriver.m
danielpunkass: Updated with 8531e1924f757495295da07ec6fb5ef6a69cbd07

zorgiepoo:
The delay in the installation side is intertwined with IPC from Autoupdate between the host app and the progress agent, so I think it'd be less risky to have a way to invoke an immediate way of showing the window, which would be needed for the _retryTerminatingApplication != nil path as well.

danielpunkass:
Does it seem like parameterizing the delay constant would be suitable? Then any client who needs it immediately could set the delay to 0?

zorgiepoo:
Sounds logical.

zorgiepoo:
Another thing I'm noticing is when I click Install from the release notes window in the Test App, there is a noticeable delay until the status window comes front, so that should likely be immediate.

I am not actually sure the newer approach is better than the older one you proposed (which I have a little trouble recalling at the moment). The older one may have been touching less things by not changing logic within the status controller and I think I just had a couple of comments but otherwise understood it.

zorgiepoo:
I'm going to tackle this in https://github.com/sparkle-project/Sparkle/pull/2919 I think.

It's a lot simpler to not handle the post / minimum duration delay of the checking window and you get most of the benefit by handling the first half. The post delay could also make Sparkle be perceived as slower than it is.

danielpunkass:
I totally dropped the ball on this and never came back to it, so if you can get the equivalent done in your other work that would be amazing.

PR #2883, opened 2026-06-09
Bump minimum deployment target to 12.0

zorgiepoo:
To support building Sparkle in Xcode 27, we need to bump the deployment target to 12.0

## Misc Checklist

## Testing

- [x] Sparkle Test App
- [ ] Unit Tests
- [ ] My own app
- [ ] Other (please specify)

Tested building Sparkle with:
Xcode 27 beta 1
Xcode 26.5 (17F42)

macOS version tested: 26.5 (25F71)


Eitot:
The [release notes mention](https://developer.apple.com/documentation/xcode-release-notes/xcode-27-release-notes#:~:text=The%20minimum%20supported%20deployment%20target%20on%20macOS%20has%20been%20increased%20to%2011.0.) (albeit in the wrong subsection) that the minimum deployment target is macOS 11. It may be worth keeping an eye on this in case this is just an error in the beta release.

lemkesoft:
Can't we stick with macOS 10.13 as the minimum requirement? Many users are still running older versions of macOS because their computers can no longer be updated.

zorgiepoo:
Updated the PR description. Without this change, the project does not build because the deployment target minimum was bumped. You can try for yourself by downloading Xcode 27 beta.

Pretty soon developers that build Sparkle as a subproject dependency (rather than as a precompiled binary or from a supported package manager) are going to ask for this because it will block them.

I will merge this PR into 2.x branch soon. Developers as usual can use a tagged release of Sparkle which supports 10.13 for now. I probably won't do a non-beta tagged release until macOS 27 is closer to shipping. From now until then we can see if anything changes in Xcode and we can revise the minimum support if so.

> The [release notes mention](https://developer.apple.com/documentation/xcode-release-notes/xcode-27-release-notes#:~:text=The%20minimum%20supported%20deployment%20target%20on%20macOS%20has%20been%20increased%20to%2011.0.) (albeit in the wrong subsection) that the minimum deployment target is macOS 11. It may be worth keeping an eye on this in case this is just an error in the beta release.

I think that release note mention is specific to the C++ section actually, rather than the overall minimum deployment target.

PR #2884, opened 2026-06-14
Modernise the status dialog

schriftgestalt:
Adapt the status dialog to the style of Big Sur alert dialogs.
Bildschirmfoto 2026-06-14 um 10 24 22

All macOS versions before Bug Sur get the original UI.


zorgiepoo:
I don't think I am a fan compared the current UI. Big Sur is also superseded by Tahoe where things aren't as center aligned as much. We should also not maintain a legacy nib, just use a UI that also scales to older OSs (and OS versions before 12.0 may get dropped in the next major version anyway).

schriftgestalt:
Alert with one line of text are stil getting centered layout. 

And left aligning doesn’t look better here:
Bildschirmfoto 2026-06-14 um 20 42 55

zorgiepoo:
I like the current UI better. If you propose major changes subject to controversy you should open a discussion first. This PR also skipped the pull request template (which I figure I should update now).

PR #2907, opened 2026-09-01
Handle .service bundle identifiers in cache paths

youngjunning:
## Summary

Sparkle already appends `.sparkle` when a bundle identifier ends in `.app`, preventing the cache directory from being treated as an application bundle. macOS also recognizes `.service` as a Service Application bundle, so identifiers ending in `.service` need the same neutral cache suffix.

This change normalizes the final identifier extension case-insensitively, appends `.sparkle` for `.app` and `.service`, and preserves ordinary bundle identifiers. It also adds focused regression tests for both package-like extensions and unaffected identifiers.

This change was prepared with Codex assistance. I reviewed the complete diff and ran Sparkle's official unit-test workflow locally.

Related discussion: https://github.com/sparkle-project/Sparkle/discussions/2881

## Misc Checklist

- [x] My change was generated/assisted using AI and if so was reviewed by me in whole (explained in the summary)

## Testing

- [ ] Sparkle Test App
- [x] Unit Tests
- [x] My own app
- [x] Other: Sparkle's CI `build-for-testing` and `test-without-building` commands

The regression tests cover `.app` and `.service` suffixes with mixed casing, plus ordinary identifiers that must remain unchanged. The full `Distribution` unit-test suite passed with 183 tests, 0 failures, and 0 skipped tests.

macOS version tested: 26.5.2 (Xcode 26.6)

### DIFF
--- a/Sparkle/SPULocalCacheDirectory.m
+++ b/Sparkle/SPULocalCacheDirectory.m
@@ -18,15 +18,7 @@ static NSTimeInterval OLD_ITEM_DELETION_INTERVAL = 86400 * 10; // 10 days
 
 + (NSString *)_cachePathForCacheDirectory:(NSURL *)cacheURL bundleIdentifier:(NSString *)bundleIdentifier SPU_OBJC_DIRECT
 {
-    // If an app has an ill-formed bundle identifier that ends with ".app" we will append ".sparkle" to it
-    // so that the cache directory doesn't look like an app bundle directory, which can cause other systematic issues
-    // https://github.com/sparkle-project/Sparkle/discussions/2881
-    NSString *appCacheIdentifier;
-    if ([bundleIdentifier hasSuffix:@".app"] || [bundleIdentifier hasSuffix:@".APP"]) {
-        appCacheIdentifier = [bundleIdentifier stringByAppendingString:@".sparkle"];
-    } else {
-        appCacheIdentifier = bundleIdentifier;
-    }
+    NSString *appCacheIdentifier = SPUCacheIdentifierForBundleIdentifier(bundleIdentifier);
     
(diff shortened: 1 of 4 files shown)


Sparkle/SPULocalCacheDirectory.m
```diff
@@ -18,15 +18,7 @@ @implementation SPULocalCacheDirectory
 
 + (NSString *)_cachePathForCacheDirectory:(NSURL *)cacheURL bundleIdentifier:(NSString *)bundleIdentifier SPU_OBJC_DIRECT
 {
-    // If an app has an ill-formed bundle identifier that ends with ".app" we will append ".sparkle" to it
-    // so that the cache directory doesn't look like an app bundle directory, which can cause other systematic issues
-    // https://github.com/sparkle-project/Sparkle/discussions/2881
-    NSString *appCacheIdentifier;
-    if ([bundleIdentifier hasSuffix:@".app"] || [bundleIdentifier hasSuffix:@".APP"]) {
```
zorgiepoo, line 25: Please just add the one/two more `||` conditions here for `.service`. The rest of this pull request is not needed, including the unit tests which I don't want. This is working around a macOS bug in a specific context.

zorgiepoo:
Does your own app actually have a bundle identifier that ends with `.service` and did you reproduce the issue while running the updater? (The unit tests don't help validate this). I presume this to be a much rarer case.

Sparkle/SPULocalCacheDirectory.m
zorgiepoo: Feel free to use `caseInsensitiveCompare:` on the bundleIdentifier.pathExtension in here to simplify testing for case.

zorgiepoo:
Confirmed locally that .service is indeed problematic, at least on macOS 26 (I seem to not reproduce on my macOS 27 beta setup -- not sure if they fixed it). I will address this in #2909 along with other bundle-like candidates just in case.. Thank you for bringing this up.

PR #2912, opened 2026-09-07
Update European Portuguese localization

h3x4d3x4:
The pt-PT strings are spelled to the pre-1990 Orthographic Agreement — `actualização`, `Está actualizado!`. Portugal adopted the Agreement in 2009 and it has been mandatory in official and technical writing since 2015, so the update dialog reads dated to a Portuguese user. This updates all 13 affected strings.

While in there, a few things that changed what the strings said rather than how they were spelled:

- **`"An error occurred while parsing the update feed."`** — "**Ocorrer** um erro" is the infinitive where the preterite belongs. The other three error strings in the file already say "Ocorreu".
- **Two error strings** said "Por favor **novamente tente** mais tarde" — wrong word order. The other two say "tente novamente"; now all four agree.
- **`"A new version of %@ is available!"`** — "está **dísponível**", a stray accent.
- **`"…has been downloaded and is ready to use! Would you like to install it and relaunch %1$@ now?"`** — the translation asked whether to install now and relaunch **`posteriormente`** (*later*). The English asks to relaunch now, and the button beside it says "Instalar e reiniciar", so the question contradicted both.
- **`"%1$@ can’t be updated because it was opened from a read-only or a temporary location."`** — the translation still carried the older English it replaced, describing disk images and optical discs ("como uma imagem de disco ou disco óptico"). It now says what the current string says.
- **`"Should %1$@ automatically check for updates?…"`** — "procurar **por** atualizações" is an anglicism, and inconsistent with the same sentence's second half, which already says "procurar atualizações".
- **Anonymous system profile disclosure** — "usada para **no futuro** nos ajudar a planear o trabalho de desenvolvimento" attaches *future* to the helping. In the English it qualifies the work: "…a planear o trabalho de desenvolvimento **futuro**".

I deliberately left `óptico` out of the orthography pass — it is a double-spelling case under the Agreement — but the string it appeared in was retranslated anyway, so the word is gone.

All 38 keys, their order, and every format specifier (`%1$@`, `%2$@`, `\n`) are unchanged; `plutil -lint` passes. Found while shipping a pt-PT app that uses Sparkle.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01159teNEMecUjsH1dB8opDd

### DIFF
--- a/Sparkle/InstallerProgress/SPUInstallerProgressLocalizedStrings.h
+++ b/Sparkle/InstallerProgress/SPUInstallerProgressLocalizedStrings.h
@@ -122,9 +122,9 @@ static NSDictionary<NSString *, NSDictionary<NSString *, NSString *> *> *const S
         @"Updating %@": @"Atualizando o app %@",
     },
     @"pt-PT": @{
-        @"Cancel Update": @"Cancelar actualização",
-        @"Installing update…": @"A instalar actualização…",
-        @"Updating %@": @"A actualizar o %@",
+        @"Cancel Update": @"Cancelar atualização",
+        @"Installing update…": @"A instalar atualização…",
+        @"Updating %@": @"A atualizar o %@",
     },
     @"ro": @{
         @"Cancel Update": @"Anulează actualizarea",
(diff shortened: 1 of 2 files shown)


zorgiepoo:
Thank you. I will try closing and re-opening the pull request because the GitHub actions didn't run, maybe due to the initial wrong target branch.

zorgiepoo:
Can you run `./Configurations/generate_progress_tool_localizations.py` and commit the result?

Issue #2917, opened 2026-09-12
New Style for Update Permission Prompt


zorgiepoo:
### Discussed in https://github.com/sparkle-project/Sparkle/discussions/2916

<div type='discussions-op-text'>

<sup>Originally posted by **Anderson4366** September 12, 2026</sup>
The update permission prompt still uses the horizontal layout of NSAlerts from the days before Big Sur.

That makes it fall out of line with the prompts the system and other apps throw at users all the time, like TCC permission prompts.

I know this dialog deliberately isn't a system alert since it would block the main thread, as mentioned in commit 3a201a45.

I also read #2731, where you said you prefer these windows not presenting like NSAlerts, but also that you would like a vertical top to bottom layout with the app icon as part of the stack rather than beside everything else.

So my question is: does that still hold for this prompt, or would a change to bring it in line with modern alerts be welcome?

I have a reworked version running that resembles a Tahoe alert, while keeping the behavior identical, same options, same responses, no new strings.

<img width="412" height="454" alt="Bildschirmfoto 2026-09-05 um 16 02 28" src="https://github.com/user-attachments/assets/82e925e4-6075-41ba-96b9-8099359cbba8" />
<img width="412" height="655" alt="Bildschirmfoto 2026-09-05 um 16 02 31" src="https://github.com/user-attachments/assets/f08dc08d-eb5f-40c0-831c-ab94db821221" />
<img width="549" height="329" alt="Bildschirmfoto 2026-09-05 um 16 05 14" src="https://github.com/user-attachments/assets/942e271d-9667-4d95-b3a7-0559c0f037a4" />
<img width="549" height="534" alt="Bildschirmfoto 2026-09-05 um 16 05 16" src="https://github.com/user-attachments/assets/bd43af92-2f92-43e4-b060-0f70376cbe2f" />
</div>

zorgiepoo:
I don't entirely like the current window style so turning this into an issue for potential improvements.

Your proposed screenshot is not too bad. Is the background gray just because you took a screenshot of the window (without the background/desktop)?

I don't think I've seen a section expansion button on the right side of a control (for include anonymous system profile) anywhere else, so that might be strange. I wonder if an info button would be more adequate.

Anderson4366:
Yes, the window appears gray because it was captured solo, without the desktop behind it. It uses the macOS 26+ NSGlassEffectView and looks exactly like the other system alerts.

You are right regarding the strange positioning of the expansion button.

[...]

PR #2918, opened 2026-09-12
Changed Update Permission Prompt to resemble Tahoe Alert

Anderson4366:
Opening this as suggested in #2917 so the approach can be tried out.

The update permission prompt still used the horizontal layout that alerts had before Big Sur, which made it stand out next to the prompts the system and other apps show users today.

It now follows the metrics of a system prompt, 296 points wide with a 64 point app icon on top, the question and description below it, the choices next, and the two response buttons stacked at the bottom. The width and the 26 point corner radius are measured off a system prompt. On macOS 26 the buttons use the large control size and the dialog is drawn on an NSGlassEffectView, since AppKit offers no way to round a window itself. Before macOS 26 it stays a regular window with centered icon and text.

Behavior is identical, same options, same responses, same delegate paths, no new strings. All sections now live inside the window's stack view instead of being added programmatically, so the dialog shows up in Interface Builder again, while hidden sections are still detached by the stack view. Checkbox titles wrap now, which lets the dialog keep a fixed width without truncating longer translations, Ukrainian being the worst case at two lines.

Two things worth pointing out: the corner radius is a measured constant since AppKit does not expose the corner radius of a window, and the wrapping labels get their preferredMaxLayoutWidth set after the first layout pass, because the checkbox width is only known by then. Both are commented in the code.

I left the table view as it is for now so this stays a layout change. This is meant as something to try out rather than a finished proposal, so I'm happy to adjust the details as well as the code style.

Addresses #2917

## Misc Checklist

Used Claude Code for the implementation. I made the design decisions and took the measurements against real alerts and system prompts. I also reviewed and tested every change in full, including edge cases like the Ukrainian localization.

## Testing

- [x] Sparkle Test App
- [ ] Unit Tests
- [ ] My own app
- [ ] Other (please specify)

Tested in English, German and Ukrainian where the long checkbox titles wrap to a second line without truncating and the dialog keeps its width. Checked the system profile disclosure expanding and collapsing, keyboard handling for Return and Escape, and Reduce Transparency. No constraint conflicts were logged in any of those runs.

macOS version tested: 26.5.2

### DIFF
--- a/Sparkle/SUUpdatePermissionPrompt.m
+++ b/Sparkle/SUUpdatePermissionPrompt.m
@@ -20,6 +20,8 @@
 
 static NSString *const SUUpdatePermissionPromptTouchBarIdentifier = @"" SPARKLE_BUNDLE_IDENTIFIER ".SUUpdatePermissionPrompt";
 
+static const CGFloat SUUpdatePermissionPromptGroupElementSpacing = 12.0;
+
 @interface SUUpdatePermissionPrompt () <NSTouchBarDelegate>
 
 // These properties are used for bindings
@@ -35,14 +37,16 @@ static NSString *const SUUpdatePermissionPromptTouchBarIdentifier = @"" SPARKLE_
     
     IBOutlet NSStackView *_stackView;
     IBOutlet NSView *_promptView;
+    IBOutlet NSImageView *_applicationIconImageView;
+    IBOutlet NSLayoutConstraint *_applicationIconLeadingLayoutConstraint;
     IBOutlet NSView *_moreInfoView;
     IBOutlet NSView *_placeholderView;
-    IBOutlet NSView *_responseView;
     IBOutlet NSView *_infoChoiceView;
     IBOutlet NSView *_automaticallyDownloadUpdatesView;
     IBOutlet NSButton *_cancelButton;
     IBOutlet NSButton *_checkButton;
     IBOutlet NSTextField *_checkForUpdatesAutomaticallyTextField;
+    IBOutlet NSTextField *_promptDescriptionTextField;
     IBOutlet NSButton *_includeAnonymousSystemProfileButton;
     IBOutlet NSButton *_anonymousInfoDisclosureButton;
     IBOutlet NSButton *_automaticallyDownloadAndInstallUpdatesButton;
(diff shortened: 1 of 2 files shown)


zorgiepoo:
xib

The xib has issues in interface builder (it just needs to be re-saved probably).

This change makes the stack view in the xib instead of setting adding the arranged views in code. The issue is that the views are kind of a mess to edit in Interface Builder with all overlapping each other.

Sparkle/SUUpdatePermissionPrompt.m
```diff
@@ -88,28 +92,101 @@ - (NSString *)description { return [NSString stringWithFormat:@"%@ <%@>", [self
...
+    window.titleVisibility = NSWindowTitleHidden;
+
+    // The dialog has a fixed width, so let long localized checkbox titles wrap instead of being
+    // truncated. Note button cells only wrap when their width is constrained, which it is here.
+    for (NSButton *checkbox in @[_automaticallyDownloadAndInstallUpdatesButton, _includeAnonymousSystemProfileButton]) {
+        NSCell *cell = checkbox.cell;
+        cell.usesSingleLineMode = NO;
+        cell.lineBreakMode = NSLineBreakByWordWrapping;
+        cell.wraps = YES;
+    }
```
zorgiepoo, line 120: I feel like the window should maybe be wide enough so these two options don't wrap in any language (I wonder if that's the case with the current permission prompt window). Matching the Tahoe alert exactly is not really my goal.

Sparkle/SUUpdatePermissionPrompt.m
```diff
@@ -88,28 +92,101 @@ - (NSString *)description { return [NSString stringWithFormat:@"%@ <%@>", [self
...
+    if (@available(macOS 26, *)) {
+        _cancelButton.controlSize = NSControlSizeLarge;
+        _checkButton.controlSize = NSControlSizeLarge;
+    } else {
+        // Alerts center their icon and text before macOS 26
+        _applicationIconLeadingLayoutConstraint.active = NO;
+        [_applicationIconImageView.centerXAnchor constraintEqualToAnchor:_promptView.centerXAnchor].active = YES;
+
+        _checkForUpdatesAutomaticallyTextField.alignment = NSTextAlignmentCenter;
+        _promptDescriptionTextField.alignment = NSTextAlignmentCenter;
```
zorgiepoo, line 131: Not sure I want to make these center aligned on earlier OS's. We don't do that kind of alignment treatment anywhere else.

Sparkle/SUUpdatePermissionPrompt.m
```diff
@@ -88,28 +92,101 @@ - (NSString *)description { return [NSString stringWithFormat:@"%@ <%@>", [self
...
+    }
+
     _checkButton.title = SULocalizedStringFromTableInBundle(@"Check Automatically", SPARKLE_TABLE, sparkleBundle, nil);
     _cancelButton.title = SULocalizedStringFromTableInBundle(@"Don’t Check", SPARKLE_TABLE, sparkleBundle, nil);
     _checkForUpdatesAutomaticallyTextField.stringValue = SULocalizedStringFromTableInBundle(@"Check for updates automatically?", SPARKLE_TABLE, sparkleBundle, nil);
     _includeAnonymousSystemProfileButton.title = SULocalizedStringFromTableInBundle(@"Include anonymous system profile", SPARKLE_TABLE, sparkleBundle, nil);
     _automaticallyDownloadAndInstallUpdatesButton.title = SULocalizedStringFromTableInBundle(@"Automatically download and install updates", SPARKLE_TABLE, sparkleBundle, nil);
     _anonymousSystemProfileDisclosureInformation.stringValue = SULocalizedStringFromTableInBundle(@"Anonymous system profile information is used to help us plan future development work. Please contact us if you have any questions about this.\n\nThis is the information that would be sent:", SPARKLE_TABLE, sparkleBundle, nil);
+
```
zorgiepoo, line 140: All the custom layout changes and hardcoded glass radius below here I'm kind of risk adverse to wanting.

Sparkle/SUUpdatePermissionPrompt.m
zorgiepoo: ..but maybe the window will look worse if it's wider, I don't know. So maybe that's a bad idea.

Sparkle/SUUpdatePermissionPrompt.m
zorgiepoo: The app icon maybe should be center aligned, but the text fields look odd to me. I will need to test this on an older OS.

Sparkle/SUUpdatePermissionPrompt.m
zorgiepoo: Is there any locale/language where the alert looks differently if you comment out the preferredMaxLayoutWidth setting and layoutIfNeeded calls?

Sparkle/SUUpdatePermissionPrompt.m
```diff
@@ -88,28 +92,101 @@ - (NSString *)description { return [NSString stringWithFormat:@"%@ <%@>", [self
...
+// Clipping the glass view is what makes the corners work, without it the content is drawn as a
+// rectangle over the rounded glass. The window keeps its title bar style mask so that it can still
+// become the key window, which also means no window subclass is needed. Before macOS 26 the dialog
+// stays a regular window.
+- (void)_embedContentInGlassBackgroundForWindow:(NSWindow *)window
+{
+    if (@available(macOS 26, *)) {
+        // AppKit does not expose the corner radius of a window, and the layout regions only report
+        // content insets rather than corner geometry. This value is measured off a system prompt:
+        // fitting its edge profile against a circle gives 52 pixels at 2x on every sample point.
```
zorgiepoo, line 169: If it's important to be precise, it doesn't sound like this code is going to age well.

Sparkle/SUUpdatePermissionPrompt.m
zorgiepoo: My AI assistant says to remove the layoutIfNeeded block of code and just make the `preferredMaxLayoutWidth=` for these controls in the xib from 260 -> 256, since the window is fixed sized at 256 + 20 + 20 (including leading/trailing margins).

Sparkle/SUUpdatePermissionPrompt.m
zorgiepoo: I am not too attached to adding a glass view, we don't have one for the update alert, and it looks OK without.

PR #2923, opened 2026-09-15
Fix appcast elements being ignored when the Sparkle namespace uses a custom prefix

Dev-next-gen:
While reading `SUAppcast.m` I noticed that `-sparkleNamespacedNameOfNode:` — the helper that resolves a node through the Sparkle namespace URI instead of through whatever prefix the feed happens to use — is applied to every element in the parser except two places: the children of `<sparkle:informationalUpdate>` and the children of `<sparkle:tags>`, which are still compared against their raw qualified name (`child.name`).

So a feed that binds the Sparkle namespace to a prefix other than `sparkle` (something `testnamespaces.xml` already exercises with `customNS`) gets these two elements parsed as empty:

- `<customNS:informationalUpdate><customNS:version>2.5</customNS:version></customNS:informationalUpdate>` yields an empty version set, and in `SPUAppcastItemStateResolver` an empty set means "informational for every host version". The item silently becomes information-only for everybody instead of only for the versions listed, so those users are sent to a link rather than offered the update.
- `<customNS:tags><customNS:criticalUpdate/></customNS:tags>` is not recognized at all, so the legacy critical-update tag is dropped.

The fix is to route both loops through the same helper the rest of the parser uses.

This is safe for existing feeds: `-sparkleNamespacedNameOfNode:` returns `[node name]` unchanged whenever the node's URI is not the Sparkle namespace, and returns `sparkle:` + local name when it is. For a feed using the `sparkle` prefix, or a feed with no namespace declaration at all, the resulting string is identical to `child.name`, so nothing that parses correctly today changes.

## Misc Checklist

The change was produced by an automated defect-hunting pipeline I build and run: it read the parser, spotted that the namespace helper was applied inconsistently, wrote the failing test, and confirmed the before/after behaviour on macOS CI. The two extra items in `testnamespaces.xml` and the two assertions in `testNamespaces()` are the whole of the new test coverage; the source change is three lines in `SUAppcast.m`.

## Testing

- [ ] Sparkle Test App
- [x] Unit Tests
- [ ] My own app
- [ ] Other (please specify)

I have no Mac here, so I ran the tests on GitHub Actions macOS runners from my fork, with the test change isolated from the source change:

- test only, without the fix — both new assertions fail at `SUAppcastTest.swift:1021` and `:1022`: https://github.com/Dev-next-gen/Sparkle/actions/runs/35037148184
- same test with the fix — passes: https://github.com/Dev-next-gen/Sparkle/actions/runs/35037151364
- the whole `Sparkle Unit Tests` target with the fix — passes: https://github.com/Dev-next-gen/Sparkle/actions/runs/35037446753

macOS version tested: macOS 26 (GitHub Actions `macos-26` runner, Xcode 26.6)


zorgiepoo:
Thank you.

PR #2925, opened 2026-09-16
Skip en and Base when generate_appcast records delta Sparkle locales

Dev-next-gen:
While reading how `generate_appcast` fills `sparkle:deltaFromSparkleLocales`, I noticed that the guard meant to leave out English and Base never matches. The loop in `ArchiveItem.swift` iterates over the directory names in `Sparkle.framework/Resources`, so `filename` is `en.lproj` or `Base.lproj`, but the guard compares it against `"en"` and `"Base"`. It only strips the `.lproj` extension on the next line.

As a result, both directories get recorded, even though the comment says they are the least likely to be stripped and are not worth checking. Only 7 locales are sampled (`maxLocalesToProcess`), and Sparkle.framework always ships `Base.lproj`, so these entries take slots that should go to locales an app is actually likely to strip. The client then checks `en.lproj` and `Base.lproj` for nothing.

The fix strips the extension first and then runs the guard on the locale name. Nothing else changes: the other locales are recorded exactly as before, and the client-side check in `SUAppcastDriver` reads the attribute the same way.

## Misc Checklist

The change was produced by an automated defect-hunting pipeline I build and run. It read the locale sampling code, saw that the guard compared the full directory name, and checked the before/after output of the real `generate_appcast` binary on macOS CI (details below). The source change moves one line and renames the variable in the guard.

## Testing

- [ ] Sparkle Test App
- [ ] Unit Tests
- [ ] My own app
- [x] Other (please specify)

There are no unit tests for `generate_appcast` and I have no Mac here, so I tested on GitHub Actions macOS runners from my fork. The workflow builds `generate_appcast` and `generate_keys` in Release and creates a signing key under a throwaway keychain account. It then builds two versions of a minimal app whose embedded `Sparkle.framework/Resources` contains `Base.lproj`, `en.lproj`, `de.lproj` and `fr.lproj`, zips them, runs `generate_appcast --ed-key-file` on the folder and reads the delta enclosure:

- without the fix, the delta enclosure has `sparkle:deltaFromSparkleLocales="de,en,Base,fr"`: https://github.com/Dev-next-gen/Sparkle/actions/runs/35090229016
- with the fix, it has `sparkle:deltaFromSparkleLocales="de,fr"`: https://github.com/Dev-next-gen/Sparkle/actions/runs/35090227993

The delta itself was generated and signed in both runs, so the rest of the delta path is unaffected.

macOS version tested: macOS 26 (GitHub Actions `macos-26` runner, Xcode 26.6)


zorgiepoo:
Makes sense. Note Sparkle 2.8+ dropped the `en` localization folder but we can still create delta updates from older versions.

PR #2927, opened 2026-09-16
Drop stale sparkle:format when generate_appcast embeds HTML release notes

Dev-next-gen:
When generate_appcast rewrites the embedded description of an existing item, it only ever adds `sparkle:format` for plain-text and markdown notes. It never removes it when the notes are HTML. So if an item was first generated from `Test2.txt` (or `.md`) with `--embed-release-notes`, and the notes later become an HTML fragment `Test2.html`, the next run replaces the CDATA with the HTML but leaves `sparkle:format="plain-text"` on the element. `SUAppcastItem` then reports the description as plain text and the update alert shows the raw tags instead of rendering them.

The fix removes the attribute in the HTML branch, since html is the default when the attribute is absent. The plain-text and markdown paths don't change, and nothing changes for new items or for descriptions that were already HTML.

I don't have a Mac here, so I ran the real binaries on a macos-26 runner (Xcode 26.6) on my fork. The workflow builds generate_appcast and generate_keys and makes a signed Test2.zip. Then it runs generate_appcast three times on the same folder: once with `Test2.txt` and `--embed-release-notes`, once after swapping in an HTML fragment, and once with `Test2.md` and `--embed-release-notes`.

- Before, on 2.x ([run](https://github.com/Dev-next-gen/Sparkle/actions/runs/35163107182)): after the second run the description is `<description sparkle:format="plain-text"><![CDATA[<ul><li>Fixed a crash.</li></ul>]]>`.
- After, with this branch ([run](https://github.com/Dev-next-gen/Sparkle/actions/runs/35163106754)): after the second run the description is `<description><![CDATA[<ul><li>Fixed a crash.</li></ul>]]>`. The third run then sets `sparkle:format="markdown"` exactly once.

About AI use: Claude Code found this while reading `writeAppcast`. It wrote the change and the proof workflow above, and the before/after runs are linked.

## Misc Checklist

## Testing

- [ ] Sparkle Test App
- [ ] Unit Tests
- [ ] My own app
- [x] Other (please specify)

The three generate_appcast runs described above: txt, then html, then md, all on the same existing item. They ran on a GitHub Actions macos-26 runner on my fork.

macOS version tested: macOS 26 (GitHub Actions macos-26 runner, Xcode 26.6)


generate_appcast/FeedXML.swift
```diff
@@ -519,6 +519,9 @@ func writeAppcast(appcastDestPath: URL, keys: PrivateKeys, disableEmbeddedSignWa
             descElement.setChildren([cdata])
             if descriptionFormat != "html" {
                 descElement.addAttribute(XMLNode.attribute(withName: SUAppcastAttributeFormat, stringValue: descriptionFormat) as! XMLNode)
+            } else {
+                // html is the default format, so drop a format left over from previously embedded release notes
+                descElement.removeAttribute(forName: SUAppcastAttributeFormat)
```
zorgiepoo, line 524: Did you test when it is already HTML, and the appcast description element has no attribute format, and confirmed this line is a no-op when no attribute format exists? The documentation for this method says nothing about whether or not the attribute name has to exist.

Dev-next-gen: Right, the docs don't say, and my earlier runs never hit that case: the second run always had an attribute to remove. So I ran both cases on a macos-26 runner (Xcode 26.6) with this branch: [job](https://github.com/Dev-next-gen/Sparkle/actions/runs/35180019407/job/105069937031).

First, a small Swift script calls `removeAttribute(forName: "sparkle:format")` on two `<description>` elements that don't have it, one with no attributes and one with only `xml:lang="en"`. `xmlString` is the same before and after for both, and `xml:lang` stays in place. Second, the real generate_appcast runs three times on the same folder, with `Test2.html` only. Run 1 creates the item, run 2 rewrites the existing HTML description that has no `sparkle:format`, and run 3 does the same after the HTML has changed. None of the three appcasts contains `sparkle:format`, the output of run 2 is byte-identical to run 1 (`diff` prints nothing), and run 3 shows the new `<ul><li>Fixed two crashes.</li></ul>`.

zorgiepoo:
Thanks

